PR #1, opened 2015-05-11
Good work.

aMarCruz:
:+1:

### DIFF
--- a/snabbdom.js
+++ b/snabbdom.js
@@ -66,7 +66,7 @@ function updateProps(elm, oldProps, props) {
       for (name in val) {
         on = val[name];
         if (on !== oldProps.style[name]) {
-          elm.style[name] = val[name];
+          elm.style[name] = on;
         }
       }
     } else if (key === 'class') {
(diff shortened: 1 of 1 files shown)


aMarCruz:
just for fun
alternative for 'Parse selector' in h()
```javascript
var selector = "div#foo.bar.boo";  //test
var props = {};  //test

var re  = /[#\.]([^#\.]+)/g;

//para mostrar las ocurrencias que se encontrarán
var cls = '';
selector = selector.replace(re, function(ch, token) {
  ch = ch[0];
  if (ch === '#') {
    props.id = token;
  } else if (ch === '.') {
    cls ? (cls += ' ' + token) : (cls = token);
  }
  return '';
});
props.className = cls;

//test...
console.log(selector);
console.log(props.id);
console.log(props.className);
//output:
//div
//foo
//bar boo
```

paldepind:
Hello

Thanks a lot for your corrections :) Great catches!

I don't know how you found this library. But since you're here let me share my goals:

1/ Clean and simple virtual DOM implementation – mostly done.
2/ Performance – (I have implemented the vdom benchmark and it looks good).
3/ Advanced hooks for implementing components – not done.

paldepind:
Your selector parser is really clever! And it handles when the id is not first. Like `div.some#id.class`. But it's a bit longer.

aMarCruz:
thanks @paldepind 
yeah, the selector parser is verbose and I don't test it, but I think is fast. Benchmark it, there is only one call to replace and only work if neccesary.
Good luck.

PR #2, opened 2015-05-26
Failed spec when using mixed keyed & non-keyed virtual nodes.

jails:
Nice implementation !
However just wanted to warn you that your patching algorithm doesn't work well when playing with keyed & non-keyed virtual nodes, there's probably still a minor glitch somewhere.

### DIFF
--- a/test/core.js
+++ b/test/core.js
@@ -112,7 +112,13 @@ describe('snabbdom', function() {
       assert(!elm.classList.contains('horse'));
     });
     describe('updating children with keys', function() {
-      function spanNum(n) { return h('span', {key: n}, n.toString()); }
+      function spanNum(n) {
+        if (typeof n === "string") {
+          return h('span', {}, n);
+        } else {
+          return h('span', {key: n}, n.toString());
+        }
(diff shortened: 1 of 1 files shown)


paldepind:
Thank you :)

Thanks for letting me know. I'll take a look at it. How did you discover this?

jails:
I found your `updateChildren()` algorithm so sexy that I wanted to adapt it for a personnal virtual dom projet. Since I wasn't able to make a spec to pass I added the spec in your test suite to see where I was wrong, but it didn't pass either :-p

paldepind:
I'm glad you find it sexy! Feel free to borrow whatever you want :) The children reconciliation is somewhat inspired by the one used in cito.js and kivi (they use almost the exact same algorithm).

It's really great that you found this issue! It's definitely a lot nice to discover while writing tests than in actual usage. The commit cd7eb76a should fix the issue.

jails:
Awesome it works great now :+1: !
No prob, I will keep you updated if I found other ordering issues ;-)

PR #15, opened 2015-08-13
module for setting element attributes

yelouafi:
Attributes are added and updated using `setAttribute`. In case of an attribute that has been previously added/set is no longer present in the `attrs` object, it is removed from the DOM element's attribute list using `removeAttribute`. 

In the case of HTML boolean attributes (.e.g. `disabled`, `hidden`, `selected` ...) attributes are removed if the attribute value is falsy in the JavaScript sens.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,11 +198,15 @@ h('a', {props: {href: '/foo'}, 'Go to Foo');
 
 ### The attributes module
 
-Same as props but set attributes instead of properties on DOM elements (i.e. `setAttribute`).
+Same as props but set attributes instead of properties on DOM elements
 
 ```javascript
 h('a', {attrs: {href: '/foo'}, 'Go to Foo');
 ```
+
+Attributes are added and updated using `setAttribute`. In case of an attribute that has been previously added/set is no longer present in the `attrs` object, it is removed from the DOM elemnt's attribute list using `removeAttribute`. 
+
+In the case of boolean attributes (.e.g. `disabled`, `hidden`, `selected` ...). The meaning doesn't depend on the attribute value (`true` of `false`) but depends instead on the presence/absence of the attribute itself in the DOM element. Those attributes are handled differently by the module : if a boolean attribute is set to a [falsy value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) (`0`, `-0`, `null`, `false`, `NaN`, `undefined`, or the empty string (`""`)) then the attribute will be removed from the attribute list of the DOM element.
 
 ### The style module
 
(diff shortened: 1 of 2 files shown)


paldepind:
This looks very good to me! I especially appreciate the documentation.

Maybe it would be better to construct an object where `obj[attrName] === true` for any boolean attribute. As in, iterate over an array of the attribute names an initialize the object with `true` values. I think it would perform better than the regexp.

What do you think?

README.md
```diff
@@ -196,6 +196,18 @@ Allows you to set properties on DOM elements.
 h('a', {props: {href: '/foo'}, 'Go to Foo');
 ```
 
+### The attributes module
+
+Same as props but set attributes instead of properties on DOM elements
+
+```javascript
+h('a', {attrs: {href: '/foo'}, 'Go to Foo');
```
ooflorent: There is a typo. A `}` is missing.

yelouafi:
@paldepind  'would perform better' is not that accurate, shoule be ['would perform fa..ar much better :)'](http://jsperf.com/regexp-vs-property-access)

@ooflorent thanks! fixed.

README.md
```diff
@@ -196,6 +196,18 @@ Allows you to set properties on DOM elements.
...
 
+### The attributes module
+
+Same as props but set attributes instead of properties on DOM elements
+
+```javascript
+h('a', { attrs: {href: '/foo'} }, 'Go to Foo');
+```
+
+Attributes are added and updated using `setAttribute`. In case of an attribute that has been previously added/set is no longer present in the `attrs` object, it is removed from the DOM element's attribute list using `removeAttribute`.
```
paldepind: These lines are too long. Please break them at 80 chars.

README.md
```diff
@@ -196,6 +196,18 @@ Allows you to set properties on DOM elements.
...
+
+Same as props but set attributes instead of properties on DOM elements
+
+```javascript
+h('a', { attrs: {href: '/foo'} }, 'Go to Foo');
+```
+
+Attributes are added and updated using `setAttribute`. In case of an attribute that has been previously added/set is no longer present in the `attrs` object, it is removed from the DOM element's attribute list using `removeAttribute`. 
+
+In the case of boolean attributes (.e.g. `disabled`, `hidden`, `selected` ...). The meaning doesn't depend on the attribute value (`true` of `false`) but depends instead on the presence/absence of the attribute itself in the DOM element. Those attributes are handled differently by the module : if a boolean attribute is set to a [falsy value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) (`0`, `-0`, `null`, `false`, `NaN`, `undefined`, or the empty string (`""`)) then the attribute will be removed from the attribute list of the DOM element.
```
paldepind: "`true` of `false`" should be _or_. No space before ":".

paldepind:
That's a nice performance boost! I've left some notes. Looks great otherwise :+1:

PR #20, opened 2015-08-24
Keep original handler object

stieizc:
The eventlistener helper is directly modifying the handler object (the `on` part of `data` passed to `h`). `data` will be silently modified and if the user want to reuse it, for example, in a loop, it has to be cloned on every call to `h`.
I don't know if my fix is correct, but I think this is a serious problem.

### DIFF
--- a/modules/eventlisteners.js
+++ b/modules/eventlisteners.js
@@ -11,6 +11,13 @@ function fnInvoker(o) {
   return function(ev) { o.fn(ev); };
 }
 
+function assoc(k, v, obj) {
+  var keys = Object.keys(obj), i = keys.length, copy = {};
+  while (i--) copy[keys[i]] = obj[keys[i]];
+  copy[k] = v;
+  return copy;
+}
+
 function updateEventListeners(oldVnode, vnode) {
   var name, cur, old, elm = vnode.elm,
       oldOn = oldVnode.data.on || {}, on = vnode.data.on;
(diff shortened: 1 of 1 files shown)


paldepind:
What are you trying to do? I'd like to see. It might be possible to change the approach so that this is not an issue. For instance instead of creating the object and reusing it in the loop you could probably create a new object in each loop.

I think modules are allowed to modify the data object. Otherwise the library would have to do a lot of cloning. It would waste performance and memory and in almost all cases it would be unneeded. Thus I think it would make more senses if users did cloning or in other ways avoided reusing objects.

stieizc:
Sorry for the long delay, been busy these days.
I tried to create a list of elements with the same click handler. I could not remember how I did it, but it did reuse the old data object, and since the eventhandler module  replace the handler function with its wrapper, the following elements gets the wrong handler. In the end I end up generating a new data object for each element (it turns out that I need to do that anyway). I will look into my code when I have time.
I don't like the cloning, too. If a library is modifying the object passed to it, especially when it is not obvious, it would be nice to have a big warning somewhere :).
Have a nice day.

paldepind:
Am I understanding you correctly that improved documentation would be an acceptable solution to you?

stieizc:
Yeah. I can help add it to the section in README after this week.

Sincerely,
王文鑫
Wenxin Wang
Department of Electronic Engineering,
Tsinghua University,
Beijing 100084, P. R. China
(+86)18811369901
Email：wangwx12@mails.tsinghua.edu.cn

Issue #21, opened 2015-09-08
style module does not remove old styles


ericgj:
If you remove a style from the vnode between patches, the style is not removed from the underlying element. It only adds and updates styles. Don't know how common this would be in practice, but I just got bitten by it today. I could open a PR if you agree it's a bug.

paldepind:
Yes. You're right. I think that is a bug.

It has never bothered me though. So I'm curious,  how did you get bit?

ericgj:
Toggling `display: none` , which in retrospect is usually unnecessary in virtual DOM views, and simple opacity transitions work nicely.  (Speaking of which there are a few typos in the README transitions examples which I will send a PR on too).

gaydenko:
@ericgj , have you tried `display: ''` ?

paldepind:
I've updated the documentation on the style module. What do you think @ericgj?

ericgj:
Looks good. Revert away :+1:

PR #22, opened 2015-09-09
styles module: remove styles in old vnode not in current (#21)

ericgj:


modules/style.js
```diff
@@ -23,6 +24,19 @@ function updateStyle(oldVnode, vnode) {
...
+  for (name in oldStyle) {
+    if (name === 'delayed') {
+      for (name in oldStyle.delayed) {
+        if (!curHasDel || !(name in style.delayed)) {
+          elm.style[name] = '';
+        }
+      }
+    }
+    if (name !== 'remove' && name !== 'destroy' && !(name in style)) {
+      elm.style[name] = '';  // the DOM is weird eh?
```
paldepind: Doesn't this mean that instead of supporting remove users can just set CSS values to the empty string?

ericgj: That's why I wondered if you considered it a bug - vs user error? On the one hand, it seems like 'style' should work like props, attrs, and h itself -- you are just describing the dom state at a point in time, and don't have to worry about imperatively removing past state. On the other, it does add to the code complexity tho and a (tiny) performance hit - so I would be ok taking it out and just documenting how to remove styles in the standard weird dom way.

paldepind: I think I would prefer just improving the documentation. It doesn't seem like it is worth it. It is very easy to do something like: `{position: shouldFollow ? 'fixed' : ''}`.

PR #23, opened 2015-09-09
fix style transition examples in README

ericgj:
Also, I am not clear when you would use the destroy vs remove style hook. It seems the remove hook is preferred because it guarantees the animation finishes before element removal? Why would you use the destroy hook - maybe a sentence should be added to the readme on this - unless the destroy hook is no longer supported?

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ the change is not applied until after the next frame.
 
 ```javascript
 h('span', {
-  style: {opacity: '0', transitionDuration: 'opacity 1s', delayed: {opacity: '1'}}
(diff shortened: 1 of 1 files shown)


paldepind:
Thank you :+1: 

This was my attempt at an explanation:

> The hook (`remove`) is only triggered when and element is to be removed from its parent – not if it is the child of an element that is removed. For that see the destroy hook.

But that could likely be improved. I'm thinking about it ;)

paldepind:
@ericgj I've added [a description of the destroy hook](https://github.com/paldepind/snabbdom#the-destroy-hook). Does it make sense?

ericgj:
Very clear, thanks!

PR #24, opened 2015-10-05
Update README.md

indus:


ooflorent:
Why did you change this? The previous line seems correct.

indus:
```
... {props: {href: '/foo'}, 'I\'ll take you places!'} ...
```
can´t be correct => ```Uncaught SyntaxError: Unexpected token }```

paldepind:
Nice catch! Thank you :+1:

PR #27, opened 2015-10-10
Make it possible to run a patch inside another patch

finnsson:
This pull request makes it possible to run a patch inside another patch. Previously the variable `insertedVnodeQueue` was shared between patch-runs. This commit makes the variable local to the current patch-run.

The use case is custom elements running snabbdom that render other custom elements that might be using snabbdom.

I checked the min+gzip size and found the previous size to be 1403 bytes and the new 1423 bytes.

### DIFF
--- a/snabbdom.js
+++ b/snabbdom.js
@@ -14,8 +14,6 @@ function emptyNodeAt(elm) {
 
 var emptyNode = VNode('', {}, [], undefined, undefined);
 
-var insertedVnodeQueue;
-
 function sameVnode(vnode1, vnode2) {
   return vnode1.key === vnode2.key && vnode1.sel === vnode2.sel;
 }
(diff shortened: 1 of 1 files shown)


ismail-codar:
+1 
Sometimes I have to use setTimeout.

paldepind:
Thanks a lot for this patch! Snabbdom aims to be as flexible and versatile as possible and this change makes it even more so.

I can certainly see how this can be useful. I have never needed it myself so this behaviour hasn't bothered me. I think this is a splendid change. Using a shared variable so that different `patch` invocations can affect each other really is a quite bad idea.

Thank you :+1:

paldepind:
It occurred to me that I hadn't published a new version with this change to npm. It is done now :)

PR #35, opened 2015-11-08
add warning to readme about sharing eventlistener data between VNodes

ericgj:


paldepind:
Splendid! This is exactly the kind of solution I had in mind. Generally only using literals inside `h` avoids the issue.

PR #42, opened 2015-12-14
Fixed hero example to account for #container being overwritten.

SkaterDad:
The hero animation example was not showing correctly (width would be 100% of device width, even on desktop).

The CSS was targeting the app container, which gets replaced as soon as the app starts.  I changed the CSS to target the wrapper div which is part of the vdom, and it looks right again.


paldepind:
Thank you!

PR #46, opened 2015-12-23
Fix typo

chenxsan:
Thought it's a typo?


paldepind:
Indeed. Thanks for fixing it. It's always nice to get rid of such a potential source of confusion.

PR #48, opened 2015-12-28
Fixed 'Uncaught ReferenceError'

chenxsan:
When using snabbdom with webpack, the webpack output file is in strict mode, and that would cause error:

> Uncaught ReferenceError: requestAnimationFrame is not defined at ...

This pull request fixed it.


paldepind:
Thank you :)

PR #51, opened 2016-01-04
Hero module updates

SkaterDad:
These commits make the hero module more flexible and easier to use.

It now handles inherited styles, text alignment, and many other situations with less friction.

The official example has been hacked to work with the new module as well, though it could be simplified greatly.

### DIFF
--- a/examples/hero/build.js
+++ b/examples/hero/build.js
@@ -28,14 +28,17 @@ var fadeInOutStyle = {
 
 var detailView = function detailView(movie) {
   return h('div.page', { style: fadeInOutStyle }, [h('div.header', [h('div.header-content.detail', {
-    style: { opacity: '1', remove: { opacity: '0' } } }, [h('div.rank', [h('span.header-rank.hero', { hero: { id: 'rank' + movie.rank } }, movie.rank), h('div.rank-circle', {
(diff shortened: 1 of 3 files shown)


paldepind:
This is really great! The hero module know looks much more robust :+1:

SkaterDad:
Happy to help out!

SkaterDad:
@paldepind Do you have a plan for when you're going to push a new release to npm?  I'd like to be able to use the updated hero module without copying it to my project.

Alternatively, have you thought of splitting the modules into their own repos which can be published separately?  With npm namespacing (ex:  @cycle/core, @cycle/http, etc), it would be easy to find them all.  It could make it easier to update the module examples too, since you could put in the Browserify/webpack configs.

paldepind:
I've just pushed a new release to npm with the updated hero module.

I don't think the modules are large or numerous enough that keeping them in separate repositories is warranted.

kuraga:
You may have one repo but many packages...

SkaterDad:
@paldepind I'm still seeing v0.2.8 in NPM.  Can you try releasing v0.3 again?

paldepind:
I just published to npm. I probably forgot the actual `npm publish` the first time around.

PR #52, opened 2016-01-09
Added failing style tests

garth:
The first test `explicitly removes styles` passes as expected.

The other two tests fail. Ideally setting a key should ensure that all element and child state is in sync.

Classes have the same issue.

### DIFF
--- a/test/style.js
+++ b/test/style.js
@@ -32,6 +32,39 @@ describe('style', function() {
     assert.equal(elm.style.fontSize, '10px');
     assert.equal(elm.style.display, 'block');
   });
+  it('explicialy removes styles', function() {
+    var vnode1 = h('i', {style: {fontSize: '14px'}});
+    var vnode2 = h('i', {style: {fontSize: ''}});
+    var vnode3 = h('i', {style: {fontSize: '10px'}});
+    patch(vnode0, vnode1);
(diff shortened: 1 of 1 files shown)


paldepind:
I'm fairly certain that this is because `key`s on the root node are not considered at all. The `key` property is only used when checking children to see if any of them has been moved.

garth:
I will update the tests and see. There is no documentation on keys so I'm just guessing.

garth:
@paldepind Other than explicitly unsetting every possible class or style, or having an intermediate render which "flushes" the previous dom, how are others solving this problem?

paldepind:
> No change when moving keyed items under a root node

That sounds like a bug. I'll have to take a closer look at it.

> Other than explicitly unsetting every possible class or style, or having an intermediate render which "flushes" the previous dom, how are others solving this problem?

I'm explicitly unsetting every possible class or style. I'm doing something like this:

```
h('div', {style: {selected ? '14px' : ''}})
```

garth:
@paldepind in any sufficiently complex system, each render can result in vnodes of different origins being compared. This is the case that I'm trying to solve. Unsetting a class or style is easy when you know that the same vnode will be rendered in the same location on every render, but how can you do it with comparing completely different structures?

garth:
Superseded by PR #57

PR #56, opened 2016-01-11
Thunks fixes

sprat:
I've fixed a bunch of problems with thunks, but it was harder than I thought. I encountered many problems:
- many tests assumed that the root node would never be replaced. But, if I remember correctly, I needed to mutate the root node for thunks, so I removed this limitation. I had to change many tests to remove the invalid assumptions.
- some tests use a real DOM node as second parameter of `patch`. I'm not sure if it's a valid use case. In doubt, I didn't change the test cases, but it's probably broken! (and was probably already broken). I think we'd better not support that at all, it's just a matter of removing the instanceof check for vnode in my merge request.
- I think it still doesn't work with nested thunks, i.e. thunks returning thunks. It may not be very difficult to implement that. But I'm wondering how to perform the matching of arbitrarily nested nodes when we compare thunks in the diff process. Maybe something similar to what is done in `updateChildren`. Also, what about the performance hit?
- there was a case sensitivity problem when comparing DOM nodes and vnodes, DOM node were always patched. I don't think I addressed the problem correctly but I have no idea how I can do that.
- I think it's not a good idea to include id & class in sel for the comparison, it triggers node replacement when a simple patching would suffice. I didn't addressed this problem, maybe I missed something.
- I'm not sure if I have all the cases covered in the unit tests. I'll see if it works fine enough in real-use ;)

### DIFF
--- a/snabbdom.js
+++ b/snabbdom.js
@@ -213,8 +213,7 @@ function init(modules) {
     for (i = 0; i < cbs.pre.length; ++i) cbs.pre[i]();
     if (oldVnode instanceof Element) {
       if (oldVnode.parentElement !== null) {
-        createElm(vnode, insertedVnodeQueue);
-        oldVnode.parentElement.replaceChild(vnode.elm, oldVnode);
+        oldVnode.parentElement.replaceChild(createElm(vnode, insertedVnodeQueue), oldVnode);
       } else {
         oldVnode = emptyNodeAt(oldVnode);
         patchVnode(oldVnode, vnode, insertedVnodeQueue);
--- a/snabbdom.js
+++ b/snabbdom.js
@@ -8,8 +8,9 @@ var is = require('./is');
 function isUndef(s) { return s === undefined; }
 function isDef(s) { return s !== undefined; }
 
+// deal with case sensivity better than that
 function emptyNodeAt(elm) {
-  return VNode(elm.tagName, {}, [], undefined, elm);
+  return VNode(elm.tagName.toLowerCase(), {}, [], undefined, elm);
 }
 
 var emptyNode = VNode('', {}, [], undefined, undefined);
@@ -27,14 +28,9 @@ function createKeyToOldIdx(children, beginIdx, endIdx) {
   return map;
 }
 
-function createRmCb(vnode, listeners) {
-  var i, elm;
-  if (isDef(i = vnode.data) && isDef(i = i.vnode)) vnode = i;
-  elm = vnode.elm;
+function createRmCb(childElm, listeners) {
   return function() {
-    if (--listeners === 0) {
-        elm.parentElement.removeChild(elm);
-    }
+    if (--listeners === 0) childElm.parentElement.removeChild(childElm);
   };
 }
 
@@ -50,10 +46,13 @@ function init(modules) {
   }
 
   function createElm(vnode, insertedVnodeQueue) {
-    var i, data = vnode.data;
+    var i, thunk, data = vnode.data;
     if (isDef(data)) {
       if (isDef(i = data.hook) && isDef(i = i.init)) i(vnode);
-      if (isDef(i = data.vnode)) vnode = i;
+      if (isDef(i = data.vnode)) {
+          thunk = vnode;
+          vnode = i;
+      }
     }
     var elm, children = vnode.children, sel = vnode.sel;
     if (isDef(sel)) {
@@ -83,6 +82,7 @@ function init(modules) {
     } else {
       elm = vnode.elm = document.createTextNode(vnode.text);
     }
+    if (isDef(thunk)) thunk.elm = vnode.elm;
     return vnode.elm;
   }
 
@@ -113,7 +113,7 @@ function init(modules) {
         if (isDef(ch.sel)) {
           invokeDestroyHook(ch);
           listeners = cbs.remove.length + 1;
-          rm = createRmCb(ch, listeners);
+          rm = createRmCb(ch.elm, listeners);
           for (i = 0; i < cbs.remove.length; ++i) cbs.remove[i](ch, rm);
           if (isDef(i = ch.data) && isDef(i = i.hook) && isDef(i = i.remove)) {
             i(ch, rm);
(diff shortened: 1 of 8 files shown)


paldepind:
Hello @sprat. I'm sorry for responding to you so very late.

> many tests assumed that the root node would never be replaced. But, if I remember correctly, I needed to mutate the root node for thunks, so I removed this limitation. I had to change many tests to remove the invalid assumptions.

Can you explain this? I don't think one can replace the root node with Snabbdom.

> some tests use a real DOM node as second parameter of patch. I'm not sure if it's a valid use case. In doubt, I didn't change the test cases, but it's probably broken! (and was probably already broken). I think we'd better not support that at all, it's just a matter of removing the instanceof check for vnode in my merge request.

Yes. That does not sound right. The second argument to `patch` mus be a virtual DOM node.

> there was a case sensitivity problem when comparing DOM nodes and vnodes, DOM node were always patched. I don't think I addressed the problem correctly but I have no idea how I can do that.

I don't understand. DOM nodes are not compared with vnodes. Only vnodes are compared against vnodes.

> I think it's not a good idea to include id & class in sel for the comparison, it triggers node replacement when a simple patching would suffice. I didn't addressed this problem, maybe I missed something.

That is intentional. It makes it easier to detect reordered notes without using the `key` property. To avoid node replacement one should set non-static classes with the class module.

sprat:
1. In fact, I'm building something on top of snabbdom, and I wanted to use a thunk as root node. In this context, the root node may mutate depending on the result of the function call. But it's also a valid use-case IMO when using regular vnodes. Imagine I have a widget that has a div as root node when expanded and becomes a link when collapsed. I think it's better to make the root node mutable for consistency.

2. So I'll fix the tests using a DOM node a second parameter and remove my fix...

3. DOM nodes are compared to vnodes in the first call to patch. Or, more exactly, you build vnodes from DOM nodes, but don't really mirror the tagName, attributes, classes from the DOM node. So in the first call to patch, you almost always have a destroy/create operation even if you have the same tagName in both sides.

4. Fair enough. I don't remember if it caused some problems, but it may not be optimal.

snabbdom.js
```diff
@@ -8,8 +8,9 @@ var is = require('./is');
 function isUndef(s) { return s === undefined; }
 function isDef(s) { return s !== undefined; }
 
+// deal with case sensivity better than that
```
paldepind: How can we deal with it better? Using `toLowerCase` seems fine to me?

snabbdom.js
```diff
@@ -45,10 +46,13 @@ function init(modules) {
   }
 
   function createElm(vnode, insertedVnodeQueue) {
-    var i, data = vnode.data;
+    var i, thunk, data = vnode.data;
```
paldepind: Snabbdom should not know about thunks. The functionality with regards to thunks should be contained in the thunk module. It should be possible.

thunk.js
```diff
@@ -9,7 +9,7 @@ function prepatch(oldThunk, thunk) {
   var i, old = oldThunk.data, cur = thunk.data;
   var oldArgs = old.args, args = cur.args;
   cur.vnode = old.vnode;
-  if (oldArgs.length !== args.length) {
+  if (old.fn !== cur.fn || oldArgs.length !== args.length) {
```
paldepind: Great idea! I had not thought about how the function might change as well.

test/attachto.js
```diff
@@ -33,9 +34,9 @@ describe('attachTo', function() {
         attachTo(elm, h('div#attached', 'New text')),
       ]),
     ]);
-    patch(vnode0, vnode1);
+    elm = patch(vnode0, vnode1).elm;
```
paldepind: Why do we need to update `elm`? It hasn't changed has it?

sprat: Not sure if it's necessary for this specific test, but it is for most tests since some root nodes are mutated. At first, I modified all the tests then I changed my mind, maybe this one was left out.

But maybe your remark is for all the tests?

snabbdom.js
sprat: Is it just a vocabulary issue? Since we descend into vnode.data.vnode, I needed to keep a reference on the starting point of the descent. Since it's only used for thunk or thunk-like things, I called the variable "thunk", but we could probably find a better name. But we can't use "parent" as it would be ambiguous. Do you have a suggestion?

snabbdom.js
sprat: I left this comment because people can also use uppercase tags in their views. I guess we should lowercase the tag after removing classes and ids.

test/attachto.js
paldepind: My remark was for all the tests. How are the root notes mutated?

sprat: You create another node whenever "sel" is different than the previous vnode one. Since the starting node is just "div" when you start from a dom node, you have a root node re-creation in most tests due to a class or a different tagName in the new vnode.

Before my work, the root node was never re-created, so we missed the ability to change either its tagName, class or id, when specified in "sel". That's fixed in my implementation but I needed to fix some tests too to make it work.

Hope I'm clear.

kay999:
How about merging this patch? I really need the ability to change the root-node. In the moment I use lots of unneccessary wrapping-divs to circumvent this problem.

About "why is it needed?": I use this code to create componets which can redraw themselfs after something is changed:

```javascript
exports.stateful = function stateful(tag, opts, func) {
	var state = { vnode:null };

	function redraw() {							// patch into stored parent-node
		var chs = state.vnode.children;
		var c = func(redraw);

		//--- this doesn't work because c is now the root-node, which can't be changed in current snabbdom
		chs[0] = M.patch(chs[0], c);			// only redraw the changed first element
	}

	state.vnode = h(tag, {}, [func(redraw)]);	// put stateful-element into first element
	return state.vnode;
};
```

kay999:
Hm, after playing with this patch I noticed that the remove-hook isn't working correctly if the root element has changed: Using { style:{remove: ...}} doesn't work anymore.

sprat:
Feel free to improve my patch as I've no time to work on it.

PR #57, opened 2016-01-11
Ensure that class, props and style artefacts are not left after re-render

garth:
I realise that this may have a minor performance impact, but it does fix a lot of render issues when comparing very different vnode structures.

It may also need to be done to on-* events as well.

### DIFF
--- a/modules/class.js
+++ b/modules/class.js
@@ -2,6 +2,11 @@ function updateClass(oldVnode, vnode) {
   var cur, name, elm = vnode.elm,
       oldClass = oldVnode.data.class || {},
       klass = vnode.data.class || {};
+  for (name in oldClass) {
+    if (!klass[name]) {
+      elm.classList.remove(name);
+    }
+  }
   for (name in klass) {
     cur = klass[name];
     if (cur !== oldClass[name]) {
--- a/modules/props.js
+++ b/modules/props.js
@@ -1,6 +1,11 @@
 function updateProps(oldVnode, vnode) {
   var key, cur, old, elm = vnode.elm,
       oldProps = oldVnode.data.props || {}, props = vnode.data.props || {};
+  for (key in oldProps) {
+    if (!props[key]) {
+      delete elm[key];
+    }
+  }
   for (key in props) {
     cur = props[key];
     old = oldProps[key];
--- a/modules/style.js
+++ b/modules/style.js
@@ -10,6 +10,11 @@ function updateStyle(oldVnode, vnode) {
       oldStyle = oldVnode.data.style || {},
       style = vnode.data.style || {},
       oldHasDel = 'delayed' in oldStyle;
+  for (name in oldStyle) {
+    if (!style[name]) {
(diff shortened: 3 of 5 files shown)


garth:
@paldepind any comments on this PR?

paldepind:
Sorry for the late response. I've just had exams so I've been putting of keeping up with GitHub.

I'm not opposed to this. The implementation is great. But it does add a performance overhead.

In my use cases unsetting values has been the most natural thing to do for me. I'm curious about your use case and what you mean when you talk about comparing very different vnode structures.

garth:
snabbdom is only comparing and updating params on the incoming vnode, not the outgoing. This means that if they happen to be from the same origin it will work, but if not then some items will not be compared and class names, styles, etc.. will get left behind from the previous render.

The fix is very simple, as well as comparing the incoming vnode, also check that the outgoing vnode for differences.

The overhead is very small, but makes the library reliable.

Lets say I have a vdom which is rendered via a set of nested components where a router defines which componets are rendered in the tree. When my app goes to /home the following set of components are rendred into a single vdom:

```
  App
    -> Header
      -> PageTitle
    -> Home
    -> Footer
```

Now navigate to /login and re-render

```
  App
    -> Header
      -> PageTitle
    -> Login
    -> Footer
```

App, Header, PageTitle and Footer update fine since they're the same vnode structure, but swapping Home for Login can cause problems because Login may have a similar structure, but does not know what needs to be unset from the previous render (Home or other).

paldepind:
Ok. Thank you for explaining. I understand your use case now.

In your example I'd probably put a different class on the Home and on the Login element. If the first is created like `h('div.home', ...)` and the second like `h('div.login', ...)` then Snabbdom will see it as a brand new element (because it determines that based on the string passed to `h` and the key) and recreate it from scratch. In most cases I guess just recreating the node will be more efficient than making a lot of updates to the existing element if they are very unlike.

But I can certainly see where you're coming from. And the current behavior is not intuitive. I guess it's one of those cases where it's never been a problem for me because I know Snabbdom inside out so I don't step into the traps. 

I'll merge this PR if you don't find the above workaround satisfiable.

garth:
@paldepind Since each of these issues is with a plugin, maybe each plugin could be configured via a factory and have a `{ strict: true }` option? that way the user of the lib can decided if they wan't maximum performance or if they prefer the guarantee that this PR offer?

paldepind:
I have given this some more thought and I've decided that I'm not happy with the above workaround. I think this problem will manifest in rare cases and lead to problems that may be hard to debug for someone who are not aware of the issue. That is a bad user experience.

Sure, this will slightly affect performance. But the modules are optional and if this ever becomes a bottleneck for anyone they can use unsafe modules. I think Snabbdom by default should ship with modules that are user friendly and with minimal surprises.

Thank you @garth for bringing this to my attention and for explaining the cases in which this is actually a problem.

paldepind:
I have published a new release with the changes in this PR. Thanks again.

PR #60, opened 2016-01-23
Add devDpendencies

mkwtys:


paldepind:
Yes. These should indeed be added. Thank you.

PR #62, opened 2016-02-05
Update README.md

jkleiser:
Several minor corrections. Added a few commas to make reading easier. Please review carefully, as my understanding of the matter is still not the best.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ needed.
 
 Snabbdom consists of an extremely simple, performant and extensible core that
 is only ≈ 200 SLOC. It offers a modular architecture with rich functionality
-for extensions through custom modules. To keep the core simple all non-essential
+for extensions through custom modules. To keep the core simple, all non-essential
 functionality is delegated to modules.
 
 You can mold Snabbdom into whatever you desire! Pick, choose and customize the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -203,22 +203,21 @@ could affect the position of the inserted elements.
...
-The hook is only triggered when and element is to be removed from its parent –
-not if it is the child of an element that is removed. For that see the destroy
+The hook is only triggered when an element is to be removed from its parent –
+not if it is the child of an element that is removed. For that, see the destroy
 hook.
 
 #### The `destroy` hook
 
-This hook is invoked whenever an element removed from the DOM or the child
+This hook is invoked whenever an element removed from the DOM or from the child
```
paldepind: I'm not sure if either of these are understandable. The `destroy` hook is invoked on a virtual node when its element is being removed from the DOM or when it is the child of an element that is being removed from the DOM.

paldepind:
Very great changes! Thank you.

I have left a single comment.

README.md
jkleiser: Could you yourself just put in the words that you think is most correct, or do you prefer that I do the rephrasing according to your suggestion?

jkleiser:
Could you yourself just put in the words that you think is most correct, or do you prefer that I do the rephrasing according to your suggestion?

README.md
```diff
@@ -280,22 +279,22 @@ h('a', {props: {href: '/foo'}}, 'Go to Foo');
...
 ```
 
 Attributes are added and updated using `setAttribute`. In case of an attribute
-that has been previously added/set is no longer present in the `attrs` object,
+that has been previously added/set and is no longer present in the `attrs` object,
 it is removed from the DOM element's attribute list using `removeAttribute`.
 
-In the case of boolean attributes (.e.g. `disabled`, `hidden`, `selected` ...).
-The meaning doesn't depend on the attribute value (`true` or `false`) but depends
+In the case of boolean attributes (.e.g. `disabled`, `hidden`, `selected` ...),
```
jkleiser: Should be no "." before "e.g.".

paldepind:
I've merged the PR and fixed the two issues discussed above.

Thank you a lot for these improvements to the readme :+1:

Issue #64, opened 2016-02-11
Can't patch elements inside an iframe


rayd:
I'm using snabbdom-jsx along with snabbdom for this (contrived) example. But I'd like to be able to patch the `<div>` inside the `<iframe>`. This code *should* do it I would think, but I run into an error.
```javascript
let frame = document.createElement('iframe');
frame.srcdoc = "<div>Thing</div>";
frame.onload = () => {
    patch(frame.contentDocument.body.querySelector('div'), <div>Other thing</div>);
};
document.body.appendChild(frame);
```

I'm getting the error `Uncaught TypeError: Cannot read property 'props' of undefined` because we're running the [`patchVnode`](https://github.com/paldepind/snabbdom/blob/master/snabbdom.js#L180) function when the `oldVnode` is still the actual `Element`. It hasn't been patched to a `VNode` because this is the first time `patch` is being called on it. 

I think the issue stems from [this line](https://github.com/paldepind/snabbdom/blob/master/snabbdom.js#L214). The `instanceof` check is failing because the element inside the iframe is actually an instance of `frame.contentWindow.Element`, not the `Element` reference in the outer frame. Hopefully that makes sense. 

I'm wondering if we can do a bit looser check when looking at the `oldVnode` to see if it's an `Element` or not. Maybe something along the lines of 
```javascript
if (oldVnode.nodeType === Element.ELEMENT_NODE) {
```

paldepind:
Yes. That looks like a bug to me. And you analysis looks right as well.

I think the check you propose is a fine solution. I can't think of anything better myself.

Will you try it out and see if it fixes your problem?

rayd:
Just tried it out and seems to be addressing that issue. Just made a PR for it. Thanks!

PR #66, opened 2016-02-12
Should be able to patch nodes created inside an iframe.

rayd:
Added a failing test case and the change to fix it. For #64.


paldepind:
Really nice! Thumbs up for adding a test case as well :smile: 

I will merge an do a release.

PR #67, opened 2016-02-12
Fix element node detection in patch

jimf:
`Element.ELEMENT_NODE` is non-standard and undefined in phantomjs.
Updating to use `Node.ELEMENT_NODE`.

References:
- MDN Element: https://developer.mozilla.org/en-US/docs/Web/API/Element
- MDN Node.nodeType: https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType

### DIFF
--- a/snabbdom.js
+++ b/snabbdom.js
@@ -1,5 +1,5 @@
 // jshint newcap: false
-/* global require, module, document, Element */
+/* global require, module, document, Node */
 'use strict';
 
 var VNode = require('./vnode');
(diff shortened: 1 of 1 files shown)


paldepind:
Oops. You are indeed right. I hope the latest release with this bug didn't cause you too much trouble. I'll do a new release with this patch right now.

jimf:
Thanks! :beers:

Nope, minimal trouble. Just had to define the constant in my tests.

PR #68, opened 2016-02-13
Make sure we remove a single text node when patching it's parent element.

rayd:
When a vnode has a `text` property  and is patched with a vnode with no `text`, the text node is never removed. Similarly, when a vnode has a `text` property and is patched with `children`, the text node created via the old `text` property isn't removed, so you end up the children being appended after the extra text node.
```html
<div>One</div>
``` 
becomes 
```html
<div>One<div>Two</div><span>Three</span></div>
```
instead of 
```html
<div><div>One</div><span>Two</span></div>
```
which, I believe, is unexpected behavior. This adds some test cases to catch this and the fix.


paldepind:
Thank you for finding the bug and for the fix! I never though about this case.

PR #70, opened 2016-02-15
Added carousel example with README

jkleiser:


paldepind:
I really like this. Thank you :smile:

paldepind:
I've now added the example to the readme. Thanks again.

Issue #63, opened 2016-02-08
Save excursion


katyo:
When we change property `value` of text inputs, usually we would like to save cursor position, selection range and focus of element, which is in focus.
In my experience, there is two cases, which assume different workarounds:
1. The new `value` is same as the `value` of focussed input: We may simply don't assign new value without any extra work.
2. The new `value` is different from current `value` of focussed input (for example, the value was changed by controller): We need get *selection* before changing value, and set selection after that. If *selection* is also set using properties, we must use it instead of current *selection*.
I tried to change **props** plugin to do that. Is it appropriate way? Or I need to create different plugin to work with inputs.
At the moment, all my fixes is quite simple:

```javascript
function updateProps(oldVnode, vnode) {
  var key, cur, old, elm = vnode.elm,
      oldProps = oldVnode.data.props || {}, props = vnode.data.props || {},
      preserveExcursion = "value" in props && typeof elm.selectionStart === "number" &&
                          document.activeElement === elm && props.value !== elm.value,
      /* save excursion */
      selection = preserveExcursion ? [
        typeof props.selectionStart === "number" ? props.selectionStart : elm.selectionStart,
        typeof props.selectionEnd === "number" ? props.selectionEnd : elm.selectionEnd
      ] : null;
  for (key in oldProps) {
    if (!props[key]) {
      delete elm[key];
    }
  }
  for (key in props) {
    cur = props[key];
    old = oldProps[key];
    if (old !== cur && (key !== "value" || elm[key] !== cur)) {
      elm[key] = cur;
    }
  }
  /* restore excursion */
  if (preserveExcursion) {
    elm.setSelectionRange(selection[0], selection[1]);
    elm.focus();
  }
}

module.exports = {create: updateProps, update: updateProps};
```

paldepind:
I agree that it would be nice if the props module automatically implemented this more intuitive behavior.

If I understand correctly you check `props.selectionStart` and `props.selectionEnd` such that users can control the selection manually. Otherwise I defaults to preserving the existing cursor position.

As far as I can tell this might still result in odd behavior. For instance if `values` is `more|text` (where `|` is the cursor position) and it is changed to `somemoretext`. Then the result would be ``some|moretext` because the cursor index is preserved. However fixing that would probably end up being too complex.

Would you like to submit a PR?

katyo:
Yes, we may provide selectionStart/End properties when we need to change value to correctly set cursor position in this case.

[...]

PR #71, opened 2016-02-15
Fix cursor jumping to the end of "controlled" inputs

garth:
DOM `INPUT.value` should never be updated when it matches the new vnode.

This fixes issues where typing into an INPUT with a sync event handler cause a render (such as for validation feedback) where setting `INPUT.value` also pushes the cursor to the end of the input.

### DIFF
--- a/modules/props.js
+++ b/modules/props.js
@@ -9,7 +9,7 @@ function updateProps(oldVnode, vnode) {
   for (key in props) {
     cur = props[key];
     old = oldProps[key];
-    if (old !== cur) {
+    if (old !== cur && (key !== 'value' || elm[key] !== cur)) {
       elm[key] = cur;
     }
   }


paldepind:
@katyo brought something similar up just recently here #63. In his/her code the same fix was made among other things. I'd like to hear what you think about it?

I think the cursor preservation in @katyo proposal is very nice. But this PR certainly fixes a significant part of the issue and since you mentioned that this is a present problem for you I will merge this PR and do a release.

garth:
With this fix #63 should also be fixed. Range will be preserved.

paldepind:
@garth With the code #63 one can additionally change the input `value` with `props` and still preserve cursor position.

garth:
@paldepind if the re-render happens on the same tick then the cursor position or selection should be unaffected by the render, in which case this PR should be sufficient.

If you want to update the value (and preserve or modify the selection range) then I think that is a more specialised operation that could possibly be handled by the app which can monitor the selection, as with any regular app that manipulates the DOM directly would have the same issue.

PR #72, opened 2016-02-16
Add iframe support.

rayd:
Not sure if you'll want to include this in snabbdom or not but figured I'd throw it out there. These changes allow proper patching of `<iframe>` elements. This would allow one to do something like:
```javascript
var vnode0 = document.createElement('iframe');
document.body.appendChild(vnode0);
var vnode1 = h('iframe', [
    h('ul', [
        h('li', 'Thing 1'),
        h('li', 'Thing 2')
    ]),
    h('span', 'Another thing.')
]);
patch(vnode0, vnode1);
patch(vnode1, h('iframe', [
    h('ul', [
        h('li', 'Thing 0'),
        h('li', 'Thing 1'),
        h('li', 'Thing 2')
    ])
]));
```
And the childNodes will be added/removed from the `contentDocument.body` of the iframe when the iframe is fully ready/loaded. 

In our application, we're using iframes to sandbox certain content, which is why I'm hoping to be able to perform patches on iframes. The main issue with trying to patch iframes is that appending a child to the iframe doesn't do what one would expect - we really should append children to the body of the iframe, so that they appear on the page. The basic gist of this PR is that I've added a function for performing abstracted DOM operations. It performs differently when applied to an iframe vs. a "normal" element. For an iframe it waits until the iframe is ready/loaded and then performs the DOM operation specified on the `contentDocument.body` of the iframe. I wrote the function as small as I could manage so as to not bloat the snabbdom codebase (which also means that it's not as clear/obvious as it could be, although I think it's still easy enough to follow). I also duplicated most of the core tests to verify that they work when patching `<iframe>` elements.

I'd tried adding support for iframes as a snabbdom plugin but the only workable solution seems to be to perform different DOM operations for the special case of iframes.

Let me know what you think! Thanks!

### DIFF
--- a/snabbdom.js
+++ b/snabbdom.js
@@ -33,6 +33,26 @@ function createRmCb(childElm, listeners) {
   };
 }
 
+// Perform DOM operations differently for iframes.
+function dom(op, elm, first, second) {
+  if (elm.tagName !== 'IFRAME') {
+    if (op === 'textContent') {
+      elm.textContent = first;
+    } else {
+      elm[op](first, second);
+    }
+  } else {
+    // Make sure the iframe is loaded (i.e. we have contentDocument) before
+    // performing operations on the body.
+    var f = function() { dom(op, elm.contentDocument.body, first, second); }
+    if (elm.contentDocument && elm.contentDocument.readyState === 'complete') {
+      f();
+    } else {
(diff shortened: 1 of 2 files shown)


rayd:
I'm going to close this and re-implement using the new dom api impl added by #80.

PR #74, opened 2016-02-23
patching root node

kay999:
Continued a little bit sprat's work on making root-node changes work correctly, including calling the remove hook (important for doing animations).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,5 @@ node_modules
 
 # Vim
 *.swp
+
+test/browserified.js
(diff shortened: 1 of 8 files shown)


kay999:
With this patch my "stateful" components seem to work correctly. I use the following code:

```javascript
function stateful(func) {
	function redraw() {							
		vnode.data.vnode = M.patch(vnode.data.vnode, func(redraw));
	}
	let vnode = h('_thunk', { vnode: func(redraw)} );
	return vnode;
};
```

to create components with it's own state like this:

```javascript
function testButton() {
    let state = 1;
	return stateful(function(redraw) {
		function incr() {
			state++;
			redraw();
		}
		return h('button', { on:{click:incr} }, "value:" + state)
	});
}

```

PR #77, opened 2016-02-25
small cleanups

kay999:
did some final cleanups.
This works fine for my use cases and can be released as far as I can tell.


paldepind:
Released :smile:

Issue #47, opened 2015-12-24
Supporting other platforms other than the DOM


TylorS:
I was wondering if it would be possible, and if it is, if it would be *desirable*, to split the codebase up a little bit to allow for alternative platforms. Snabbdom being a facility to this. 

I'm currently working on my own framework, [Motorcycle.js](https://github.com/motorcyclejs), and I'd like to add support for [nativescript](https://www.nativescript.org/) in the near future. I was hoping to avoid the need to for creating something entirely specific for Motorcycle if possible, and thought it might be possible to rework Snabbdom to be able to organize this.

Love to hear the thoughts on this.

paldepind:
If I understand you correctly you're suggesting to split the DOM specific parts of Snabbdom from the non-DOM specific parts? What would you then put in place of the DOM specific parts? I'm not familiar with NativeScript? But would that make it possible to use parts of Snabbdom with the UI APIs from NativeScript?

TylorS:
Yeah, that would be my desire. I haven't investigated this too far yet to be honest.
I was thinking it may be possible to have Snabbdom to provide an API or interface that allows for different backends to cooperate with it. i.e. `appendChild()`, `updateProps()` etc.

The ultimate goal, would being able to share a considerable amount of UI code between web and mobile, sort of like React/React-Native without any of the React stuff. I would be highly motivated to create a working backend for NativeScript. 

Again, this is something I haven't fully investigated, I wanted to feel out some of the possible ways to go first. This would be a project I would plan to start in the next couple weeks, after I'm able to bring Motorcycle to v1.0.

If this isn't something you're interested in, no worries. Just figured it couldn't hurt to ask :smile:

paldepind:
I assume it would be possible. One could identify all the places where Snabbdom interacts with the DOM, and then inject those methods into Snabbdom. Given the small size of Snabbdom that would certainly we feasible. If it is possible to implement equivalents of the DOM functions with NativeScript then that might be a worthwhile approach.

But then there is the modules. Some of those are very HTML and CSS specific.

user471:
:+1: I am also looking for DOM Diff library whitch could be used with any dom and not only html.
I think, instead of using DOM methods directly there could be some abstraction over it.

DomMonipulator = {
     getChildNodes = ...
     appendChild = ...
     ...
}
And different implementations HtmlDomManipulator, NativeDomManipulator, etc.

paldepind:
[...]

PR #80, opened 2016-02-29
dom api implementation

ptol:
resolves https://github.com/paldepind/snabbdom/issues/47

### DIFF
--- a/snabbdom.js
+++ b/snabbdom.js
@@ -4,13 +4,14 @@
 
 var VNode = require('./vnode');
 var is = require('./is');
+var api = require('./htmldomapi.js');
 
 function isUndef(s) { return s === undefined; }
 function isDef(s) { return s !== undefined; }
 
 // deal with case sensivity better than that
 function emptyNodeAt(elm) {
-  return VNode(elm.tagName.toLowerCase(), {}, [], undefined, elm);
+  return VNode(api.tagName(elm).toLowerCase(), {}, [], undefined, elm);
 }
 
 var emptyNode = VNode('', {}, [], undefined, undefined);
@@ -30,14 +31,18 @@ function createKeyToOldIdx(children, beginIdx, endIdx) {
 
 function createRmCb(childElm, listeners) {
   return function() {
-    if (--listeners === 0) childElm.parentElement.removeChild(childElm);
+    if (--listeners === 0) {
+      var parent = api.parentNode(childElm);
+      api.removeChild(parent, childElm);
+    }
   };
 }
 
 var hooks = ['create', 'update', 'remove', 'destroy', 'pre', 'post'];
 
-function init(modules) {
+function init(modules, domApi) {
   var i, j, cbs = {};
+  api = domApi || api;
   for (i = 0; i < hooks.length; ++i) {
     cbs[hooks[i]] = [];
     for (j = 0; j < modules.length; ++j) {
@@ -62,16 +67,16 @@ function init(modules) {
       var hash = hashIdx > 0 ? hashIdx : sel.length;
       var dot = dotIdx > 0 ? dotIdx : sel.length;
       var tag = hashIdx !== -1 || dotIdx !== -1 ? sel.slice(0, Math.min(hash, dot)) : sel;
-      elm = vnode.elm = isDef(data) && isDef(i = data.ns) ? document.createElementNS(i, tag)
-                                                          : document.createElement(tag);
+      elm = vnode.elm = isDef(data) && isDef(i = data.ns) ? api.createElementNS(i, tag)
+                                                          : api.createElement(tag);
       if (hash < dot) elm.id = sel.slice(hash + 1, dot);
       if (dotIdx > 0) elm.className = sel.slice(dot+1).replace(/\./g, ' ');
       if (is.array(children)) {
         for (i = 0; i < children.length; ++i) {
-          elm.appendChild(createElm(children[i], insertedVnodeQueue));
+          api.appendChild(elm, createElm(children[i], insertedVnodeQueue));
         }
       } else if (is.primitive(vnode.text)) {
-        elm.appendChild(document.createTextNode(vnode.text));
+        api.appendChild(elm, api.createTextNode(vnode.text));
       }
       for (i = 0; i < cbs.create.length; ++i) cbs.create[i](emptyNode, vnode);
       i = vnode.data.hook; // Reuse variable
@@ -80,7 +85,7 @@ function init(modules) {
         if (i.insert) insertedVnodeQueue.push(vnode);
       }
     } else {
-      elm = vnode.elm = document.createTextNode(vnode.text);
+      elm = vnode.elm = api.createTextNode(vnode.text);
     }
     if (isDef(thunk)) thunk.elm = vnode.elm;
     return vnode.elm;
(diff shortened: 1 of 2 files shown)


ptol:
The idea is anyone can implement this interface
```
module.exports = {
  createElement: createElement,
  createElementNS: createElementNS,
  createTextNode: createTextNode,
  appendChild: appendChild,
  removeChild: removeChild,
  insertBefore: insertBefore,
  parentNode: parentNode,
  nextSibling: nextSibling,
  tagName: tagName,
  setTextContent: setTextContent
};
```
and use snabbdom with any type of dom.

rayd:
I like this. Similar to the changes I made to add <iframe> support (see #72), but more flexible. I think I could re-implement what's in #72 without having to change the core for people who don't care about iframes.

TylorS:
I'll be that guy who does a :+1:

snabbdom.js
```diff
@@ -232,7 +237,7 @@ function init(modules) {
     var insertedVnodeQueue = [];
     for (i = 0; i < cbs.pre.length; ++i) cbs.pre[i]();
 
-    if (oldVnode.nodeType === Node.ELEMENT_NODE) {
+    if (isUndef(oldVnode.sel)) {
```
paldepind: Nice workaround here!

paldepind:
This is nice. What are you planning to use this for @ptol?

I'm a bit hesitant to merge this as I'm not sure if and how people will use it?

@rayd That sounds like a really good idea. Iframe support is nice but also rarely needed.

ericgj:
:+1: from me.  Was just looking at the feasibility of rendering vnodes to something like [blessed](https://github.com/chjj/blessed) (ncurses), this would help I think.

ptol:
@paldepind There are dozen of js libraries with their own dom (all game engines and canvas/webgl libraries) so if someone need a virtual dom for such library the snabbdom would be a possible solution.

I have been using it with pixi.js
```
var nodeTypes = {};
nodeTypes.text = () => new PIXI.Text("", {font: "14px Verdana", fill: "white"});
nodeTypes.group = () => new PIXI.Container();
nodeTypes.sprite = () => new PIXI.Sprite(px.Texture.EMPTY);
...

function createElement(tagName){
  var node = nodeTypes[tagName]();
  if(node){
    node.tagName = tagName;
  }
  return node;
}

function insertBefore(parentNode, newNode, referenceNode){
    if(referenceNode){
      var i = parentNode.getChildIndex(referenceNode);
      parentNode.addChildAt(newNode,i);
    }else{
      parentNode.addChild(newNode);
    }
}

function removeChild(node, child){
  node.removeChild(child);
}

function appendChild(node, child){
  node.addChild(child);
}

function parentNode(node){
  return node.parent;
}

function nextSibling(node){
  var parent = node.parent;
  var i = parent.getChildIndex(node);
  var sublingIndex = i + 1;
  if(parent.children.length <= sublingIndex) return null;
  return parent.getChildAt(sublingIndex);
}

function tagName(node){
  return node.tagName;
}

module.exports = {
  createElement: createElement,
  appendChild: appendChild,
  removeChild: removeChild,
  insertBefore: insertBefore,
  parentNode: parentNode,
  nextSibling: nextSibling,
  tagName: tagName
};
```

ptol:
@paldepind Or do you have something different in mind for how this https://github.com/paldepind/snabbdom/issues/47 could be solved?

paldepind:
@ptol Nope. This definitely seems to be the way to solve #47. The above is exactly what I wanted to see. A proof of concept that this abstracted interface can actually be usefully implemented by anything besides the DOM. Using it to control the pixi.js scene graph is certainly useful.

snabbdom.js
```diff
@@ -30,14 +31,18 @@ function createKeyToOldIdx(children, beginIdx, endIdx) {
...
+      var parent = api.parentNode(childElm);
+      api.removeChild(parent, childElm);
+    }
   };
 }
 
 var hooks = ['create', 'update', 'remove', 'destroy', 'pre', 'post'];
 
-function init(modules) {
+function init(modules, domApi) {
```
paldepind: This is nitpicking but I think the variable names are slightly confusing. We call the actual DOM api `api` and the user supplied API `domApi`. But the user supplied interface implementation might have nothing to do with the DOM while the DOM interface certainly does.

What about renaming `api` to `domApi` and `domApi` to `userApi`. Alternatively rename `api` to `domImpl` and `domApi` to `apiImpl`. The former is probably best though.

ptol: I don't mind. You can change any names you think would be more suitable. For example, I changed parentElement to parentNode but maybe it should stay parentElement.
```
function parentNode(node){
  return node.parentElement;
}
```

paldepind: I think `parentNode` is good.

paldepind:
@ptol Thank you a lot for this PR! As you can see I've merged it.

Unless I'm mistaken then the code in this PR would force all the `patch` functions returned by `init` to use the same DOM API. I've made a commit that should keep the different APIs separate. So that one, for instance, can use Snabbdom to patch both the DOM and pixi.js on the same page with different `patch` instances. Will you try it ou @ptol? After that I'll do a release.

Thanks again :smile:

ptol:
>Unless I'm mistaken then the code in this PR would force all the patch functions returned by init to use the same DOM API. 
I don't think so,  `domApi` is a local variable 
```
var domApi = require('./htmldomapi.js');
```
so
```
var patch1 = snabbdom.init([], require('./pixiapi')); //pixi
var patch2 = snabbdom.init([]); //html
```
shouldn't affect each other.

> I've made a commit that should keep the different APIs separate

Do you mean this commit https://github.com/paldepind/snabbdom/commit/fb80c1fd80235e93803d51a7ab274d659fd1a317 ?
I don't see any related changes.

ericgj:
No strong feelings about it, but what about a new `patch` constructor instead of overloading `init` ?  Also to me it feels like the api should be the first parameter, and the modules the second. 

```js
var patch1 = snabbdom.initWithAPI( api, modules );
var patch2 = snabbdom.init( modules );   // === snabbdom.initWithAPI( htmldomapi, modules );
```

paldepind:
> I don't think so,  `domApi` is a local variable 

But `api` is local to the entire module t

> Do you mean this commit https://github.com/paldepind/snabbdom/commit/fb80c1fd80235e93803d51a7ab274d659fd1a317 ?
I don't see any related changes.

I've changed it so that `api` is local to the function returned by `patch` and move a few function that use `api` so that they are inside the created closure.

@ericgj

> No strong feelings about it, but what about a new `patch` constructor instead of overloading `init` ?  Also to me it feels like the api should be the first parameter, and the modules the second. 

A new `initWithApi` function wouldn't bother me—but I'm indifferent. I don't mind the overloading here but I can see the appeal in avoiding it. Why do you feel the API argument should be first?

ptol:
Oh, I see. api should definitely be in function scope. I used last version with different dom on one page and everything work fine.

paldepind:
Released and published to npm.

@ptol If you feel like it I think publishing your pixi.js implementation of the DOM interface would be really great! We can create a folder in the repo for API implementations or you can create a repository of your own and I will link to it from the Snabbdom README.

ptol:
@paldepind Yes, I have plans to create a repository but I need to fix some problems before.
In canvas/webgl libraries there is a method `destroy`. This method need to be called when you don't need the object anymore. 

```
var vnode1 = h("group", {}, [h("text", {props: {text: "foo"}} )]);
var vnode1 = h("group", {}, []);
```
For example, this text object should be destroyed. What would be the best place to call this method? 
Can I do this?

```
function removeChild(node, child){
  node.removeChild(child);
  child.destory();
}
```
Or even it's removed this child can still be used somewhere?

ptol:
Here the repository https://github.com/ptol/virtual-pixi 
It seems there isn't any problem with 
```
function removeChild(node, child){
  node.removeChild(child);
  child.destory();
}
```

Swizz:
Hi,

This is a great valuable addition which require to be shown in the doc.
I use snabbdom for a while, however this is our google friend that driven me on that issue when I was looking for virtualdom tool for NativeScript.

Cheers,

PR #81, opened 2016-03-01
Catch errors thrown trying to set input type

musocrat:
IE eats it if an unsupported input type is attempted to be set. This patch drops into a try/catch if the prop is `type` and ~~defaults to `text` if it fails~~ deletes the `type` property if it fails. In that case the browser will fallback to the default type for the element.

### DIFF
--- a/modules/props.js
+++ b/modules/props.js
@@ -12,7 +12,7 @@ function updateProps(oldVnode, vnode) {
     if (old !== cur && (key !== 'value' || elm[key] !== cur)) {
       if (key !== 'type') elm[key] = cur;
       else {
-        try { elm[key] = cur; } catch (e) { elm.type = 'text' };
+        try { elm.type = cur; } catch (e) { elm.type = 'text'; }
       }
     }
   }
--- a/modules/props.js
+++ b/modules/props.js
@@ -10,7 +10,10 @@ function updateProps(oldVnode, vnode) {
     cur = props[key];
     old = oldProps[key];
     if (old !== cur && (key !== 'value' || elm[key] !== cur)) {
-      elm[key] = cur;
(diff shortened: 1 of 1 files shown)


musocrat:
I know this seems like a silly patch. If a browser doesn't support an input type don't include it in your app if you intend to support that browser.

That being said, it's not like HTML where it would simply ignore the property. In this case it throws an exception and the entire app stops working. For me it was a "date" input type on IE11, but I'm sure there are plenty of other HTML5 types that aren't supported by IE and will cause unexpected bugs in snabbdom driven apps.

musocrat:
You can close this if you want. I would close it myself, but I do see value in sticking it in the default module. 

Regardless, I've made a replacement module [`snabbdom-safe-props`](https://github.com/musocrat/snabbdom-safe-props)

paldepind:
> I would close it myself, but I do see value in sticking it in the default module.

Why would you close it?

I'm wondering if setting the troublesome properties using the attributes module works for you? It uses `setAttribute` and I don't think that functoin throws.

musocrat:
For my case it was resolved with the module override, but wasn't sure if this approach would be fitting for the main module. Using the attributes module would be cleaner and based on a quick test it seems that will work for this application.

Thanks

krainboltgreene:
Perhaps this belongs on the helper level? I have https://www.npmjs.com/package/snabbdom-helpers and I do content catching for `void elements`.

paldepind:
I wonder if one could instead set the troublesome properties using the attributes module. I don't think it would throw errors.

musocrat:
I can confirm that using the attributes module throws no errors in the same case. I'm going to go ahead  and close this pull request. 

Thanks

PR #85, opened 2016-03-13
Fix safety check for requestAnimationFrame

staltz:
Doing `window &&` does not achieve the expected behavior, it instead crashes with an error `ReferenceError: window is not defined`. This fix is simple and necessary for server-side rendering.

### DIFF
--- a/modules/hero.js
+++ b/modules/hero.js
@@ -1,4 +1,4 @@
-var raf = (window && window.requestAnimationFrame) || setTimeout;
+var raf = (typeof window !== `undefined` && window.requestAnimationFrame) || setTimeout;
 var nextFrame = function(fn) { raf(function() { raf(fn); }); };
 
 function setNextFrame(obj, prop, val) {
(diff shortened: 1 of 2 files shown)


paldepind:
You are indeed correct. Thank you for the fix.

avesus:
Awesome!

-- 
Best Regards,
Ivan Borisenko

TylorS:
This actually introduces an error with uglify-js
The check for uses backticks instead of regular quotes which causes a parsing error.

avesus:
Great that you've reported the bug. I've seen the same thing with
snabbdom-to-html which forces me to include a compiled file from the lib
subdirectory instead index.js. Node.js doesn't yet support backticks.

>

-- 
Best Regards,
Ivan Borisenko

staltz:
Oh, my bad. I thought it was ES6 code.

paldepind:
No problem. This is fixed in 0.4.2.

TylorS:
Thank you both @staltz and @paldepind :+1:

PR #88, opened 2016-03-24
h: Test args for defined values instead of args.length

jryans:
In my app, I am wrapping calls to `h` with another helper. This helper passes through args to `h` after applying some logic, as in:

```js
function helper(first, second, ...rest) {
  // ... logic ...
  h(first, second, ...rest);
}
```

Checking `arguments.length` causes this helper to fail, since it is just passing through values, so there could be only 2 "truthy" arguments, but `h` would receive `h(first, second, undefined)`, breaking the argument processing in `h`.

Switching to truthy checks appears to solve this and preserve previous behavior. What do you think of this?

Thanks for working on this library! :smile:

### DIFF
--- a/h.js
+++ b/h.js
@@ -12,11 +12,11 @@ function addNS(data, children) {
 
 module.exports = function h(sel, b, c) {
   var data = {}, children, text, i;
-  if (arguments.length === 3) {
+  if (c !== undefined) {
     data = b;
     if (is.array(c)) { children = c; }
     else if (is.primitive(c)) { text = c; }
-  } else if (arguments.length === 2) {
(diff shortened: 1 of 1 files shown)


paldepind:
Hello @jryans. Sorry for the late response :cry: I think this a fine change and I can understand your issue.

But I don't like truthy testing. I'd rather use `!== undefined` instead as this is explicit about what is being tested.

Fix that and I will merge and release :+1:

jryans:
Okay, I believe I've made the changes you asked for. Thanks for taking a look!

paldepind:
Splendid. Thank you.

Issue #90, opened 2016-03-25
dataset module


hville:
**Summary - tl;dr**
I'm currently using a tiny and trivial `dataset` module and I am wondering if it is worth sharing (tiny size but limited benefits). I can prepare a PR with tests if there is any interest and/or recommendations.

**Background**
Because the `node.dataset` property is read-only, using `{props: {dataset: {foo: 'foo'}}}` does not work. The attributes module can be used but wrapping the attribute names get a little tedious: `{attrs: {'data-foo': foo}}`. The little module below allows for a new `dataset` item in the `data` object so that we can write: `{dataset: {foo: 'foo'}}`.

```
function updateDataset(oldVnode, vnode) {
  var key
  var elm = vnode.elm
  var oldDataset = oldVnode.data.dataset || {}
  var dataset = vnode.data.dataset || {}

  for (key in oldDataset) if (!dataset[key]) delete elm.dataset[key]
  for (key in dataset) if (oldDataset[key] !== dataset[key]) elm.dataset[key] = dataset[key]
}
module.exports = {create: updateDataset, update: updateDataset}
```

**Use Case**
Works nicely for event handlers where we can update a `data-action` parameter instead of recreating a function specific to every elements. In my case, I use it to store the origin JSONPointer in `data-path` so that the event handler can produce a JSONPatch. Example: 

```    
function commonHandler(e) {  //same handler declared once for all element on every render
  store.patch([{
    op: 'replace', 
    path: e.target.dataset.path, 
    value: e.target.value  // ommitting additional formatting and validation
  }])
}
vtree = h('input', {
  props: {
    value: store.get().foo.bar.baz,
    oninput: commonHandler
  },
  dataset: {path: '/foo/bar/baz'} //same outcome as attrs:{'data-path': '/foo/bar/baz'}
})
```

Both `{attrs: {'data-foo': foo}}` and `{dataset: {foo: 'foo'}}` modify the underlying HTML so I am not sure if it is not be better to simply use some random custom property instead: `{props: {datasetfoo: 'foo'}}`.

Feedback/suggestions welcome

TylorS:
[...]

PR #91, opened 2016-03-27
Pr dataset module

hville:
new dataset module and module tests as discussed in #90

### DIFF
--- a/modules/dataset.js
+++ b/modules/dataset.js
@@ -1,0 +1,19 @@
+function updateDataset(oldVnode, vnode) {
+	var elm = vnode.elm,
+		oldDataset = oldVnode.data.dataset || {},
+		dataset = vnode.data.dataset || {},
+		key
+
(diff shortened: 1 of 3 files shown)


paldepind:
Thank you @hville. This is a great extra convenience to have :smile:

krainboltgreene:
No documentation was added to the README.md for this :(

hville:
@paldepind I don't mind adding documentation but I was concerned it would crowd-up the main page since it more of a niche module. If you want, I could create a separate `readme.md` in the modules folder to park info for current and future modules.

@krainboltgreene With the `dataset` module installed, you can do `h('div', {dataset: {foo: 'foo'}})`. This has the same result as`h('div', {attrs: {'data-foo': 'foo'}})` with the `attributes` module.

krainboltgreene:
Oh shit, that's awesome.

We should do that for all the microformats like aria and itemprop.

PR #96, opened 2016-04-11
reference htmldomapi by module name

squiddle:
require call should follow convention, even though the node loader won't complain either way.


paldepind:
You are right. But what about the two `require`s just above?

squiddle:
I do not understand your question. they already use the module name and not the file name.

paldepind:
I don't know what I was thinking of. I clearly didn't look at them properly.

Thank you for fixing this!

PR #97, opened 2016-04-13
New thunk implementation

paldepind:
This is a new thunk implementation.

When a thunk is inserted in a virtual DOM tree a dummy thunk is placed instead of an actual vnode. However during patching an actual node will have to be associated with the thunk. Either a vnode from the previous thunk or one obtained by calling the function associated with the thunk. The old thunk implementation worked by setting a property `vnode.data.vnode` on thunks during patching. This turned out to have the disadvantage that everything in the code that uses a `vnode` whould have to check the property `vnode.data.vnode` to see if that was actually the real vnode. Doing that increased the complexity and was never done properly.

The new implementation instead overrides all properties in the thunk so that it becomes an actual proper vnode. I tried another approach as well but this one turned out to be the simplest. One consequence is that `thunk` will now have to be told the tag/selector of the element of the vnode that it's function will return. Like this:

```js
    function oddEven(n) {
      var prefix = (n % 2) === 0 ? 'Even' : 'Odd';
      return h('div', {key: oddEven}, prefix + ': ' + n);
    }
    var vnode = h('div', [thunk('div', 'oddEven', oddEven, 4)]);
```

So, the first argument to `thunk` is the tag name, the second is a key, the third is a function and the rest is arguments for the function.

### DIFF
--- a/snabbdom.js
+++ b/snabbdom.js
@@ -52,12 +52,11 @@ function init(modules, api) {
   }
 
   function createElm(vnode, insertedVnodeQueue) {
-    var i, thunk, data = vnode.data;
+    var i, data = vnode.data;
     if (isDef(data)) {
-      if (isDef(i = data.hook) && isDef(i = i.init)) i(vnode);
-      if (isDef(i = data.vnode)) {
-          thunk = vnode;
-          vnode = i;
+      if (isDef(i = data.hook) && isDef(i = i.init)) {
+        i(vnode);
+        data = vnode.data;
       }
     }
     var elm, children = vnode.children, sel = vnode.sel;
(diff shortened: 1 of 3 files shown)


kay999:
Is it possible to replace the content of a think without redrawing the whole vnode-tree fro the root? I think it's very useful to have vnodes which don't apear in the DOM.

With the data.vnode mechanism this was possible and quite useful to create components which could redraw themself locally without having to redraw the whole tree. This seems to be impossible in this implementation without creating some kind of dummy-node in the DOM (a div for exampe).

paldepind:
@kay999 If I understand you correctly you want to diff only a subtree of the DOM. You might be able to achieve what you want, by using the `prepatch` hook. If you return the old vnode in the `prepatch` hook Snabbdom will not continue patching that subtree and you are free to update that subtree in any way you choose.

paldepind:
I'm merging. Please open an issue @kay999 if you cannot find a solution to your problem.

PR #102, opened 2016-04-19
Return early in module update functions

yyx990803:
Currently for each module, on every update, for every node, empty objects are allocated, enumerated and iterated over regardless of whether that node has data for the module or not.

Returning early prevents useless object allocation and `for ... in` enumeration, which should result in better memory usage and more efficient patching.

I'm observing around 10% fps increase in the dbmon benchmark.

### DIFF
--- a/modules/attributes.js
+++ b/modules/attributes.js
@@ -1,19 +1,23 @@
-var booleanAttrs = ["allowfullscreen", "async", "autofocus", "autoplay", "checked", "compact", "controls", "declare", 
-                "default", "defaultchecked", "defaultmuted", "defaultselected", "defer", "disabled", "draggable", 
-                "enabled", "formnovalidate", "hidden", "indeterminate", "inert", "ismap", "itemscope", "loop", "multiple", 
-                "muted", "nohref", "noresize", "noshade", "novalidate", "nowrap", "open", "pauseonexit", "readonly", 
-                "required", "reversed", "scoped", "seamless", "selected", "sortable", "spellcheck", "translate", 
+var booleanAttrs = ["allowfullscreen", "async", "autofocus", "autoplay", "checked", "compact", "controls", "declare",
+                "default", "defaultchecked", "defaultmuted", "defaultselected", "defer", "disabled", "draggable",
+                "enabled", "formnovalidate", "hidden", "indeterminate", "inert", "ismap", "itemscope", "loop", "multiple",
+                "muted", "nohref", "noresize", "noshade", "novalidate", "nowrap", "open", "pauseonexit", "readonly",
+                "required", "reversed", "scoped", "seamless", "selected", "sortable", "spellcheck", "translate",
                 "truespeed", "typemustmatch", "visible"];
-    
+
 var booleanAttrsDict = {};
 for(var i=0, len = booleanAttrs.length; i < len; i++) {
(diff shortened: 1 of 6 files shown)


trueadm:
@yyx990803 if you drop `for ... in` and use `Object.keys` you should see further performance benefits.

leeoniya:
@trueadm huh, i always imagined the extra array alloc from `Object.keys` would be more costly, no?

trueadm:
@leeoniya It was that way until a few versions of V8 ago where `for ... in` had a massive performance drop and started giving a `deopt`. With Inferno, we've replaced all `for ... in` statements with `Object.keys`  and saw a noticeable performance gain.

yyx990803:
@trueadm interesting, I did bench `for ... in` vs `Object.keys()` a while back and `for ... in` without `hasOwnProperty` check was faster in Node 5. But I assume that's because Node 5 is still using a pretty old version of V8...

leeoniya:
since jsperf is down, i wrote a quick bench. it does look like somewhat faster for `Object.keys`, though not by much. However, in Firefox `Object.keys` is *massively* slower than `for...in`. If you're targeting only Node, then it makes sense, but absolutely not as a general case. Correct me if I'm wrong:

```js
// single-hash
var hash = {}, key;
for (var i = 0, len = 1e6; i < len; i++) {
	key = Math.random().toString(36).substring(4);
	hash[key] = true;
}

var x;

console.time("for-in");
for (var i in hash)
	x = hash[i];
console.timeEnd("for-in");


console.time("for/keys");
for (var i = 0, keys = Object.keys(hash), len = keys.length; i < len; i++)
	x = hash[keys[i]];
console.timeEnd("for/keys");


// multi-hash
var len = 1e3, hashes = Array(len), key;
for (var j = 0; j < len; j++) {
	var hash = {};
	for (var i = 0; i < len; i++) {
		key = Math.random().toString(36).substring(4);
		hash[key] = true;
	}
	hashes[j] = hash;
}

console.time("for-in (multi)");
for (var j = 0; j < len; j++) {
	hash = hashes[j];
	for (var i in hash)
		x = hash[i];
}
console.timeEnd("for-in (multi)");


console.time("for/keys (multi)");
for (var j = 0; j < len; j++) {
	hash = hashes[j];
	for (var i = 0, keys = Object.keys(hash), len = keys.length; i < len; i++)
		x = hash[keys[i]];
}
console.timeEnd("for/keys (multi)");
```

trueadm:
@leeoniya Unfortunately, due to the nature of JIT compilers, it's very hard to benchmark such things with microbenchmarks. Real-world tests are much more effective. It's wise to actually compare the ASM of both your microbenchmarks rather than their output times, to see what they really look like once V8 gets its hands on them – you might find out something surprising.

leeoniya:
@trueadm no argument regarding V8, it's faster anyhow.

Edge, IE11, IE10 seem to perform the same for both versions. But Firefox is 50% slower (even in nightlies). Can't seem to find any bugs open for this in bugzilla. Time to do some reporting...

capture

leeoniya:
bug filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1266140

paldepind:
Thank you for the contribution @yyx990803. This is certainly a more sensible approach.

PR #108, opened 2016-05-06
eventlisteners should work if a listener is removed

kay999:
In the current version, a event-listener is never removed, which leads to problems in certain use cases. This patch fixes this.

(I hope I got everything correct, because I made this changes online).

### DIFF
--- a/modules/eventlisteners.js
+++ b/modules/eventlisteners.js
@@ -2,13 +2,17 @@ var is = require('../is');
 
 function arrInvoker(arr) {
   return function() {
+    if (!arr.length) return;
     // Special case when length is two, for performance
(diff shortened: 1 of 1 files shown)


paldepind:
Wouldn't it be a better idea to remove the event listeners with `removeEventListener` instead of still having them but not doing anything in them due to `if (!arr.length) return;`.

kay999:
I went with the "current way" of just pathing the handler-code for changes.

If there are lots of add/removes, simple patching the handler-objects to do nothing is probably faster. But using removeEventListener is cleaner. Hard to tell.

paldepind:
Well. That is a fair point. I'll take it.

Issue #103, opened 2016-04-21
classNames in `.sel` and `data.props.className` can't coexist.


TylorS:
Hey, I'm not sure if this is a bug or expected behavior, but I figured I might as well report it and see if it *is* a bug. A user of one of my libraries found this to be counter-intuitive.

Currently when using both a css selector in the first argument to `h()`, and also a `className` to the props module, only the className defined in props will be applied to the DOM node.

e.g.
`h('div.something', {props: { className: 'other' }}, [..])` 

you'll end up with `<div class="other">`, when I'd expect `<div class="something other">`.

Heres a real test case that reproduces the 'bug'. If you inspect the output, you'll see that the `.sel` of the 3rd element is clobbered.
http://esnextb.in/?gist=dd2f6f68f466b1d88088b2d3fbf458e8

paldepind:
Setting `className` on a DOM element always overwrites all existing classes. That is just the way the property works. Just like setting `innerHTML` overwrites all existing content of an element.

If that is not the desired behaviour then one should use a different method for setting classes. Like the classes module.

TylorS:
Thank you, good points!

PR #117, opened 2016-05-24
Don't overwrite other class attribute assignments.

bloodyKnuckles:
When a snabbdom virtual node has both a class as a component of the `sel` property AND a property in an `attrs` object, the `attrs` object class overwrites the `sel` subcomponent class.

https://github.com/paldepind/snabbdom/blob/master/modules/attributes.js#L26
`elm.setAttribute(key, cur);`

Here's an example of `attrs` class overwriting `sel` subcomponent class:
http://esnextb.in/?gist=0512c49748550e57039bce437225081e

To prevent overwriting I suggest adding a check to determine if `key` equals "class", and if so use `classList.add` rather than `setAttribute`. 

Also to accommodate a list of classes in the `attrs` class property, `split` the value on spaces and loop over the resulting array adding each item.

### DIFF
--- a/modules/attributes.js
+++ b/modules/attributes.js
@@ -19,11 +19,18 @@ function updateAttrs(oldVnode, vnode) {
     cur = attrs[key];
     old = oldAttrs[key];
     if (old !== cur) {
-      // TODO: add support to namespaced attributes (setAttributeNS)
-      if(!cur && booleanAttrsDict[key])
-        elm.removeAttribute(key);
-      else
-        elm.setAttribute(key, cur);
+      if ('class' === key) {
+        var clslst = cur.split(/ +/);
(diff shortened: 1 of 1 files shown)


paldepind:
Thank you for the PR, the well written explanation with clear examples. The implementation is great as well.

However, please see my response in #103. I think the same reasoning applies to this case.

To my mind, setting the `class` attribute should behave just like setting the class attribute always does. Because the attributes module is plain and simply about setting attributes.

The class module is for setting classes in a friendly and surprise free way. Unless I'm mistaken setting your classes with the classes module would solve your issues.

bloodyKnuckles:
Makes sense. Thanks for the explanation. I can work with that.

paldepind:
Great :smile:

PR #118, opened 2016-05-24
Default to DIV and allow any selector order.

bloodyKnuckles:
When parsing the `sel` property, while preparing a `patch` operation, default the tag name to DIV when none is provided, such as: `h('#myId', 'my undetermined element')`. Currently snabbdom `patch` throws an error if no tag name is provided in the `sel` property on the virtual dom object.

Example error: http://esnextb.in/?gist=34aae6ae833d915496c546bb1b60858f

Also, allow for any order and number of id's and classes. (Of course only one id is allowed on the `id` property.) Currently snabbdom `patch` ignores classes that occur before id's in the `sel` property.

Example ignore classes before id: http://esnextb.in/?gist=7927c373899c61a3451e79a881a8824b

Some explanation of code implementation:

``` javascript
// Parse selector
var selectors = sel.split(/((\.|#).+)/); // split on first id or class symbol
var tag = selectors[0] || 'div'; // first element is tag, default to DIV if empty string
selectors = selectors[1] || ''; // the rest are id's and classes, if anything

// no change here
elm = vnode.elm = isDef(data) && isDef(i = data.ns) ? api.createElementNS(i, tag)
                                                    : api.createElement(tag);

var hashparts = selectors.split('#'); // start with splitting out id components
var id,
  classes = hashparts.shift().split('.').slice(1);
  // first element of hash split has no id components, only classes, if anything
  // shift it off hashparts, and split it, and dump first item (always empty string)

for (i = 0; i < hashparts.length; i++) { // loop over remaining hash components
  var dotparts = hashparts[i].split('.'); // split classes, first item is an id
  id = dotparts.shift() || undefined; // shift id off dot components
  // each subsequent id overwrites the earlier id's
  // so only one id passes through, the last one in the sel string
  classes.push.apply(classes, dotparts);
  // push and flatten dot components onto classes array
}
if (id) elm.id = id;
if (classes.length) elm.className = classes.join(' ');
// classes is array of class names, so join with a space
```

I've also modularized the script and wrote a bunch of tests:
https://github.com/bloodyKnuckles/simple-selector-parse/blob/master/test/index.js

### DIFF
--- a/snabbdom.js
+++ b/snabbdom.js
@@ -61,16 +61,26 @@ function init(modules, api) {
     }
     var elm, children = vnode.children, sel = vnode.sel;
     if (isDef(sel)) {
+
       // Parse selector
-      var hashIdx = sel.indexOf('#');
-      var dotIdx = sel.indexOf('.', hashIdx);
-      var hash = hashIdx > 0 ? hashIdx : sel.length;
-      var dot = dotIdx > 0 ? dotIdx : sel.length;
-      var tag = hashIdx !== -1 || dotIdx !== -1 ? sel.slice(0, Math.min(hash, dot)) : sel;
+      var selectors = sel.split(/((\.|#).+)/);
+      var tag = selectors[0] || 'div';
+      selectors = selectors[1] || '';
+
       elm = vnode.elm = isDef(data) && isDef(i = data.ns) ? api.createElementNS(i, tag)
                                                           : api.createElement(tag);
-      if (hash < dot) elm.id = sel.slice(hash + 1, dot);
-      if (dotIdx > 0) elm.className = sel.slice(dot + 1).replace(/\./g, ' ');
+      var hashparts = selectors.split('#');
+      var id,
+        classes = hashparts.shift().split('.').slice(1);
+
+      for (i = 0; i < hashparts.length; i++) {
+        var dotparts = hashparts[i].split('.');
+        id = dotparts.shift() || undefined;
+        classes.push.apply(classes, dotparts);
+      }
+      if (id) elm.id = id;
+      if (classes.length) elm.className = classes.join(' ');
+
       if (is.array(children)) {
         for (i = 0; i < children.length; ++i) {
           api.appendChild(elm, createElm(children[i], insertedVnodeQueue));


krainboltgreene:
While common in other templating libraries I think this is actually a mistake. It lets you ignore semantic HTML tags. The `<div>` tag was meant as [a fallback, not the default](https://www.w3.org/wiki/HTML/Elements/div).

An upper library should maybe let you get away with it, but snabbdom shouldn't.

StreetStrider:
@krainboltgreene, semantic tags doesn't mean too much in context of UI. If you construct UI, you tend to use more common tag (like div) in most of the cases. It would have less browser issues due to wide use, it is fully visual customizable and it always can be tuned via `display` to serve as any other element.

I would like to see this feature. I'm using Jade/Pug for a long time and have no troubles with analogous feature in it. Using div as default tag always turn out to be better readable and sped up the markup process.

krainboltgreene:
Semantic tags means a whole lot in the context of UI. See: Accessability tooling. The `<div>` tag is *not* a "common tag". That's not really how it works. A `<div>` has *the exact same* "browser issues* as any other tag, every other tag is as "fully visual customizable" as the `<div>` tag, and any tag can have it's `display` changed, but *why would you if there's already a semantically correct tag that does the job*.

StreetStrider:
@krainboltgreene, hello. I understand your point on accessability, and using specific tags for the right job and I didn't mean other. I didn't mean «use div for everything» I just mean that the majority of UI's structure is still container elements without any special semantic meaning. Backbone of UI is still containers, pseudoelements and special display and position agreements and there's nothing fancy you can do by applying semantic tags here. The volumes of such «glue» markup are debatable though, but this doesnt cancel the fact, that div is the most common tool for such task.

mightyiam:
@bloodyKnuckles thank you for your contribution.

I personally feel that this is a nicety outside of the scope of snabbdom. It's easy to make a wrapper for `h` that does whatever defaults anyone fancies having.

StreetStrider:
@mightyiam it's not easy, since you need to parse first argument: separate possible tag portion from others, then (maybe) prepend `div` and then join them. Too much internals to be treated as easy.

bloodyKnuckles:
Reading back to krainboltgreene's comment about `div` is not default—yes, 3 years old—I realize a better way to describe this pull request is:

**1. (When not specified) Fallback to `DIV`—and 2. allow any selector order**

...since I agree, use semantic tagging as much possible—however when someone does not, I don't think throwing an error is the most effective solution.

The order of `ID`s and `CLASSES` may be better served in a separate pull request—which is not a matter of responding to lack of "best practice" like falling back to `DIV` when tag not specified, but rather a design flaw that is prone to confound the programmer why classes are ignored.

mightyiam:
@bloodyKnuckles thank you for expanding. Regarding selectors order, #157. Regarding falling back to `div`: semantically, I see it as a default and not a fallback, because fallback to me sounds like a choice in lieu of some examined factor, while default more appropriate for in lieu of input. So I changed the title.

Now, from the readme:

> A virtual DOM library with focus on simplicity, modularity, powerful features and performance.

I feel that such a default would clearly stray significantly away from the qualification of simplicity. Therefore I'll leave this issue closed.

Regarding a wrapping `h` function, @StreetStrider that provides this, here's a rough implementation:

```js
import { h as original } from 'snabbdom/h'
export const h = (sel, ...args) => {
  if (sel.startsWith('#') || sel.startsWith('.')) {
    return original('div' + sel, ...args)
  }
  return original(sel, ...args)
}
```

Do you feel this could work?

PR #119, opened 2016-05-25
Added build system and standalone modules

### LINKED ISSUE #109: Provide a standalone file
There's currently no way to include this library without browserify or bundling a standlone. It would be nice to have a prebundled version of the library available through git.

And wzrd.in can't be used because `snabbdom` has multiple entry points (ie `snabbdom/h`)

Risto-Stevcev:
resolves #109 
https://gist.github.com/Risto-Stevcev/d51f45997f1a818b6d21492475e07913
http://jsbin.com/jimecutaye/edit?html,output


grimalschi:
+1

paldepind:
This looks very good!

What about a bundle file that includes Snabbdom and the default modules and exports a `patch` function that uses those? I don't know if it's really necessary but it would make it possible to try it with only a single script tag.

Risto-Stevcev:
@paldepind Yeah, I thought about that initially. It would be good to have both, I'll add another PR.

amirouche:
Being able to use snabbdom using a single script tag will be nice.

PR #121, opened 2016-05-28
Adding modules on the fly with patch.extend

grimalschi:
Before it was only one way to add modules - only when you init snabbdom.

Now you can add modules on the fly, like this:

`var patch = snabbdom.init([snabbdom_class, snabbdom_attrs]);`
`patch.extend([{`
`create: function () {},`
`update: function () {},`
`}, {`
`create: function () {},`
`update: function () {},`
`}]);`

or

`var patch = snabbdom.init([snabbdom_class, snabbdom_attrs]);`
`patch.extend({`
`create: function () {},`
`update: function () {},`
`});`

It is very important if you want to make standalone bundle, which already includes all default modules, because in this case there is no way to add additional modules, except this `patch.extend` function.

### DIFF
--- a/snabbdom.js
+++ b/snabbdom.js
@@ -31,13 +31,6 @@ function init(modules, api) {
 
   if (isUndef(api)) api = domApi;
 
-  for (i = 0; i < hooks.length; ++i) {
-    cbs[hooks[i]] = [];
-    for (j = 0; j < modules.length; ++j) {
-      if (modules[j][hooks[i]] !== undefined) cbs[hooks[i]].push(modules[j][hooks[i]]);
-    }
-  }
-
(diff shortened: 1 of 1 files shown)


Hypnosphi:
How about making in non-mutating? Like `extendedPatch = patch.extend(modules)`, where `patch` stays the same

mightyiam:
Thank you for your contribution, @grimalschi.

Please correct me if I'm wrong. I see no benefit of this over discarding the existing `patch` function and making a new one with `init` again.

grimalschi:
I don't remember what was the case, honestly. I think we can close this.

mightyiam:
Thank you for the follow-up, @grimalschi.

PR #122, opened 2016-05-28
chore(TS): add typescript definitions

TylorS:
Add typescript definitions for consumption of other libraries and users that prefer typescript :)


paldepind:
This is really great! Supporting TypeScript is becoming more and more important. Thank you :+1:

TylorS:
Thanks for accepting these :)

PR #128, opened 2016-07-04
fix style example in README.md

### LINKED ISSUE #127: README has one tiny problem
In the *inline example* of README

`fontStyle: 'italics'`  is not working,  change to  `fontStyle: 'italic'` will work.

squiddle:
fixes #127


paldepind:
Thank you :+1:

PR #133, opened 2016-07-18
126 svg foreignobject

harrywincup:
Hi there,

This PR adds a new condition to the `addNS` function to halt recursion if it encounters the `foreignObject` SVG element. The result of this is that embedded HTML elements now retain the correct namespace (`http://www.w3.org/1999/xhtml`) and should render correctly inside the SVG parent.

I've appended a few new assertions to the existing namespace test case, but please let me know if there is a smarter way to do any of this! 

P.s I had some troubles running the tests whereby the page continuously reloads in both Firefox and Chrome. I was able to see the new tests passing/failing as i'd expect, but I'm not sure if there's something going wrong with the test runner?

Thanks 😄

### DIFF
--- a/h.js
+++ b/h.js
@@ -1,11 +1,12 @@
 var VNode = require('./vnode');
 var is = require('./is');
 
-function addNS(data, children) {
+function addNS(data, children, sel) {
   data.ns = 'http://www.w3.org/2000/svg';
-  if (children !== undefined) {
+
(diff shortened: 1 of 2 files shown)


paldepind:
Thank you for the PR. It looks great. I can't think of any smarter way to do it :smile:

Also you are right that the test runner sometimes keeps reloading. Unfortunately I have not been able to figure out why that is.

PR #144, opened 2016-07-28
Added details on using classes with SVG

sk1981:
Using SVG elements with classes has two issues due to className being readonly for these elements and some browsers not accepting classlist on them. Added details and workaround.


paldepind:
Really great! Thank you.

PR #146, opened 2016-07-29
New implementation of eventlisteners module

katyo:
Key differences:

* Full backward compatibility
* Improved adding/removing event listeners to/from real DOM node
* Using unified function as real event listener per node
* Access to virtual node from event handler through `this`
* Support for multiple listeners for same event
* No problems with shared handlers

### DIFF
--- a/modules/eventlisteners.js
+++ b/modules/eventlisteners.js
@@ -1,62 +1,79 @@
-var is = require('../is');
-
-function arrInvoker(arr) {
-  return function() {
-    if (!arr.length) return;
-    // Special case when length is two, for performance
-    arr.length === 2 ? arr[0](arr[1]) : arr[0].apply(undefined, arr.slice(1));
-  };
+function invokeHandler(handler, vnode, event) {
+  if (typeof handler === "function") {
+    // call function handler
+    handler.call(vnode, event);
+  } else if (typeof handler === "object") {
+    // call handler with arguments
+    if (typeof handler[0] === "function") {
+      // special case for single argument for performance
+      handler.length === 2 ?
+        handler[0].call(vnode, handler[1], event) :
+        handler[0].apply(vnode, handler.slice(1).concat(event));
+    } else {
+      // call multiple handlers
+      for (var i = 0; i < handler.length; i++) {
+        invokeHandler(handler[i]);
+      }
+    }
+  }
 }
 
-function fnInvoker(o) {
-  return function(ev) { 
-    if (o.fn === null) return;
-    o.fn(ev); 
-  };
+function handleEvent(event, vnode) {
+  var name = event.type,
+      on = vnode.data.on;
+
+  // call event handler(s) if exists
+  if (on && on[name]) {
+    invokeHandler(on[name], vnode, event);
+  }
+}
+
+function createListener() {
+  return function handler(event) {
+    handleEvent(event, handler.vnode);
+  }
 }
 
 function updateEventListeners(oldVnode, vnode) {
-  var name, cur, old, elm = vnode.elm,
-      oldOn = oldVnode.data.on, on = vnode.data.on;
+  var oldOn = oldVnode.data.on,
+      oldListener = oldVnode.listener,
+      oldElm = oldVnode.elm,
+      on = vnode && vnode.data.on,
+      elm = vnode && vnode.elm;
 
-  if (!on && !oldOn) return;
-  on = on || {};
-  oldOn = oldOn || {};
+  // improvement for immutable handlers
+  if (oldElm === elm && oldOn === on) {
+    return;
+  }
 
-  for (name in on) {
-    cur = on[name];
-    old = oldOn[name];
-    if (old === undefined) {
-      if (is.array(cur)) {
-        elm.addEventListener(name, arrInvoker(cur));
-      } else {
-        cur = {fn: cur};
-        on[name] = cur;
-        elm.addEventListener(name, fnInvoker(cur));
+  // remove existing listeners which no longer used
+  if (oldOn && oldListener) {
+    for (var _name in oldOn) {
+      // remove listener if element was changed or existing listener(s) removed
+      if (elm !== oldElm || !on || !on[_name]) {
+        oldElm.removeEventListener(_name, oldListener, false);
       }
-    } else if (is.array(old)) {
-      // Deliberately modify old array since it's captured in closure created with `arrInvoker`
-      old.length = cur.length;
-      for (var i = 0; i < old.length; ++i) old[i] = cur[i];
-      on[name]  = old;
-    } else {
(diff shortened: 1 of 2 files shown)


modules/eventlisteners.js
```diff
@@ -1,62 +1,79 @@
...
-        cur = {fn: cur};
-        on[name] = cur;
-        elm.addEventListener(name, fnInvoker(cur));
+  var oldOn = oldVnode.data.on,
+      oldListener = oldVnode.listener,
+      oldElm = oldVnode.elm,
+      on = vnode && vnode.data.on,
+      elm = vnode && vnode.elm;
+
+  // improvement for immutable handlers
```
paldepind: The comment probably shouldn't say "improvement" since that is a specific reference to the _old_ implementation.

modules/eventlisteners.js
```diff
@@ -1,62 +1,79 @@
...
+  if (oldElm === elm && oldOn === on) {
+    return;
+  }
+
+  // remove existing listeners which no longer used
+  if (oldOn && oldListener) {
+    for (var _name in oldOn) {
+      // remove listener if element was changed or existing listener(s) removed
+      if (elm !== oldElm || !on || !on[_name]) {
+        oldElm.removeEventListener(_name, oldListener, false);
```
paldepind: What about taking `elm !== oldElm || !on` and setting in a variable right before the `for` loop? Then we only make the checks once and get the opportunity to give it a name. Something like this:

```js
var removeAll = elm !== oldElm || on === undefined;
```

paldepind:
This looks really nice. Many improvements (y) I'd like to get this merged in.

I've left a few notes.

modules/eventlisteners.js
```diff
@@ -1,62 +1,79 @@
...
+function invokeHandler(handler, vnode, event) {
+  if (typeof handler === "function") {
+    // call function handler
+    handler.call(vnode, event);
+  } else if (typeof handler === "object") {
+    // call handler with arguments
+    if (typeof handler[0] === "function") {
+      // special case for single argument for performance
+      handler.length === 2 ?
+        handler[0].call(vnode, handler[1], event) :
```
sabberworm: Maybe pass the vnode both as the context and an argument (after `event`) so that people who use `bind` or arrow functions can access it too…

modules/eventlisteners.js
katyo: Yes, you're right.
I just added it later because I implemented this module from scratch.
I think this comment must looks like: optimization for immutable handlers.

modules/eventlisteners.js
katyo: Of course ;)

modules/eventlisteners.js
katyo: Initially I planning use the next sequence of arguments: `event`, `vnode`, `handler's arguments...`, but it breaks backward compatibility.

It seems nothing breaks if I simply add `vnode` as last argument.

katyo:
Can we do things more better?

paldepind:
I'll repeat: This PR is very great! I really enjoy the new features.

Initially I was a quite confused by `elmChanged` because when Snabbdom patches elements it always reuses the old element. I've however realized that when `elmChanged === true` then it is because `oldElm !== undefined || elm !== undefined` and `updateEventlisteners` is called by either the `create` hook or the `destroy` hook. But one thing that I can't figure out is why we need `elmChanged` in the first place? I've tried removing it like shown below:

```diff
       oldElm = oldVnode.elm,
       on = vnode && vnode.data.on,
       elm = vnode && vnode.elm,
-      elmChanged = oldElm !== elm,
       name;
 
   // optimization for immutable handlers
-  if (!elmChanged && oldOn === on) {
+  if (oldOn === on) {
     return;
   }
 
   // remove existing listeners which no longer used
   if (oldOn && oldListener) {
     // if element changed or deleted we remove all existing listeners unconditionally
-    if (elmChanged || !on) {
+    if (!on) {
       for (name in oldOn) {
         // remove listener if element was changed or existing listeners removed
         oldElm.removeEventListener(name, oldListener, false);
@@ -74,7 +73,7 @@ function updateEventListeners(oldVnode, vnode) {
     listener.vnode = vnode;
 
     // if element changed or added we add all needed listeners unconditionally
-    if (elmChanged || !oldOn) {
+    if (!oldOn) {
       for (name in on) {
         // add listener if element was changed or new listeners added
         elm.addEventListener(name, listener, false);
```

All the tests still pass?

Also since `vnode` shouldn't change why not make `vnode` and argument of `createListener`? Then the returned handler could simply keep the `vnode` in it's closure. This seems nicer to me than setting the `vnode` as a property of the returned function.

Also the "optimization for immutable handlers" is a great idea. But maybe it should instead say "optimization for reused immutable handlers" to indicate that the optimizations only apply when you reuse the exact same `on` object?

Finally both `slice` and `concat` creates new arrays. This means we end up creating two arrays in the line `handler[0].apply(vnode, handler.slice(1).concat(event, vnode));`. If we instead push the two additional elements as below then we only allocate a single array.
```diff
@@ -1,4 +1,5 @@
 function invokeHandler(handler, vnode, event) {
+  var arr;
   if (typeof handler === "function") {
     // call function handler
     handler.call(vnode, event, vnode);
@@ -6,9 +7,14 @@ function invokeHandler(handler, vnode, event) {
     // call handler with arguments
     if (typeof handler[0] === "function") {
       // special case for single argument for performance
-      handler.length === 2 ?
+      if (handler.length === 2) {
         handler[0].call(vnode, handler[1], event, vnode) :
-        handler[0].apply(vnode, handler.slice(1).concat(event, vnode));
+      } else {
+        arr = handler.slice(1);
+        arr.push(event);
+        arr.push(vnode);
+        handler[0].apply(vnode, arr);
+      }
     } else {
```

I'd be wonderful if you'd consider the above suggestions :smile: If you don't have the time to implement them I can merge the PR and do it.

katyo:
It seems I don't completely understand vdom. Do you mean that element cannot be changed never? In other words in case of changing tag (sel) it will destroy and create new element.

katyo:
As I understand `vnode` may be changed on each patch call. So we need to update it to provide right `vnode` for handler.

paldepind:
> It seems I don't completely understand vdom. Do you mean that element cannot be changed never? In other words in case of changing tag (sel) it will destroy and create new element.

Yes. That is correct.

> As I understand `vnode` may be changed on each patch call. So we need to update it to provide right `vnode` for handler.

You are right. I don't know what I was thinking :crying_cat_face: Of course the `vnode` have changed each time `updateEventListeners` is called.

Thank you for the last changes and the PR :+1: I'll merge it now and do a release.

paldepind:
It's now on npm. Thank you for the great wrok

PR #154, opened 2016-08-27
handle the case when the SVG element is a <text> and the child is a string

monfera:
Be able to handle the [<text>](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text) element so that SVG text can be rendered. Currently, an error is raised.

### DIFF
--- a/h.js
+++ b/h.js
@@ -2,8 +2,9 @@ var VNode = require('./vnode');
 var is = require('./is');
 
 function addNS(data, children, sel) {
-  data.ns = 'http://www.w3.org/2000/svg';
-
+  if (data !== undefined) {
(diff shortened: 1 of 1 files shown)


amirouche:
No test?

paldepind:
Thank you for the bugfix @monfera :smile: 

What about checking for the `text` property in the for-loop where `addNS` is called? Would that solve the issue as well?

Also, please add tests :+1:

jbucaran:
@paldepind Checking whether the children have data or not before calling `addNS` also works. 

Here is the fix https://github.com/snabbdom/snabbdom/pull/208

Unless I am missing something, svg is broken for me.

Issue #158, opened 2016-09-02
Using shadowRoot as the root node throws an exception


tusharmath:
**Code**
on creating [custom elements] —
[custom elements]: html5rocks.com/en/tutorials/webcomponents/customelements/
```js

createdCallback() {
  const root = this.shadowRoot()
  patch(root, h('div', ['Hello World']))
}

```

**Actual Behaviour:**
```error
Uncaught TypeError: Cannot read property 'toLowerCase' of undefined
```
Line: https://github.com/paldepind/snabbdom/blob/5b2d0de463d0de68db5b62c579031905131fe29e/snabbdom.js#L42

**Expected Behavior**
No exception

paldepind:
Hello @tusharmath. Thank you for reporting this issue :smile:

You mentioned that #159 might fix the issue for you. Did it change anything?

tusharmath:
@paldepind Nope :( This issue happens because `tagName` isn't available on shadowRoot components. We could use `nodeName` in that case.

The second issue is snabbdom looks for the parent node and tries to insert the new node into it. Again with the shadowRoot there is no parent. 

The solution that I have adopted is to create a dummy node inside the shadow root and send that to snabbdom for patching, this solves the above two issues. I think im fine by it and will so be closing this issue.

PR #159, opened 2016-09-04
Fix patch() so that the root is patched instead of recreated

staltz:
Previously, patch(element, vnode) would create a new element every time
for the top-level element, unless the top-level vnode had no id and
className. This is because emptyNodeAt() would create a vnode with sel
equal to the element's tagName, NOT including id and className.

This seems to be just a small fix, however this fix was crucial to get
Web Components (custom elements) working with Cycle.js, because the
custom element's "create" callback was being called infinitely and
recursively, because snabbdom was creating the top-level element every
time, even if the sel of oldVNode and vnode were supposed to be the
same.

### DIFF
--- a/test/core.js
+++ b/test/core.js
@@ -136,6 +136,17 @@ describe('snabbdom', function() {
         done();
       }
     });
+    it('is a patch of the root element', function () {
+      var elmWithIdAndClass = document.createElement('div');
+      elmWithIdAndClass.id = 'id';
+      elmWithIdAndClass.className = 'class';
+      var vnode1 = h('div#id.class', [h('span', 'Hi')]);
+      elm = patch(elmWithIdAndClass, vnode1).elm;
+      assert.strictEqual(elm, elmWithIdAndClass);
+      assert.equal(elm.tagName, 'DIV');
+      assert.equal(elm.id, 'id');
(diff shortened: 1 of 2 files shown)


test/core.js
```diff
@@ -670,7 +681,7 @@ describe('snabbdom', function() {
           {remove: function(_, rm) { rm2 = rm; }},
         ]);
         var vnode1 = h('div', [h('a', {hook: {remove: function(_, rm) { rm3 = rm; }}})]);
-	var vnode2 = h('div', []);
+        var vnode2 = h('div', []);
```
staltz: This was my editor that automatically converted the (previously) tab to spaces. But it's anyway good to avoid mixed tabs and spaces.

TylorS:
LGTM :+1:

tusharmath:
This would potentially fix #158

paldepind:
Thank you for the PR. To me this seems clearly better. I'd say the old behavior was pretty much a bug.

paldepind:
I've just released 0.5.3 with this patch. Thanks again. I'm sorry for the trouble caused but it never occurred to me how

staltz:
Thanks for merging 👍 the synergy between our communities is good

PR #161, opened 2016-09-07
Don't treat Object.prototype properties as boolean attributes

fix-fix:
Currently boolean element attributes have a special behavior: they are omitted if the value for the attribute key passed in an `attrs` object is falsy.

For performance reasons all the boolean attributes are stored in a native object that acts like a Map. However, objects constructed in a literal form have its prototype set to Object.prototype, so property lookup operation `booleanAttrsDict[key]` will also check for all the properties up on the prototype chain, including Object.prototype.
This makes the attributes with the same names as Object.prototype properties and methods (like `constructor`, `valueOf`, `toString`, etc.) behave like boolean attributes.

This change makes the boolean attributes dictionary prototype-less ([explanation](http://www.2ality.com/2013/10/dict-pattern.html)) to prevent false positives in checking whether the attribute is boolean or not.
I also added some basic tests for attributes module and for this issue in particular.

---

The other way to achieve the same result is to use `Object.prototype.hasOwnProperty` on a plain literal object, but I don't know which one has more efficient performance and how to benchmark it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ node_modules
 *.swp
 
 test/browserified.js
+browserified.js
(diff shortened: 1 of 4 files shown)


paldepind:
Great :smile: This seems like a very reasonable bugfix. Also, thank you for adding tests for the attributes module :+1:

Issue #162, opened 2016-09-08
is.primitive should check for all primitive values


Hypnosphi:
Now it checks for string and number only, which leads to unnecessary errors when null value is passed. Alternatively, one might check b for being an object, and set data to b only in that case
https://github.com/paldepind/snabbdom/blob/master/h.js#L22

kuraga:
Still buggy?

mightyiam:
It seems that this is fixed in https://github.com/snabbdom/snabbdom/commit/7ff8d13b4e5a037ac3a43519e746db26dc2c0173#diff-fecfa4a97c0ff7750b8c1e1549f7f3faR70-R75.

PR #163, opened 2016-09-08
Handle null and boolean arguments in h helper

valentin-nemcev:
Right now when you use short form `h('em', person.name)` and person name is `null`, you get `Uncaught TypeError: Cannot read property 'key' of null` in `vnode.js:2`. 
After this PR, `h` helper (called with 2 or 3 arguments) ignores `null` values and converts number and boolean values to strings.
This is to improve robustness of vnode helper and reduce need for null checks in client code.

### DIFF
--- a/is.js
+++ b/is.js
@@ -1,4 +1,6 @@
 module.exports = {
   array: Array.isArray,
-  primitive: function(s) { return typeof s === 'string' || typeof s === 'number'; },
+  primitive: function(s) {
+      return typeof s === 'string' || typeof s === 'number' || typeof s === 'boolean';
+  },
 };
--- a/h.js
+++ b/h.js
@@ -13,13 +13,13 @@ function addNS(data, children, sel) {
 
 module.exports = function h(sel, b, c) {
   var data = {}, children, text, i;
-  if (c !== undefined) {
+  if (c != null) {
     data = b;
     if (is.array(c)) { children = c; }
-    else if (is.primitive(c)) { text = c; }
-  } else if (b !== undefined) {
+    else if (is.primitive(c)) { text = String(c); }
+  } else if (b != null) {
     if (is.array(b)) { children = b; }
-    else if (is.primitive(b)) { text = b; }
+    else if (is.primitive(b)) { text = String(b); }
     else { data = b; }
   }
   if (is.array(children)) {
--- a/test/core.js
+++ b/test/core.js
@@ -54,6 +54,34 @@ describe('snabbdom', function() {
       var vnode = h('a', {}, 'I am a string');
       assert.equal(vnode.text, 'I am a string');
     });
+    it('can create vnode with props and null text content', function() {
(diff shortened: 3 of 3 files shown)


is.js
```diff
@@ -1,4 +1,6 @@
 module.exports = {
   array: Array.isArray,
-  primitive: function(s) { return typeof s === 'string' || typeof s === 'number'; },
+  primitive: function(s) {
+      return typeof s === 'string' || typeof s === 'number' || typeof s === 'boolean';
```
fix-fix: This still doesn't cover all primitive values (e.g. 'undefined', 'symbol') and returns `false` for `null`.

To check if the value is primitive you should check that it's not an object, like so: `(x) => !isObject(x)`, where `isObject` is something like [`_.isObject from lodash`](https://github.com/lodash/lodash/blob/8176d56c725489156e19d68aeaf8851d313f4446/dist/lodash.core.js#L2699).

Hypnosphi:
This is intended to fix #162

is.js
valentin-nemcev: I've tried to change as little as possible. 
Right now this method is used to answer the question "could we use this value as a content of a text node?" and for `null` or `undefined` implicit answer is no.

More precisely, [here](https://github.com/valentin-nemcev/snabbdom/blob/d338319cf6e409a1c318228c62d9943d1b73f931/snabbdom.js#L78) we don't expect `null` or `undefined`, [here](https://github.com/valentin-nemcev/snabbdom/blob/d338319cf6e409a1c318228c62d9943d1b73f931/h.js#L16) we check for them before, and [here](https://github.com/valentin-nemcev/snabbdom/blob/d338319cf6e409a1c318228c62d9943d1b73f931/h.js#L16) considering `null` and `undefined` as primitives will change edge case behaviour, where if you have these values in a list of children, snabbdom will insert text nodes for them instead of throwing exception.

I also want to make another PR to skip `null` and `undefined` values in a list of children, for same reasons as this PR. Then you could consider them primitive, so the method behaviour would be more aligned with its name.

I just didn't considered symbols, probably they could be included in the list.

paldepind:
Hello @valentin-nemcev. Thank you for the PR :+1: 

The changes to `h` certainly seems like something that could be useful in situations where one might have data that is potentially `null`.

I don't fully understand the change to `primitive` though. When would you want to pass Snabbdom a boolean and have it show up in the DOM as `true` or `false`?

valentin-nemcev:
Thank you for the review!

The idea was to make it behave like a real DOM text node, if you set it's value to boolean or other primitive (by mistake or while debugging), it will convert it to string.

mightyiam:
I, for one, don't like leniency in such a low level library as snabbdom.

mightyiam:
I believe that this is not desired behavior. I'll close this. But if anyone disagrees, please comment or create an issue.

PR #165, opened 2016-09-26
fixed links in README (paldepind -> snabbdom)

forceuser:


paldepind:
Great. Thank you :smile:

PR #173, opened 2016-10-03
Fix thunk example

jas-chen:


paldepind:
Thank you. Must have been missed when the API was changed.

Issue #123, opened 2016-05-30
null / undefined children


AlexGalays:
It's quite convenient to be able to use null or undefined children, e.g:

```javascript
const popup = popupOpened ? popup() : undefined

return h('div', [
  h('p', 'blabla'),
  popup
])
```

virtual-dom and React support that for instance. Would it be a lot of work to add this in ? Would it mess with reconciliation indexes?

AlexGalays:
string children fill that role.

grimalschi:
yeah, const popup = popupOpened ? popup() : ''

Hypnosphi:
What's the problem though to support null/undefined as well?

PR #174, opened 2016-10-06
Make debugging undefined children easier

jayrbolton:
I find this error to be a very common case: `Cannot read property 'data' of undefined` which occurs when a child of a vnode is undefined during the patch process.

You can never use the stack trace to see where it is coming from in your original code where you construct the tree with `h()` functions. This is because the error arises during the patch process, and no part of the execution path refers to where the `h()` function was originally called. I find this hard to debug

To make the stack trace useful for this error, I added a check at the time the `h()` function is called that will throw a type error if any children are undefined. For sparse arrays, I don't have a good type check for those, so instead I filter them into non-sparse arrays.

Example instances that cause this untraceable error, which can be hard to see just by looking:

`h('div', [,h('p', 'Hi'])`

`h('div', ['hello, ', state.greeting] // state.greeting happens to be undefined or null`

### DIFF
--- a/h.js
+++ b/h.js
@@ -23,6 +23,10 @@ module.exports = function h(sel, b, c) {
     else { data = b; }
   }
   if (is.array(children)) {
+    children = children.filter(function() {return true}); // Eliminate possible sparse array
+    if(children.filter(function(c) { return !is.primitive(c) && typeof c !== 'object' || c === null}).length) {
+      throw new TypeError("VNode children array must contain Strings, Numbers, or other VNodes.")
+    }
     for (i = 0; i < children.length; ++i) {
       if (is.primitive(children[i])) children[i] = VNode(undefined, undefined, undefined, children[i]);
     }
--- a/test/core.js
+++ b/test/core.js
@@ -54,6 +54,12 @@ describe('snabbdom', function() {
       var vnode = h('a', {}, 'I am a string');
       assert.equal(vnode.text, 'I am a string');
     });
+    it('throws a type error when the child array has an undefined value', function() {
+      assert.throws(function() { h('div', [h('p'), undefined]) }, TypeError);
+    })
+    it('throws a type error when the child array has a null value', function() {
+      assert.throws(function() { h('div', [h('p'), null]) }, TypeError);
+    })
   });
   describe('created element', function() {
     it('has tag', function() {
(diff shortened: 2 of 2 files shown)


jayrbolton:
Kinda sorta related: 
- https://github.com/snabbdom/snabbdom/issues/123
- https://github.com/snabbdom/snabbdom/pull/163
-https://github.com/snabbdom/snabbdom/issues/162

paldepind:
Overall i like this idea 😄 I'm wondering how infeasible it would be to log the virtual DOM tree you're giving `patch` and see where the `undefined` value occurred? Would that still be too bad with regards to debugging?

It is interesting that #123 was a wish for `null`/`undefined` values to simply be ignored. But you're saying that for you `null`/`undefined` values sneak in by mistake so you'd like them to throw. Personally that resonates a lot better with me. I've never wanted to leave in `undefined` intentionally but I have done so accidentally.

I don't know where that leaves us with regards to #163 which is about ignoring `null` as the third argument to `h`.

I'm worried about the implementation. `filter` is called twice which copies the `children` array twice. That is not good for performance. Also having `throw` directly inside `h` prevents `h` from being optimized. That also is not so good for performance.

jayrbolton:
Yeah, I'm in favor of an early type error rather than ignoring/coercing, as I'd rather find errors than have "null" and "undefined" silently printed out in apps. I definitely agree about the performance concerns and  was thinking that perhaps a type check could instead go in the VNode() constructor -- I will play around with that approach instead.

paldepind:
> Yeah, I'm in favor of an early type error rather than ignoring/coercing, as I'd rather find errors than have "null" and "undefined" silently printed out in apps.

What was requested previously is that `null`/`undefined` was simply ignored so that they neither caused errors nor showed up in the DOM. So that if users have a property that may be `null` they want to pass the `null` to Snabbdom and have it ignored. But as mentioned I prefer your proposal.

Let me know how it goes with `VNode`. I'm not sure what you'd do in the example you mentioned originally where the `undefined` is where a string should have been.

AlexGalays:
I'm not fond of the perf penalty at all :( 
Some Arrays are quite big (even with virtualization, etc) and it would be a shame to pay the price for what is a development edge case.

Also, when using typescript, that kind of check is completely unnecessary. Sparse arrays? I never encountered one, ever. A sparse Array should be made non sparse before reaching a render routine I think instead of being repeatedly filtered.

Another option would be for snabbdom to accept null/undefined children . They would work like the empty string and be ignored. What do you think ?

PR #178, opened 2016-10-13
add «snabbdom in the wild» section

amirouche:
tx!


paldepind:
Hello @amirouche. Adding Vue is a great idea. But there already is a section called "Structuring applications". Couldn't we merge this new section with that section under a common name?

amirouche:
Héllo @paldepind, it's done :)

PR #180, opened 2016-10-17
Support namespaced attributes

mkaemmerer:
Per https://github.com/cyclejs/cyclejs/issues/461, this update uses setAttributeNS so that attributes such as "xlink:href" can be used safely.

### DIFF
--- a/modules/attributes.js
+++ b/modules/attributes.js
@@ -16,7 +16,7 @@ for(var i=0, len = booleanAttrs.length; i < len; i++) {
 
 function updateAttrs(oldVnode, vnode) {
   var key, cur, old, elm = vnode.elm,
-      oldAttrs = oldVnode.data.attrs, attrs = vnode.data.attrs, namespace;
+      oldAttrs = oldVnode.data.attrs, attrs = vnode.data.attrs, namespaceSplit;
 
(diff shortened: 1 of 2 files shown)


paldepind:
Thank you for the PR. It looks great and is a welcome addition :smile:

The only small improvement I can think of is that we could avoid looking for `:` twice. Both `key.split(":")` and `key.match(/:/)` does. Instead what about something like this:

```js
namespaceSplit = key.split(":");
...
} else if (namespaceSplit.length > 1 && NamespaceURIs.hasOwnProperty(namespace)) {
  elm.setAttributeNS(NamespaceURIs[namespaceSplit[0], key, cur);
}
```

Then we're only figuring out the presence of `:` once.

Additionally it seems to me like the initialization of `namespace` (or `namespaceSplit`) could be moved to after the line `if (old !== cur) {`. Then the search for `:` is completely avoided if nothing has changed. We could even move it into the `else` clause after the check for whether or not the key is in `booleanAttrsDict`. Then we'd execute the search in even fewer cases.

mkaemmerer:
Those sound like good suggestions to me! I'll update in a minute.

paldepind:
Thank you :+1: Merged and released :smile:

PR #184, opened 2016-10-25
Remove unused non-sameVnode checking in patchVnode

zhulongzheng:
It seems that every vnode pair passed to `patchVnode` should satisfy the `sameVnode` checking.

### DIFF
--- a/snabbdom.js
+++ b/snabbdom.js
@@ -195,13 +195,6 @@ function init(modules, api) {
     }
     var elm = vnode.elm = oldVnode.elm, oldCh = oldVnode.children, ch = vnode.children;
     if (oldVnode === vnode) return;
-    if (!sameVnode(oldVnode, vnode)) {
-      var parentElm = api.parentNode(oldVnode.elm);
-      elm = createElm(vnode, insertedVnodeQueue);
-      api.insertBefore(parentElm, elm, oldVnode.elm);
-      removeVnodes(parentElm, [oldVnode], 0, 0);
-      return;
-    }
     if (isDef(vnode.data)) {
       for (i = 0; i < cbs.update.length; ++i) cbs.update[i](oldVnode, vnode);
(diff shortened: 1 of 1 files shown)


nunocastromartins:
I don't think that is true [here](https://github.com/snabbdom/snabbdom/blob/master/snabbdom.js#L176). Please confirm.

zhulongzheng:
Ok, the `elmToMove` and `newStartVnode` of this [line](https://github.com/snabbdom/snabbdom/blob/master/snabbdom.js#L176) should have the same `key`, but maybe different `sel`. So there are possibilities that the vnode pair would not satisfy the `sameVnode` checking.

But I think there is another bug if it gets into this [checking](https://github.com/snabbdom/snabbdom/blob/master/snabbdom.js#L198).

A newly created element is inserted into the parent element [here](https://github.com/snabbdom/snabbdom/blob/master/snabbdom.js#L200).
```
elm = createElm(vnode, insertedVnodeQueue);
api.insertBefore(parentElm, elm, oldVnode.elm);
```

And `oldVnode` is removed from it's parent element [here](https://github.com/snabbdom/snabbdom/blob/master/snabbdom.js#L202).
```
removeVnodes(parentElm, [oldVnode], 0, 0);
```

We should not insert the `oldVnode.elm`(`elmToMove.elm`) again [here](https://github.com/snabbdom/snabbdom/blob/master/snabbdom.js#L178).
```
api.insertBefore(parentElm, elmToMove.elm, oldStartVnode.elm);
```


So maybe I should close this pr, and create another one?

nunocastromartins:
I agree with you, I think that checking you wanted to delete should be moved to inside that `else` in the `updateChildren` function.

I think you can just use this PR. Maybe change the title?
I'm not associated with this repo in any way, though. :stuck_out_tongue:

PR #187, opened 2016-10-31
updateStyle: handle falsy style values

mightyiam:


modules/style.js
```diff
@@ -16,7 +16,7 @@ function updateStyle(oldVnode, vnode) {
   var oldHasDel = 'delayed' in oldStyle;
 
   for (name in oldStyle) {
-    if (!style[name]) {
+    if (style[name] === undefined) {
```
mightyiam: I don't think that we really expect intentionally falsy values here, like, `false` or `null`, right?

test/style.js
```diff
@@ -71,6 +71,13 @@ describe('style', function() {
     fakeRaf.step();
     assert.equal(elm.style.fontSize, '20px');
   });
+  it('handles falsy values gracefully', function() {
+    var vnode1 = h('i', {style: {top: 0}});
+    var vnode2 = h('i', {style: {top: 0}});
+    elm = patch(vnode0, vnode1).elm;
+    elm = patch(vnode1, vnode2).elm;
+    assert.equal(elm.style.top, '0px');
+  });
```
mightyiam: Is this test concise?

mightyiam:
I'll withdraw this. The values should be strings.

PR #188, opened 2016-11-01
Fix bug of updating one child with same key but different sel.

zhulongzheng:
I've created a [pull request](https://github.com/snabbdom/snabbdom/pull/184) days before, but found that is not complete. And the [vdom code of vue](https://github.com/vuejs/vue/blob/dev/src/core/vdom/patch.js#L308) should explain this problem.


zhulongzheng:
I've updated this pull request to make it can be merged with the last typescript code.

paldepind:
Definitely a great change. I've never changed the selector of an element with a key property myself. But handling it in this way is elegant and does what one would expect! :+1:

PR #191, opened 2016-11-06
Allow h(sel, data, node) and h(sel, node) shortcut notations

AlexGalays:
For two reasons:

- Convenience (virtual-dom does it too)
- As it stands, it's very hard (impossible if you account for people's potential custom modules) to type the `h` function correctly as typescript uses structural typing and the data object being lots of optional things (it will typecheck just fine if you specify them, but the compiler is clueless if you don't), an Array or VNode could pass for an empty data object just fine. If we can't encode it, allow it!

### DIFF
--- a/h.js
+++ b/h.js
@@ -17,9 +17,11 @@ module.exports = function h(sel, b, c) {
     data = b;
     if (is.array(c)) { children = c; }
     else if (is.primitive(c)) { text = c; }
+    else if (c && c.sel) { children = [c]; }
   } else if (b !== undefined) {
     if (is.array(b)) { children = b; }
     else if (is.primitive(b)) { text = b; }
+    else if (b && b.sel) { children = [b]; }
     else { data = b; }
   }
   if (is.array(children)) {
--- a/test/core.js
+++ b/test/core.js
@@ -42,6 +42,16 @@ describe('snabbdom', function() {
       assert.equal(vnode.children[0].sel, 'span#hello');
       assert.equal(vnode.children[1].sel, 'b.world');
     });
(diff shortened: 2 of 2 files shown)


paldepind:
I think this is a fine improvement. I don't see any downsides.

But, can't the current `h` function be typed with overloads? Something roughly like this:

```typescript
type Children = VNode[] | string | number;

function h(selector: String, children: Children);
function h(selector: String, props: Object, children: Children);
function h(sel, b, c) {
  ...
}
```

staltz:
Yes it could have overloads. I'm currently working on a PR to move the source code to TypeScript, I'll submit it soon (and it has overloads for `h`). The PR is just the minimal amount of work to transfer to TypeScript, it's not an overhaul.

AlexGalays:
That kind of typing is actually ambiguous.

If someone tries to write (as of today, by mistake. It happened to a coworker of mine as he used virtual-dom before)

`h('div', someVnode)` then the following typing will be selected:

`function h(selector: string, data: VNodeData)`

It works because data is very polymorphic. You might have props, and/or attrs, some custom module, or nothing at all (everything is optional). Thus, a `VNode` is also a valid VNodeData/empty object.

it will compile but you will not see the child.

You could type VNodeData like this: "an object that at least props or attrs, or..." but that only works if we don't account for people's custom modules.

staltz:
By the way, with new features in a PR, tests are vital.

AlexGalays:
Tests added :)

AlexGalays:
Hello @paldepind! Anything I can do to to get this merged? :)

Issue #150, opened 2016-08-16
Proposal: Port to TypeScript


TylorS:
Hello @paldepind, as I know you're aware, we are making use of this fantastic library in Cycle.js and Motorcycle.js has been using it since about November or December. We've been more and more embracing TypeScript and find that the only thing that is difficult is libraries which we are dependent on, not having or having out-of-date typings, which is why I had sent a PR for type definitions previously. However maintaining type-definitions is tedious, painfully error-prone, and now I have many spread across multiple projects that I try my best to keep up-to-date, but has been difficult.

I've been working on reviving some neglected parts of Motorcycle, porting to TypeScript, and I've been doing my best to ensure that all packages are tree-shaking compatible providing ES2015 builds alongside the commonjs builds. This is where I've ran into issues.

Compiling a TypeScript library/application to ES2015 does not allow for commonjs `module.exports` (Technically `exports.prop` will not work either, but there _are_ workarounds). For type definitions to be correct, they use the following syntax

``` typescript
declare namespace "snabbdom/h" {
  export = // ...
}
```

Which restricts TypeScript to using `import h = require('snabbdom/h')` but this disallows compiling to ES2015, and thus disallows tree-shaking compatibility.

So I propose moving the current codebase, starting with a 1:1 port, to TypeScript. Type Definitions will be consistent 100% of the time for users, and are properly versioned with snabbdom itself. Other benefits include ES2015 builds for tree shaking and standards compliance. Another benefit, if tests are written in TypeScript, typings can also be tested.

I would more than happily do this work myself and submit it as a PR, if this is desired.

TylorS:
Sooo, I went ahead and did this work as I mentioned I would, but for the moment its in a separate repo. All of the same tests are passing, and performance within the vdom-benchmark (100 iterations) are not too far from the current codebase, which is better than I expected to be honest. I had much less faith in the compiled output. I tested the commonjs builds of course, there could be a difference in the ES2015 output since there is less difference between TS and ES6.

![snabbom-ts](https://cloud.githubusercontent.com/assets/6282473/17762009/14153494-64da-11e6-9ace-a8093ec2a2a5.png)

https://github.com/TylorS/snabbdom-ts
https://travis-ci.org/TylorS/snabbdom-ts

paldepind:
I like TypeScript so I think a port is a splendid idea! :+1: 

I think we should just change the API of `thunk` as discussed in #143. Making it compatible with TypeScript is yet another benefit.

[...]

PR #193, opened 2016-11-08
Migrate source code to TypeScript v2.0

staltz:
As an effort to solve #150, this PR does a *conservative* migration of snabbdom's JavaScript codebase to TypeScript. The benefits with TypeScript for this project are:

- The type definitions will be more reliable and more accurate with the source code, serving TypeScript users well
- Helps catch some bugs in the codebase itself

To keep this PR as small as possible (since it's anyway a huge thing to migrate to TS) and to make sure all tests pass and performance is the same, I basically just converted JS to TS in the least intrusive way as possible. The changes we have here are:

- The source code is now under the folder `src/` and all the files are TypeScript
- In many cases I just cast to `any` (the alternative would have involved a larger refactor, so I avoided that)
- The JavaScript that used to live in the top-level folder, such as `h.js` and `snabbdom.js` etc are now generated from TypeScript, so they were removed from git. They will be always generated before npm publish, so there's no chance of forgetting to compile it. We do like this in xstream codebase.
- All tests pass, so it means this should have zero breaking changes. This PR needs to support the previous API entirely

The generated JavaScript looks very much like the hand-coded JavaScript we had before. To make it easier to compare, you can click on each of these below to see the diff.

 
<details>
  <summary>&#9654; h.js</summary>

```diff
diff --git a/h.js b/h.js
index ac07850..2b892b8 100644
--- a/h.js
+++ b/h.js
@@ -1,34 +1,47 @@
-var VNode = require('./vnode');
+"use strict";
+var vnode = require('./vnode');
 var is = require('./is');
-
 function addNS(data, children, sel) {
   data.ns = 'http://www.w3.org/2000/svg';
-
   if (sel !== 'foreignObject' && children !== undefined) {
     for (var i = 0; i < children.length; ++i) {
       addNS(children[i].data, children[i].children, children[i].sel);
     }
   }
 }
-
-module.exports = function h(sel, b, c) {
+function h(sel, b, c) {
   var data = {}, children, text, i;
   if (c !== undefined) {
     data = b;
-    if (is.array(c)) { children = c; }
-    else if (is.primitive(c)) { text = c; }
-  } else if (b !== undefined) {
-    if (is.array(b)) { children = b; }
-    else if (is.primitive(b)) { text = b; }
-    else { data = b; }
+    if (is.array(c)) {
+      children = c;
+    }
+    else if (is.primitive(c)) {
+      text = c;
+    }
+  }
+  else if (b !== undefined) {
+    if (is.array(b)) {
+      children = b;
+    }
+    else if (is.primitive(b)) {
+      text = b;
+    }
+    else {
+      data = b;
+    }
   }
   if (is.array(children)) {
     for (i = 0; i < children.length; ++i) {
-      if (is.primitive(children[i])) children[i] = VNode(undefined, undefined, undefined, children[i]);
+      if (is.primitive(children[i]))
+        children[i] = vnode(undefined, undefined, undefined, children[i]);
     }
   }
   if (sel[0] === 's' && sel[1] === 'v' && sel[2] === 'g') {
     addNS(data, children, sel);
   }
-  return VNode(sel, data, children, text, undefined);
-};
+  return vnode(sel, data, children, text, undefined);
+}
+;
+module.exports = h;
+//# sourceMappingURL=h.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; htmldomapi.js</summary>

```diff
diff --git a/htmldomapi.js b/htmldomapi.js
index c73f29a..c6b9044 100644
--- a/htmldomapi.js
+++ b/htmldomapi.js
@@ -1,54 +1,44 @@
-function createElement(tagName){
+"use strict";
+function createElement(tagName) {
   return document.createElement(tagName);
 }
-
-function createElementNS(namespaceURI, qualifiedName){
+function createElementNS(namespaceURI, qualifiedName) {
   return document.createElementNS(namespaceURI, qualifiedName);
 }
-
-function createTextNode(text){
+function createTextNode(text) {
   return document.createTextNode(text);
 }
-
-
-function insertBefore(parentNode, newNode, referenceNode){
+function insertBefore(parentNode, newNode, referenceNode) {
   parentNode.insertBefore(newNode, referenceNode);
 }
-
-
-function removeChild(node, child){
+function removeChild(node, child) {
   node.removeChild(child);
 }
-
-function appendChild(node, child){
+function appendChild(node, child) {
   node.appendChild(child);
 }
-
-function parentNode(node){
+function parentNode(node) {
   return node.parentElement;
 }
-
-function nextSibling(node){
+function nextSibling(node) {
   return node.nextSibling;
 }
-
-function tagName(node){
-  return node.tagName;
+function tagName(elm) {
+  return elm.tagName;
 }
-
-function setTextContent(node, text){
+function setTextContent(node, text) {
   node.textContent = text;
 }
-
 module.exports = {
   createElement: createElement,
   createElementNS: createElementNS,
   createTextNode: createTextNode,
-  appendChild: appendChild,
-  removeChild: removeChild,
   insertBefore: insertBefore,
+  removeChild: removeChild,
+  appendChild: appendChild,
   parentNode: parentNode,
   nextSibling: nextSibling,
   tagName: tagName,
-  setTextContent: setTextContent
+  setTextContent: setTextContent,
 };
+//# sourceMappingURL=htmldomapi.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; is.js</summary>

```diff
diff --git a/is.js b/is.js
index c228508..1b5961c 100644
--- a/is.js
+++ b/is.js
@@ -1,4 +1,8 @@
+"use strict";
 module.exports = {
   array: Array.isArray,
-  primitive: function(s) { return typeof s === 'string' || typeof s === 'number'; },
+  primitive: function primitive(s) {
+    return typeof s === 'string' || typeof s === 'number';
+  },
 };
+//# sourceMappingURL=is.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; snabbdom.bundle.js</summary>

```diff
diff --git a/snabbdom.bundle.js b/snabbdom.bundle.js
index 2f26dec..01958b5 100644
--- a/snabbdom.bundle.js
+++ b/snabbdom.bundle.js
@@ -1,11 +1,17 @@
+"use strict";
 var snabbdom = require('./snabbdom');
-var patch = snabbdom.init([ // Init patch function with choosen modules
-  require('./modules/attributes'), // makes it easy to toggle classes
-  require('./modules/class'), // makes it easy to toggle classes
-  require('./modules/props'), // for setting properties on DOM elements
-  require('./modules/style'), // handles styling on elements with support for animations
-  require('./modules/eventlisteners'), // attaches event listeners
-]);
+var attributesModule = require('./modules/attributes'); // for setting attributes on DOM elements
+var classModule = require('./modules/class'); // makes it easy to toggle classes
+var propsModule = require('./modules/props'); // for setting properties on DOM elements
+var styleModule = require('./modules/style'); // handles styling on elements with support for animations
+var eventListenersModule = require('./modules/eventlisteners'); // attaches event listeners
 var h = require('./h'); // helper function for creating vnodes
-
-module.exports = { patch: patch, h: h }
+var patch = snabbdom.init([
+  attributesModule,
+  classModule,
+  propsModule,
+  styleModule,
+  eventListenersModule
+]);
+module.exports = { patch: patch, h: h };
+//# sourceMappingURL=snabbdom.bundle.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; snabbdom.js</summary>

```diff
diff --git a/snabbdom.js b/snabbdom.js
index 1fcc2ed..84443d0 100644
--- a/snabbdom.js
+++ b/snabbdom.js
@@ -1,58 +1,48 @@
-// jshint newcap: false
-/* global require, module, document, Node */
-'use strict';
-
-var VNode = require('./vnode');
+"use strict";
+var vnode = require('./vnode');
 var is = require('./is');
-var domApi = require('./htmldomapi');
-
+var htmlDomApi = require('./htmldomapi');
 function isUndef(s) { return s === undefined; }
 function isDef(s) { return s !== undefined; }
-
-var emptyNode = VNode('', {}, [], undefined, undefined);
-
+var emptyNode = vnode('', {}, [], undefined, undefined);
 function sameVnode(vnode1, vnode2) {
   return vnode1.key === vnode2.key && vnode1.sel === vnode2.sel;
 }
-
 function createKeyToOldIdx(children, beginIdx, endIdx) {
   var i, map = {}, key;
   for (i = beginIdx; i <= endIdx; ++i) {
     key = children[i].key;
-    if (isDef(key)) map[key] = i;
+    if (isDef(key))
+      map[key] = i;
   }
   return map;
 }
-
 var hooks = ['create', 'update', 'remove', 'destroy', 'pre', 'post'];
-
-function init(modules, api) {
+function init(modules, domApi) {
   var i, j, cbs = {};
-
-  if (isUndef(api)) api = domApi;
-
+  var api = domApi;
+  if (isUndef(domApi))
+    api = htmlDomApi;
   for (i = 0; i < hooks.length; ++i) {
     cbs[hooks[i]] = [];
     for (j = 0; j < modules.length; ++j) {
-      if (modules[j][hooks[i]] !== undefined) cbs[hooks[i]].push(modules[j][hooks[i]]);
+      if (modules[j][hooks[i]] !== undefined)
+        cbs[hooks[i]].push(modules[j][hooks[i]]);
     }
   }
-
   function emptyNodeAt(elm) {
     var id = elm.id ? '#' + elm.id : '';
     var c = elm.className ? '.' + elm.className.split(' ').join('.') : '';
-    return VNode(api.tagName(elm).toLowerCase() + id + c, {}, [], undefined, elm);
+    return vnode(api.tagName(elm).toLowerCase() + id + c, {}, [], undefined, elm);
   }
-
   function createRmCb(childElm, listeners) {
-    return function() {
+    return function rmCb() {
       if (--listeners === 0) {
-        var parent = api.parentNode(childElm);
-        api.removeChild(parent, childElm);
+        var parent_1 = api.parentNode(childElm);
+        api.removeChild(parent_1, childElm);
       }
     };
   }
-
   function createElm(vnode, insertedVnodeQueue) {
     var i, data = vnode.data;
     if (isDef(data)) {
@@ -70,39 +60,46 @@ function init(modules, api) {
       var dot = dotIdx > 0 ? dotIdx : sel.length;
       var tag = hashIdx !== -1 || dotIdx !== -1 ? sel.slice(0, Math.min(hash, dot)) : sel;
       elm = vnode.elm = isDef(data) && isDef(i = data.ns) ? api.createElementNS(i, tag)
-                                                          : api.createElement(tag);
-      if (hash < dot) elm.id = sel.slice(hash + 1, dot);
-      if (dotIdx > 0) elm.className = sel.slice(dot + 1).replace(/\./g, ' ');
+        : api.createElement(tag);
+      if (hash < dot)
+        elm.id = sel.slice(hash + 1, dot);
+      if (dotIdx > 0)
+        elm.className = sel.slice(dot + 1).replace(/\./g, ' ');
       if (is.array(children)) {
         for (i = 0; i < children.length; ++i) {
           api.appendChild(elm, createElm(children[i], insertedVnodeQueue));
         }
-      } else if (is.primitive(vnode.text)) {
+      }
+      else if (is.primitive(vnode.text)) {
         api.appendChild(elm, api.createTextNode(vnode.text));
       }
-      for (i = 0; i < cbs.create.length; ++i) cbs.create[i](emptyNode, vnode);
+      for (i = 0; i < cbs.create.length; ++i)
+        cbs.create[i](emptyNode, vnode);
       i = vnode.data.hook; // Reuse variable
       if (isDef(i)) {
-        if (i.create) i.create(emptyNode, vnode);
-        if (i.insert) insertedVnodeQueue.push(vnode);
+        if (i.create)
+          i.create(emptyNode, vnode);
+        if (i.insert)
+          insertedVnodeQueue.push(vnode);
       }
-    } else {
+    }
+    else {
       elm = vnode.elm = api.createTextNode(vnode.text);
     }
     return vnode.elm;
   }
-
   function addVnodes(parentElm, before, vnodes, startIdx, endIdx, insertedVnodeQueue) {
     for (; startIdx <= endIdx; ++startIdx) {
       api.insertBefore(parentElm, createElm(vnodes[startIdx], insertedVnodeQueue), before);
     }
   }
-
   function invokeDestroyHook(vnode) {
     var i, j, data = vnode.data;
     if (isDef(data)) {
-      if (isDef(i = data.hook) && isDef(i = i.destroy)) i(vnode);
-      for (i = 0; i < cbs.destroy.length; ++i) cbs.destroy[i](vnode);
+      if (isDef(i = data.hook) && isDef(i = i.destroy))
+        i(vnode);
+      for (i = 0; i < cbs.destroy.length; ++i)
+        cbs.destroy[i](vnode);
       if (isDef(i = vnode.children)) {
         for (j = 0; j < vnode.children.length; ++j) {
           invokeDestroyHook(vnode.children[j]);
@@ -110,28 +107,29 @@ function init(modules, api) {
       }
     }
   }
-
   function removeVnodes(parentElm, vnodes, startIdx, endIdx) {
     for (; startIdx <= endIdx; ++startIdx) {
-      var i, listeners, rm, ch = vnodes[startIdx];
+      var i_1 = void 0, listeners = void 0, rm = void 0, ch = vnodes[startIdx];
       if (isDef(ch)) {
         if (isDef(ch.sel)) {
           invokeDestroyHook(ch);
           listeners = cbs.remove.length + 1;
           rm = createRmCb(ch.elm, listeners);
-          for (i = 0; i < cbs.remove.length; ++i) cbs.remove[i](ch, rm);
-          if (isDef(i = ch.data) && isDef(i = i.hook) && isDef(i = i.remove)) {
-            i(ch, rm);
-          } else {
+          for (i_1 = 0; i_1 < cbs.remove.length; ++i_1)
+            cbs.remove[i_1](ch, rm);
+          if (isDef(i_1 = ch.data) && isDef(i_1 = i_1.hook) && isDef(i_1 = i_1.remove)) {
+            i_1(ch, rm);
+          }
+          else {
             rm();
           }
-        } else { // Text node
+        }
+        else {
           api.removeChild(parentElm, ch.elm);
         }
       }
     }
   }
-
   function updateChildren(parentElm, oldCh, newCh, insertedVnodeQueue) {
     var oldStartIdx = 0, newStartIdx = 0;
     var oldEndIdx = oldCh.length - 1;
@@ -141,37 +139,44 @@ function init(modules, api) {
     var newStartVnode = newCh[0];
     var newEndVnode = newCh[newEndIdx];
     var oldKeyToIdx, idxInOld, elmToMove, before;
-
     while (oldStartIdx <= oldEndIdx && newStartIdx <= newEndIdx) {
       if (isUndef(oldStartVnode)) {
         oldStartVnode = oldCh[++oldStartIdx]; // Vnode has been moved left
-      } else if (isUndef(oldEndVnode)) {
+      }
+      else if (isUndef(oldEndVnode)) {
         oldEndVnode = oldCh[--oldEndIdx];
-      } else if (sameVnode(oldStartVnode, newStartVnode)) {
+      }
+      else if (sameVnode(oldStartVnode, newStartVnode)) {
         patchVnode(oldStartVnode, newStartVnode, insertedVnodeQueue);
         oldStartVnode = oldCh[++oldStartIdx];
         newStartVnode = newCh[++newStartIdx];
-      } else if (sameVnode(oldEndVnode, newEndVnode)) {
+      }
+      else if (sameVnode(oldEndVnode, newEndVnode)) {
         patchVnode(oldEndVnode, newEndVnode, insertedVnodeQueue);
         oldEndVnode = oldCh[--oldEndIdx];
         newEndVnode = newCh[--newEndIdx];
-      } else if (sameVnode(oldStartVnode, newEndVnode)) { // Vnode moved right
+      }
+      else if (sameVnode(oldStartVnode, newEndVnode)) {
         patchVnode(oldStartVnode, newEndVnode, insertedVnodeQueue);
         api.insertBefore(parentElm, oldStartVnode.elm, api.nextSibling(oldEndVnode.elm));
         oldStartVnode = oldCh[++oldStartIdx];
         newEndVnode = newCh[--newEndIdx];
-      } else if (sameVnode(oldEndVnode, newStartVnode)) { // Vnode moved left
+      }
+      else if (sameVnode(oldEndVnode, newStartVnode)) {
         patchVnode(oldEndVnode, newStartVnode, insertedVnodeQueue);
         api.insertBefore(parentElm, oldEndVnode.elm, oldStartVnode.elm);
         oldEndVnode = oldCh[--oldEndIdx];
         newStartVnode = newCh[++newStartIdx];
-      } else {
-        if (isUndef(oldKeyToIdx)) oldKeyToIdx = createKeyToOldIdx(oldCh, oldStartIdx, oldEndIdx);
+      }
+      else {
+        if (isUndef(oldKeyToIdx))
+          oldKeyToIdx = createKeyToOldIdx(oldCh, oldStartIdx, oldEndIdx);
         idxInOld = oldKeyToIdx[newStartVnode.key];
-        if (isUndef(idxInOld)) { // New element
+        if (isUndef(idxInOld)) {
           api.insertBefore(parentElm, createElm(newStartVnode, insertedVnodeQueue), oldStartVnode.elm);
           newStartVnode = newCh[++newStartIdx];
-        } else {
+        }
+        else {
           elmToMove = oldCh[idxInOld];
           patchVnode(elmToMove, newStartVnode, insertedVnodeQueue);
           oldCh[idxInOld] = undefined;
@@ -181,20 +186,21 @@ function init(modules, api) {
       }
     }
     if (oldStartIdx > oldEndIdx) {
-      before = isUndef(newCh[newEndIdx+1]) ? null : newCh[newEndIdx+1].elm;
+      before = isUndef(newCh[newEndIdx + 1]) ? null : newCh[newEndIdx + 1].elm;
       addVnodes(parentElm, before, newCh, newStartIdx, newEndIdx, insertedVnodeQueue);
-    } else if (newStartIdx > newEndIdx) {
+    }
+    else if (newStartIdx > newEndIdx) {
       removeVnodes(parentElm, oldCh, oldStartIdx, oldEndIdx);
     }
   }
-
   function patchVnode(oldVnode, vnode, insertedVnodeQueue) {
     var i, hook;
     if (isDef(i = vnode.data) && isDef(hook = i.hook) && isDef(i = hook.prepatch)) {
       i(oldVnode, vnode);
     }
     var elm = vnode.elm = oldVnode.elm, oldCh = oldVnode.children, ch = vnode.children;
-    if (oldVnode === vnode) return;
+    if (oldVnode === vnode)
+      return;
     if (!sameVnode(oldVnode, vnode)) {
       var parentElm = api.parentNode(oldVnode.elm);
       elm = createElm(vnode, insertedVnodeQueue);
@@ -203,58 +209,63 @@ function init(modules, api) {
       return;
     }
     if (isDef(vnode.data)) {
-      for (i = 0; i < cbs.update.length; ++i) cbs.update[i](oldVnode, vnode);
+      for (i = 0; i < cbs.update.length; ++i)
+        cbs.update[i](oldVnode, vnode);
       i = vnode.data.hook;
-      if (isDef(i) && isDef(i = i.update)) i(oldVnode, vnode);
+      if (isDef(i) && isDef(i = i.update))
+        i(oldVnode, vnode);
     }
     if (isUndef(vnode.text)) {
       if (isDef(oldCh) && isDef(ch)) {
-        if (oldCh !== ch) updateChildren(elm, oldCh, ch, insertedVnodeQueue);
-      } else if (isDef(ch)) {
-        if (isDef(oldVnode.text)) api.setTextContent(elm, '');
+        if (oldCh !== ch)
+          updateChildren(elm, oldCh, ch, insertedVnodeQueue);
+      }
+      else if (isDef(ch)) {
+        if (isDef(oldVnode.text))
+          api.setTextContent(elm, '');
         addVnodes(elm, null, ch, 0, ch.length - 1, insertedVnodeQueue);
-      } else if (isDef(oldCh)) {
+      }
+      else if (isDef(oldCh)) {
         removeVnodes(elm, oldCh, 0, oldCh.length - 1);
-      } else if (isDef(oldVnode.text)) {
+      }
+      else if (isDef(oldVnode.text)) {
         api.setTextContent(elm, '');
       }
-    } else if (oldVnode.text !== vnode.text) {
+    }
+    else if (oldVnode.text !== vnode.text) {
       api.setTextContent(elm, vnode.text);
     }
     if (isDef(hook) && isDef(i = hook.postpatch)) {
       i(oldVnode, vnode);
     }
   }
-
-  return function(oldVnode, vnode) {
+  return function patch(oldVnode, vnode) {
     var i, elm, parent;
     var insertedVnodeQueue = [];
-    for (i = 0; i < cbs.pre.length; ++i) cbs.pre[i]();
-
+    for (i = 0; i < cbs.pre.length; ++i)
+      cbs.pre[i]();
     if (isUndef(oldVnode.sel)) {
       oldVnode = emptyNodeAt(oldVnode);
     }
-
     if (sameVnode(oldVnode, vnode)) {
       patchVnode(oldVnode, vnode, insertedVnodeQueue);
-    } else {
+    }
+    else {
       elm = oldVnode.elm;
       parent = api.parentNode(elm);
-
       createElm(vnode, insertedVnodeQueue);
-
       if (parent !== null) {
         api.insertBefore(parent, vnode.elm, api.nextSibling(elm));
         removeVnodes(parent, [oldVnode], 0, 0);
       }
     }
-
     for (i = 0; i < insertedVnodeQueue.length; ++i) {
       insertedVnodeQueue[i].data.hook.insert(insertedVnodeQueue[i]);
     }
-    for (i = 0; i < cbs.post.length; ++i) cbs.post[i]();
+    for (i = 0; i < cbs.post.length; ++i)
+      cbs.post[i]();
     return vnode;
   };
 }
-
-module.exports = {init: init};
+module.exports = { init: init };
+//# sourceMappingURL=snabbdom.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; thunk.js</summary>

```diff
diff --git a/thunk.js b/thunk.js
index 21bf7b1..9aa5cd9 100644
--- a/thunk.js
+++ b/thunk.js
@@ -1,5 +1,5 @@
+"use strict";
 var h = require('./h');
-
 function copyToThunk(vnode, thunk) {
   thunk.elm = vnode.elm;
   vnode.data.fn = thunk.data.fn;
@@ -9,15 +9,13 @@ function copyToThunk(vnode, thunk) {
   thunk.text = vnode.text;
   thunk.elm = vnode.elm;
 }
-
 function init(thunk) {
   var cur = thunk.data;
   var vnode = cur.fn.apply(undefined, cur.args);
   copyToThunk(vnode, thunk);
 }
-
 function prepatch(oldVnode, thunk) {
-  var i, old = oldVnode.data, cur = thunk.data, vnode;
+  var i, old = oldVnode.data, cur = thunk.data;
   var oldArgs = old.args, args = cur.args;
   if (old.fn !== cur.fn || oldArgs.length !== args.length) {
     copyToThunk(cur.fn.apply(undefined, args), thunk);
@@ -30,8 +28,7 @@ function prepatch(oldVnode, thunk) {
   }
   copyToThunk(oldVnode, thunk);
 }
-
-module.exports = function(sel, key, fn, args) {
+function thunk(sel, key, fn, args) {
   if (args === undefined) {
     args = fn;
     fn = key;
@@ -39,8 +36,11 @@ module.exports = function(sel, key, fn, args) {
   }
   return h(sel, {
     key: key,
-    hook: {init: init, prepatch: prepatch},
+    hook: { init: init, prepatch: prepatch },
     fn: fn,
     args: args
   });
-};
+}
+;
+module.exports = thunk;
+//# sourceMappingURL=thunk.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; vnode.js</summary>

```diff
diff --git a/vnode.js b/vnode.js
index 7057afd..16c4e67 100644
--- a/vnode.js
+++ b/vnode.js
@@ -1,5 +1,10 @@
-module.exports = function(sel, data, children, text, elm) {
+"use strict";
+function vnode(sel, data, children, text, elm) {
   var key = data === undefined ? undefined : data.key;
-  return {sel: sel, data: data, children: children,
-          text: text, elm: elm, key: key};
-};
+  return {
+    sel: sel, data: data, children: children,
+    text: text, elm: elm, key: key
+  };
+}
+module.exports = vnode;
+//# sourceMappingURL=vnode.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; helpers/attachto.js</summary>

```diff
diff --git a/helpers/attachto.js b/helpers/attachto.js
index 24c8d48..3266fed 100644
--- a/helpers/attachto.js
+++ b/helpers/attachto.js
@@ -1,3 +1,4 @@
+"use strict";
 function pre(vnode, newVnode) {
   var attachData = vnode.data.attachData;
   // Copy created placeholder and real element from old vnode
@@ -6,19 +7,16 @@ function pre(vnode, newVnode) {
   // Mount real element in vnode so the patch process operates on it
   vnode.elm = vnode.data.attachData.real;
 }
-
 function post(_, vnode) {
   // Mount dummy placeholder in vnode so potential reorders use it
   vnode.elm = vnode.data.attachData.placeholder;
 }
-
 function destroy(vnode) {
   // Remove placeholder
-  vnode.elm.parentElement.removeChild(vnode.elm);
+  vnode.elm && vnode.elm.parentElement.removeChild(vnode.elm);
   // Remove real element from where it was inserted
   vnode.elm = vnode.data.attachData.real;
 }
-
 function create(_, vnode) {
   var real = vnode.elm, attachData = vnode.data.attachData;
   var placeholder = document.createElement('span');
@@ -29,16 +27,19 @@ function create(_, vnode) {
   attachData.real = real;
   attachData.placeholder = placeholder;
 }
-
-module.exports = function(target, vnode) {
-  if (vnode.data === undefined) vnode.data = {};
-  if (vnode.data.hook === undefined) vnode.data.hook = {};
+function attachTo(target, vnode) {
+  if (vnode.data === undefined)
+    vnode.data = {};
+  if (vnode.data.hook === undefined)
+    vnode.data.hook = {};
   var data = vnode.data;
   var hook = vnode.data.hook;
-  data.attachData = {target: target, placeholder: undefined, real: undefined};
+  data.attachData = { target: target, placeholder: undefined, real: undefined };
   hook.create = create;
   hook.prepatch = pre;
   hook.postpatch = post;
   hook.destroy = destroy;
   return vnode;
-};
+}
+module.exports = attachTo;
+//# sourceMappingURL=attachto.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; modules/attributes.js</summary>

```diff
diff --git a/modules/attributes.js b/modules/attributes.js
index e486bc7..9455a8e 100644
--- a/modules/attributes.js
+++ b/modules/attributes.js
@@ -1,37 +1,37 @@
+"use strict";
 var NamespaceURIs = {
   "xlink": "http://www.w3.org/1999/xlink"
 };
-
-var booleanAttrs = ["allowfullscreen", "async", "autofocus", "autoplay", "checked", "compact", "controls", "declare",
-                "default", "defaultchecked", "defaultmuted", "defaultselected", "defer", "disabled", "draggable",
-                "enabled", "formnovalidate", "hidden", "indeterminate", "inert", "ismap", "itemscope", "loop", "multiple",
-                "muted", "nohref", "noresize", "noshade", "novalidate", "nowrap", "open", "pauseonexit", "readonly",
-                "required", "reversed", "scoped", "seamless", "selected", "sortable", "spellcheck", "translate",
-                "truespeed", "typemustmatch", "visible"];
-
+var booleanAttrs = [
+  "allowfullscreen", "async", "autofocus", "autoplay", "checked", "compact",
+  "controls", "declare", "default", "defaultchecked", "defaultmuted",
+  "defaultselected", "defer", "disabled", "draggable", "enabled",
+  "formnovalidate", "hidden", "indeterminate", "inert", "ismap", "itemscope",
+  "loop", "multiple", "muted", "nohref", "noresize", "noshade", "novalidate",
+  "nowrap", "open", "pauseonexit", "readonly", "required", "reversed", "scoped",
+  "seamless", "selected", "sortable", "spellcheck", "translate", "truespeed",
+  "typemustmatch", "visible"
+];
 var booleanAttrsDict = Object.create(null);
-for(var i=0, len = booleanAttrs.length; i < len; i++) {
+for (var i = 0, len = booleanAttrs.length; i < len; i++) {
   booleanAttrsDict[booleanAttrs[i]] = true;
 }
-
 function updateAttrs(oldVnode, vnode) {
-  var key, cur, old, elm = vnode.elm,
-      oldAttrs = oldVnode.data.attrs, attrs = vnode.data.attrs, namespaceSplit;
-
-  if (!oldAttrs && !attrs) return;
+  var key, cur, old, elm = vnode.elm, oldAttrs = oldVnode.data.attrs, attrs = vnode.data.attrs, namespaceSplit;
+  if (!oldAttrs && !attrs)
+    return;
   oldAttrs = oldAttrs || {};
   attrs = attrs || {};
-
   // update modified attributes, add new attributes
   for (key in attrs) {
     cur = attrs[key];
     old = oldAttrs[key];
     if (old !== cur) {
-      if(!cur && booleanAttrsDict[key])
+      if (!cur && booleanAttrsDict[key])
         elm.removeAttribute(key);
       else {
         namespaceSplit = key.split(":");
-        if(namespaceSplit.length > 1 && NamespaceURIs.hasOwnProperty(namespaceSplit[0]))
+        if (namespaceSplit.length > 1 && NamespaceURIs.hasOwnProperty(namespaceSplit[0]))
           elm.setAttributeNS(NamespaceURIs[namespaceSplit[0]], key, cur);
         else
           elm.setAttribute(key, cur);
@@ -47,5 +47,9 @@ function updateAttrs(oldVnode, vnode) {
     }
   }
 }
-
-module.exports = {create: updateAttrs, update: updateAttrs};
+var attributesModule = {
+  create: updateAttrs,
+  update: updateAttrs,
+};
+module.exports = attributesModule;
+//# sourceMappingURL=attributes.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; modules/class.js</summary>

```diff
diff --git a/modules/class.js b/modules/class.js
index 35e19de..8e78cff 100644
--- a/modules/class.js
+++ b/modules/class.js
@@ -1,12 +1,10 @@
+"use strict";
 function updateClass(oldVnode, vnode) {
-  var cur, name, elm = vnode.elm,
-      oldClass = oldVnode.data.class,
-      klass = vnode.data.class;
-
-  if (!oldClass && !klass) return;
+  var cur, name, elm = vnode.elm, oldClass = oldVnode.data.class, klass = vnode.data.class;
+  if (!oldClass && !klass)
+    return;
   oldClass = oldClass || {};
   klass = klass || {};
-
   for (name in oldClass) {
     if (!klass[name]) {
       elm.classList.remove(name);
@@ -19,5 +17,9 @@ function updateClass(oldVnode, vnode) {
     }
   }
 }
-
-module.exports = {create: updateClass, update: updateClass};
+var classModule = {
+  create: updateClass,
+  update: updateClass,
+};
+module.exports = classModule;
+//# sourceMappingURL=class.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; modules/dataset.js</summary>

```diff
diff --git a/modules/dataset.js b/modules/dataset.js
index 37e573b..bce15e3 100644
--- a/modules/dataset.js
+++ b/modules/dataset.js
@@ -1,13 +1,10 @@
+"use strict";
 function updateDataset(oldVnode, vnode) {
-  var elm = vnode.elm,
-    oldDataset = oldVnode.data.dataset,
-    dataset = vnode.data.dataset,
-    key
-
-  if (!oldDataset && !dataset) return;
+  var elm = vnode.elm, oldDataset = oldVnode.data.dataset, dataset = vnode.data.dataset, key;
+  if (!oldDataset && !dataset)
+    return;
   oldDataset = oldDataset || {};
   dataset = dataset || {};
-
   for (key in oldDataset) {
     if (!dataset[key]) {
       delete elm.dataset[key];
@@ -19,5 +16,9 @@ function updateDataset(oldVnode, vnode) {
     }
   }
 }
-
-module.exports = {create: updateDataset, update: updateDataset}
+var datasetModule = {
+  create: updateDataset,
+  update: updateDataset,
+};
+module.exports = datasetModule;
+//# sourceMappingURL=dataset.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; modules/eventlisteners.js</summary>

```diff
diff --git a/modules/eventlisteners.js b/modules/eventlisteners.js
index 34ad98d..47d1e53 100644
--- a/modules/eventlisteners.js
+++ b/modules/eventlisteners.js
@@ -1,20 +1,24 @@
+"use strict";
 function invokeHandler(handler, vnode, event) {
   if (typeof handler === "function") {
     // call function handler
     handler.call(vnode, event, vnode);
-  } else if (typeof handler === "object") {
+  }
+  else if (typeof handler === "object") {
     // call handler with arguments
     if (typeof handler[0] === "function") {
       // special case for single argument for performance
       if (handler.length === 2) {
         handler[0].call(vnode, handler[1], event, vnode);
-      } else {
+      }
+      else {
         var args = handler.slice(1);
         args.push(event);
         args.push(vnode);
         handler[0].apply(vnode, args);
       }
-    } else {
+    }
+    else {
       // call multiple handlers
       for (var i = 0; i < handler.length; i++) {
         invokeHandler(handler[i]);
@@ -22,36 +26,24 @@ function invokeHandler(handler, vnode, event) {
     }
   }
 }
-
 function handleEvent(event, vnode) {
-  var name = event.type,
-      on = vnode.data.on;
-
+  var name = event.type, on = vnode.data.on;
   // call event handler(s) if exists
   if (on && on[name]) {
     invokeHandler(on[name], vnode, event);
   }
 }
-
 function createListener() {
   return function handler(event) {
     handleEvent(event, handler.vnode);
-  }
+  };
 }
-
 function updateEventListeners(oldVnode, vnode) {
-  var oldOn = oldVnode.data.on,
-      oldListener = oldVnode.listener,
-      oldElm = oldVnode.elm,
-      on = vnode && vnode.data.on,
-      elm = vnode && vnode.elm,
-      name;
-
+  var oldOn = oldVnode.data.on, oldListener = oldVnode.listener, oldElm = oldVnode.elm, on = vnode && vnode.data.on, elm = vnode && vnode.elm, name;
   // optimization for reused immutable handlers
   if (oldOn === on) {
     return;
   }
-
   // remove existing listeners which no longer used
   if (oldOn && oldListener) {
     // if element changed or deleted we remove all existing listeners unconditionally
@@ -60,7 +52,8 @@ function updateEventListeners(oldVnode, vnode) {
         // remove listener if element was changed or existing listeners removed
         oldElm.removeEventListener(name, oldListener, false);
       }
-    } else {
+    }
+    else {
       for (name in oldOn) {
         // remove listener if existing listener removed
         if (!on[name]) {
@@ -69,21 +62,20 @@ function updateEventListeners(oldVnode, vnode) {
       }
     }
   }
-
   // add new listeners which has not already attached
   if (on) {
     // reuse existing listener or create new
     var listener = vnode.listener = oldVnode.listener || createListener();
     // update vnode for listener
     listener.vnode = vnode;
-
     // if element changed or added we add all needed listeners unconditionally
     if (!oldOn) {
       for (name in on) {
         // add listener if element was changed or new listeners added
         elm.addEventListener(name, listener, false);
       }
-    } else {
+    }
+    else {
       for (name in on) {
         // add listener if new listener added
         if (!oldOn[name]) {
@@ -93,9 +85,10 @@ function updateEventListeners(oldVnode, vnode) {
     }
   }
 }
-
-module.exports = {
+var eventListenersModule = {
   create: updateEventListeners,
   update: updateEventListeners,
-  destroy: updateEventListeners
+  destroy: updateEventListeners,
 };
+module.exports = eventListenersModule;
+//# sourceMappingURL=eventlisteners.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; modules/hero.js</summary>

```diff
diff --git a/modules/hero.js b/modules/hero.js
index b8b3159..e4cca55 100644
--- a/modules/hero.js
+++ b/modules/hero.js
@@ -1,58 +1,51 @@
+"use strict";
 var raf = (typeof window !== 'undefined' && window.requestAnimationFrame) || setTimeout;
-var nextFrame = function(fn) { raf(function() { raf(fn); }); };
-
+var nextFrame = function (fn) { raf(function () { raf(fn); }); };
 function setNextFrame(obj, prop, val) {
-  nextFrame(function() { obj[prop] = val; });
+  nextFrame(function () { obj[prop] = val; });
 }
-
 function getTextNodeRect(textNode) {
   var rect;
   if (document.createRange) {
     var range = document.createRange();
     range.selectNodeContents(textNode);
     if (range.getBoundingClientRect) {
-        rect = range.getBoundingClientRect();
+      rect = range.getBoundingClientRect();
     }
   }
   return rect;
 }
-
 function calcTransformOrigin(isTextNode, textRect, boundingRect) {
   if (isTextNode) {
     if (textRect) {
       //calculate pixels to center of text from left edge of bounding box
-      var relativeCenterX = textRect.left + textRect.width/2 - boundingRect.left;
-      var relativeCenterY = textRect.top + textRect.height/2 - boundingRect.top;
+      var relativeCenterX = textRect.left + textRect.width / 2 - boundingRect.left;
+      var relativeCenterY = textRect.top + textRect.height / 2 - boundingRect.top;
       return relativeCenterX + 'px ' + relativeCenterY + 'px';
     }
   }
   return '0 0'; //top left
 }
-
 function getTextDx(oldTextRect, newTextRect) {
   if (oldTextRect && newTextRect) {
-    return ((oldTextRect.left + oldTextRect.width/2) - (newTextRect.left + newTextRect.width/2));
+    return ((oldTextRect.left + oldTextRect.width / 2) - (newTextRect.left + newTextRect.width / 2));
   }
   return 0;
 }
 function getTextDy(oldTextRect, newTextRect) {
   if (oldTextRect && newTextRect) {
-    return ((oldTextRect.top + oldTextRect.height/2) - (newTextRect.top + newTextRect.height/2));
+    return ((oldTextRect.top + oldTextRect.height / 2) - (newTextRect.top + newTextRect.height / 2));
   }
   return 0;
 }
-
 function isTextElement(elm) {
   return elm.childNodes.length === 1 && elm.childNodes[0].nodeType === 3;
 }
-
 var removed, created;
-
-function pre(oldVnode, vnode) {
+function pre() {
   removed = {};
   created = [];
 }
-
 function create(oldVnode, vnode) {
   var hero = vnode.data.hero;
   if (hero && hero.id) {
@@ -60,7 +53,6 @@ function create(oldVnode, vnode) {
     created.push(vnode);
   }
 }
-
 function destroy(vnode) {
   var hero = vnode.data.hero;
   if (hero && hero.id) {
@@ -68,25 +60,21 @@ function destroy(vnode) {
     vnode.isTextNode = isTextElement(elm); //is this a text node?
     vnode.boundingRect = elm.getBoundingClientRect(); //save the bounding rectangle to a new property on the vnode
     vnode.textRect = vnode.isTextNode ? getTextNodeRect(elm.childNodes[0]) : null; //save bounding rect of inner text node
-    var computedStyle = window.getComputedStyle(elm, null); //get current styles (includes inherited properties)
+    var computedStyle = window.getComputedStyle(elm, void 0); //get current styles (includes inherited properties)
     vnode.savedStyle = JSON.parse(JSON.stringify(computedStyle)); //save a copy of computed style values
     removed[hero.id] = vnode;
   }
 }
-
 function post() {
-  var i, id, newElm, oldVnode, oldElm, hRatio, wRatio,
-      oldRect, newRect, dx, dy, origTransform, origTransition,
-      newStyle, oldStyle, newComputedStyle, isTextNode,
-      newTextRect, oldTextRect;
+  var i, id, newElm, oldVnode, oldElm, hRatio, wRatio, oldRect, newRect, dx, dy, origTransform, origTransition, newStyle, oldStyle, newComputedStyle, isTextNode, newTextRect, oldTextRect;
   for (i = 0; i < created.length; i += 2) {
     id = created[i];
-    newElm = created[i+1].elm;
+    newElm = created[i + 1].elm;
     oldVnode = removed[id];
     if (oldVnode) {
       isTextNode = oldVnode.isTextNode && isTextElement(newElm); //Are old & new both text?
       newStyle = newElm.style;
-      newComputedStyle = window.getComputedStyle(newElm, null); //get full computed style for new element
+      newComputedStyle = window.getComputedStyle(newElm, void 0); //get full computed style for new element
       oldElm = oldVnode.elm;
       oldStyle = oldElm.style;
       //Overall element bounding boxes
@@ -98,7 +86,8 @@ function post() {
         oldTextRect = oldVnode.textRect;
         dx = getTextDx(oldTextRect, newTextRect);
         dy = getTextDy(oldTextRect, newTextRect);
-      } else {
+      }
+      else {
         //Calculate distances between old & new positions
         dx = oldRect.left - newRect.left;
         dy = oldRect.top - newRect.top;
@@ -108,24 +97,24 @@ function post() {
       // Animate new element
       origTransform = newStyle.transform;
       origTransition = newStyle.transition;
-      if (newComputedStyle.display === 'inline') //inline elements cannot be transformed
-        newStyle.display = 'inline-block';        //this does not appear to have any negative side effects
+      if (newComputedStyle.display === 'inline')
+        newStyle.display = 'inline-block'; //this does not appear to have any negative side effects
       newStyle.transition = origTransition + 'transform 0s';
       newStyle.transformOrigin = calcTransformOrigin(isTextNode, newTextRect, newRect);
       newStyle.opacity = '0';
-      newStyle.transform = origTransform + 'translate('+dx+'px, '+dy+'px) ' +
-                               'scale('+1/wRatio+', '+1/hRatio+')';
+      newStyle.transform = origTransform + 'translate(' + dx + 'px, ' + dy + 'px) ' +
+        'scale(' + 1 / wRatio + ', ' + 1 / hRatio + ')';
       setNextFrame(newStyle, 'transition', origTransition);
       setNextFrame(newStyle, 'transform', origTransform);
       setNextFrame(newStyle, 'opacity', '1');
       // Animate old element
-      for (var key in oldVnode.savedStyle) { //re-apply saved inherited properties
+      for (var key in oldVnode.savedStyle) {
         if (parseInt(key) != key) {
-          var ms = key.substring(0,2) === 'ms';
-          var moz = key.substring(0,3) === 'moz';
-          var webkit = key.substring(0,6) === 'webkit';
-      	  if (!ms && !moz && !webkit) //ignore prefixed style properties
-        	  oldStyle[key] = oldVnode.savedStyle[key];
+          var ms = key.substring(0, 2) === 'ms';
+          var moz = key.substring(0, 3) === 'moz';
+          var webkit = key.substring(0, 6) === 'webkit';
+          if (!ms && !moz && !webkit)
+            oldStyle[key] = oldVnode.savedStyle[key];
         }
       }
       oldStyle.position = 'absolute';
@@ -133,14 +122,14 @@ function post() {
       oldStyle.left = oldRect.left + 'px';
       oldStyle.width = oldRect.width + 'px'; //Needed for elements who were sized relative to their parents
       oldStyle.height = oldRect.height + 'px'; //Needed for elements who were sized relative to their parents
-      oldStyle.margin = 0; //Margin on hero element leads to incorrect positioning
+      oldStyle.margin = '0'; //Margin on hero element leads to incorrect positioning
       oldStyle.transformOrigin = calcTransformOrigin(isTextNode, oldTextRect, oldRect);
       oldStyle.transform = '';
       oldStyle.opacity = '1';
       document.body.appendChild(oldElm);
-      setNextFrame(oldStyle, 'transform', 'translate('+ -dx +'px, '+ -dy +'px) scale('+wRatio+', '+hRatio+')'); //scale must be on far right for translate to be correct
+      setNextFrame(oldStyle, 'transform', 'translate(' + -dx + 'px, ' + -dy + 'px) scale(' + wRatio + ', ' + hRatio + ')'); //scale must be on far right for translate to be correct
       setNextFrame(oldStyle, 'opacity', '0');
-      oldElm.addEventListener('transitionend', function(ev) {
+      oldElm.addEventListener('transitionend', function (ev) {
         if (ev.propertyName === 'transform')
           document.body.removeChild(ev.target);
       });
@@ -148,5 +137,11 @@ function post() {
   }
   removed = created = undefined;
 }
-
-module.exports = {pre: pre, create: create, destroy: destroy, post: post};
+var heroModule = {
+  pre: pre,
+  create: create,
+  destroy: destroy,
+  post: post,
+};
+module.exports = heroModule;
+//# sourceMappingURL=hero.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; modules/props.js</summary>

```diff
diff --git a/modules/props.js b/modules/props.js
index 843c8a8..c5575ce 100644
--- a/modules/props.js
+++ b/modules/props.js
@@ -1,11 +1,10 @@
+"use strict";
 function updateProps(oldVnode, vnode) {
-  var key, cur, old, elm = vnode.elm,
-      oldProps = oldVnode.data.props, props = vnode.data.props;
-
-  if (!oldProps && !props) return;
+  var key, cur, old, elm = vnode.elm, oldProps = oldVnode.data.props, props = vnode.data.props;
+  if (!oldProps && !props)
+    return;
   oldProps = oldProps || {};
   props = props || {};
-
   for (key in oldProps) {
     if (!props[key]) {
       delete elm[key];
@@ -19,5 +18,9 @@ function updateProps(oldVnode, vnode) {
     }
   }
 }
-
-module.exports = {create: updateProps, update: updateProps};
+var propsModule = {
+  create: updateProps,
+  update: updateProps,
+};
+module.exports = propsModule;
+//# sourceMappingURL=props.js.map
\ No newline at end of file
```

</details>

<details>
  <summary>&#9654; modules/style.js</summary>

```diff
diff --git a/modules/style.js b/modules/style.js
index e03b633..48cb700 100644
--- a/modules/style.js
+++ b/modules/style.js
@@ -1,20 +1,16 @@
+"use strict";
 var raf = (typeof window !== 'undefined' && window.requestAnimationFrame) || setTimeout;
-var nextFrame = function(fn) { raf(function() { raf(fn); }); };
-
+var nextFrame = function (fn) { raf(function () { raf(fn); }); };
 function setNextFrame(obj, prop, val) {
-  nextFrame(function() { obj[prop] = val; });
+  nextFrame(function () { obj[prop] = val; });
 }
-
 function updateStyle(oldVnode, vnode) {
-  var cur, name, elm = vnode.elm,
-      oldStyle = oldVnode.data.style,
-      style = vnode.data.style;
-
-  if (!oldStyle && !style) return;
+  var cur, name, elm = vnode.elm, oldStyle = oldVnode.data.style, style = vnode.data.style;
+  if (!oldStyle && !style)
+    return;
   oldStyle = oldStyle || {};
   style = style || {};
   var oldHasDel = 'delayed' in oldStyle;
-
   for (name in oldStyle) {
     if (!style[name]) {
       elm.style[name] = '';
@@ -29,28 +25,27 @@ function updateStyle(oldVnode, vnode) {
           setNextFrame(elm.style, name, cur);
         }
       }
-    } else if (name !== 'remove' && cur !== oldStyle[name]) {
+    }
+    else if (name !== 'remove' && cur !== oldStyle[name]) {
       elm.style[name] = cur;
     }
   }
 }
-
 function applyDestroyStyle(vnode) {
   var style, name, elm = vnode.elm, s = vnode.data.style;
-  if (!s || !(style = s.destroy)) return;
+  if (!s || !(style = s.destroy))
+    return;
   for (name in style) {
     elm.style[name] = style[name];
   }
 }
-
 function applyRemoveStyle(vnode, rm) {
   var s = vnode.data.style;
   if (!s || !s.remove) {
     rm();
     return;
   }
-  var name, elm = vnode.elm, idx, i = 0, maxDur = 0,
-      compStyle, style = s.remove, amount = 0, applied = [];
+  var name, elm = vnode.elm, i = 0, compStyle, style = s.remove, amount = 0, applied = [];
   for (name in style) {
     applied.push(name);
     elm.style[name] = style[name];
@@ -58,12 +53,21 @@ function applyRemoveStyle(vnode, rm) {
   compStyle = getComputedStyle(elm);
   var props = compStyle['transition-property'].split(', ');
   for (; i < props.length; ++i) {
-    if(applied.indexOf(props[i]) !== -1) amount++;
+    if (applied.indexOf(props[i]) !== -1)
+      amount++;
   }
-  elm.addEventListener('transitionend', function(ev) {
-    if (ev.target === elm) --amount;
-    if (amount === 0) rm();
+  elm.addEventListener('transitionend', function (ev) {
+    if (ev.target === elm)
+      --amount;
+    if (amount === 0)
+      rm();
   });
 }
-
-module.exports = {create: updateStyle, update: updateStyle, destroy: applyDestroyStyle, remove: applyRemoveStyle};
+var styleModule = {
+  create: updateStyle,
+  update: updateStyle,
+  destroy: applyDestroyStyle,
+  remove: applyRemoveStyle,
+};
+module.exports = styleModule;
+//# sourceMappingURL=style.js.map
\ No newline at end of file
```

</details>


staltz:
Oh crap. I forgot to commit `src/modules/` and `src/helpers/` so those are missing entirely, and I also lost the source files. I'll have to do those all over again. 🤕 

This PR is marked work-in-progress until then. I'll be back

src/vnode.ts
```diff
@@ -0,0 +1,19 @@
...
+function vnode(sel: string, data: any): VNode;
+function vnode(sel: string, text: string): VNode;
+function vnode(sel: string, children: Array<VNode>): VNode;
+function vnode(sel: string, data: any, text: string): VNode;
+function vnode(sel: string, data: any, children: Array<VNode>): VNode;
+function vnode(sel?: any, data?: any, children?: any, text?: any): VNode;
+function vnode(sel?: any, data?: any, children?: any, text?: undefined, elm?: undefined): VNode;
+function vnode(sel?: any, data?: any, children?: any, text?: any, elm?: undefined): VNode;
+function vnode(sel?: any, data?: any, children?: any, text?: undefined, elm?: any): VNode;
+function vnode(sel?: any, data?: any, children?: any, text?: any, elm?: any): VNode {
```
TylorS: I don't think this function is actually overloaded. I think it's really just 
````typescript
function vnode (sel: string | undefined, 
                         data: any | undefined, 
                         children: Array<VNode | string> | undefined, 
                         text: string | undefined, 
                         elm: Element | HTMLElement | undefined): VNode;`

Comment (staltz) [inline src/vnode.ts]: 👍

**Comment by staltz** [comment]
Ok, I fixed the commit. This is now ready for review.

File: src/interfaces.d.ts [hunk 2]
```diff
@@ -0,0 +1,86 @@
+export interface VNodeData {
```
TylorS: @staltz Do you know a way to also allow arbitrary keys/values here? When defining your own snabbdom modules you likely need a property not defined here.

```TypeScript
interface VNodeData {
  ... // all properties here
  [ key: string ]: any;
}
```

Is there something like this that we can do?

src/modules/attributes.ts
```diff
@@ -48,4 +51,4 @@ function updateAttrs(oldVnode, vnode) {
   }
 }
 
-module.exports = {create: updateAttrs, update: updateAttrs};
+export = {create: updateAttrs, update: updateAttrs} as Hooks;
```
TylorS: This, although it overlaps, should be exported `as Module` instead of `as Hooks`.

src/modules/class.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+  for (name in klass) {
+    cur = klass[name];
+    if (cur !== oldClass[name]) {
+      (elm.classList as any)[cur ? 'add' : 'remove'](name);
+    }
+  }
+}
+
+export = {create: updateClass, update: updateClass} as Hooks;
```
TylorS: `as Module` again

src/modules/dataset.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+  }
+  for (key in dataset) {
+    if (oldDataset[key] !== dataset[key]) {
+      elm.dataset[key] = dataset[key];
+    }
+  }
+}
+
+export = {create: updateDataset, update: updateDataset} as Hooks;
```
TylorS: `as Module`

src/modules/eventlisteners.ts
```diff
@@ -94,8 +96,8 @@ function updateEventListeners(oldVnode, vnode) {
   }
 }
 
-module.exports = {
+export = {
   create: updateEventListeners,
   update: updateEventListeners,
   destroy: updateEventListeners
-};
+} as Hooks;
```
TylorS: `as Module`

src/modules/hero.ts
```diff
@@ -119,34 +128,34 @@ function post() {
...
-          document.body.removeChild(ev.target);
+          document.body.removeChild(ev.target as Node);
       });
     }
   }
   removed = created = undefined;
 }
 
-module.exports = {pre: pre, create: create, destroy: destroy, post: post};
+export = {pre: pre, create: create, destroy: destroy, post: post} as Hooks;
```
TylorS: `as Module`

src/modules/props.ts
```diff
@@ -0,0 +1,26 @@
...
+  for (key in props) {
+    cur = props[key];
+    old = oldProps[key];
+    if (old !== cur && (key !== 'value' || (elm as any)[key] !== cur)) {
+      (elm as any)[key] = cur;
+    }
+  }
+}
+
+export = {create: updateProps, update: updateProps} as Hooks;
```
TylorS: `as Module`

src/modules/style.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (amount === 0) rm();
+  });
+}
+
+export = {
+  create: updateStyle,
+  update: updateStyle,
+  destroy: applyDestroyStyle,
+  remove: applyRemoveStyle
+} as Hooks;
```
TylorS: `as Module`

paldepind:
Super awesome :+1: 

> To keep this PR as small as possible (since it's anyway a huge thing to migrate to TS) and to make sure all tests pass and performance is the same, I basically just converted JS to TS in the least intrusive way as possible.

I think that is a very good approach.

I'll probably find the time to look through it in the weekend.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.5.4",
   "description": "A virtual DOM library with focus on simplicity, modularity, powerful features and performance.",
   "main": "snabbdom.js",
-  "typings": "type-definitions/snabbdom.d.ts",
+  "typings": "snabbdom.d.ts",
```
paldepind: Is this needed? Since `declaration` is `true` in `tsconfig.json` matching `*.d.ts` should be created for every compiled file and TypeScript should pick those up automatically?

TylorS: Yeah, this allows the TypeScript compiler to find the associated typings automatically via NPM. So the user doesn't have to do anything else other than `npm install --save snabbdom`.

staltz:
Ok, I updated with Tylor's suggestions. I also added an npm script `pretest` to compile the typescript before running the tests.

paldepind:
Looks very good to me! :+1: :clap: 

Two ideas for further improvement:

* Instead of one interface file, we could define the interfaces in relevant files. For instance, define the thunk related interfaces in `thunk.ts` and the `VNode` interface in `vnode.ts`?

* We could use ES6 export syntax instead of the TS specific `export =` where applicable? For instance, in `class.ts` this
```ts
export {updateClass as create, updateClass as update};
```
gives the same result as the current
```ts
export = {create: updateAttrs, update: updateAttrs} as Module;
```
But uses standard JavaScript syntax.

Anyway, both are minor things. I'd have no problem with just merging and releasing this as is.

staltz:
> We could use ES6 export syntax instead of the TS specific export = where applicable?

I started like that, but for 100% compatibility with the current snabbdom version, I wanted to support CommonJS, that's why I used `export =`. If we use ES6 modules, we might change what happens for existing users that use `require()`, they may need to pick also the default. This is a good opportunity to ask, though, should we move entirely to ES6 modules with defaults, or remain CommonJS?

> Instead of one interface file, we could define the interfaces in relevant files. For instance, define the thunk related interfaces in thunk.ts and the VNode interface in vnode.ts?

That was my initial attempt. However, in modules where we do `export =` for compatibility with CommonJS, TypeScript doesn't allow us to export an interface. When there is `export =`, we can't have other exports.

So, I'd do these modifications you suggested, but they may end up being a breaking change to the API, so that's something we need to consider carefully.

TylorS:
I'm definitely in favor of ES2015 imports/exports as then we could also provide ES2015 builds for tree-shaking compatibility.

paldepind:
As far as I can tell most of the exports would be backwards compatible. A few of them wouldn't but maybe that is acceptable?

For the problematic ones like `h.ts` and `thunk.ts` we could either do them as default exports or export a single function. Then for `h.ts` the compiled code would be

```ts
exports.default = h;
```

or 

```ts
exports.h = h;
```

I'm not so happy about how default exports compile so I'd prefer the later.

> That was my initial attempt. However, in modules where we do `export =` for compatibility with CommonJS, TypeScript doesn't allow us to export an interface. When there is `export =`, we can't have other exports.

That makes sense. I think it is another thing in favor of moving to JavaScript exports.

acstll:
I'm also in favor of using ES2015 imports/exports.

(I was just thinking about opening an issue with some ideas and breaking changes for a 1.0 release …taking advantage of this TS-rewrite-staltz-and-TylorS-joining oportunity. Will try and do it this morning.)

staltz:
> we could either do them as default exports or export a single function

Ok, so if we export a single function, it would be e.g. in `h.ts`:

```typescript
export {h};
```

And then usage would be:

In TypeScript: `import {h} from 'snabbdom/h';`
In ES6: same as above
In CommonJS: `const {h} = require('snabbdom/h');`

Notice that this would be a breaking change from the current usage style with CommonJS:
`const h = require('snabbdom/h');`

Do we all agree on this?

staltz:
And another question: how should snabbdom top-level file be imported?
```typescript
import {init} from 'snabbdom';
```
perhaps?

acstll:
> And another question: how should snabbdom top-level file be imported?

I think you should be able to do:

```js
import { patch, h, thunk } from 'snabbdom'
```

and be done. Where `patch` has the basic built-in modules (attrs, props, class and events) (that's the current `snabbdom.bundle` file I think).

To keep the modularity and flexibility you could do:

```js
import init from 'snabbdom/init'
import h from 'snabbdom/h'
import { foo, bar } from 'snabbdom/modules'
// or import foo from 'snabbdom/modules/foo'

// etc.
```

I think this is a small changes that would make a huge difference in people adopting the library.

This would also make it easy for new-comers to try it out without having to copy/paste that example in the README (which I sometimes still do 😁).

paldepind:
> I think you should be able to do:
```js
import { patch, h, thunk } from 'snabbdom'
```
and be done.

That is a nice idea. In fact I think we'd have to do something like that if we want to meaningfully take advantage of tree shaking. Currently every separate thing is in a separate file. That is nice when one does not have three shaking—but leaves nothing for the tree shaker to do.

staltz:
Check the latest commit with the migration to ES2015 modules. Highlights are

- It's a breaking change
- You can import in two ways: `import h from 'snabbdom/h'` or `import {h} from 'snabbdom/h'`. The CommonJS equivalent is `const h = require('snabbdom/h').h` or `const h = require('snabbdom/h').default`
- No more interfaces.d.ts file. The VNode interface is in the vnode.ts file, and similarly for other interfaces

TylorS:
+ :100: for `import { h, thunk, init } from 'snabbdom'` from me :)

This would allow for significantly more simple tooling, and bundling would be extremely straightforward.

staltz:
So, what's up with this PR now? Good for merge?

acstll:
I just opened an issue https://github.com/snabbdom/snabbdom/issues/194 to continue discussing the main exports idea (and others). So this is just about the TS migration.

paldepind:
Really great improvements IMO!

> You can import in two ways: `import h from 'snabbdom/h'` or `import {h} from 'snabbdom/h'`. The CommonJS equivalent is `const h = require('snabbdom/h').default` or `const h = require('snabbdom/h').h`

Good idea! Then people can use what suits them best.

Should we merge this and continue with the rest?

staltz:
Yay! Onto the next parts now :)

PR #201, opened 2016-12-11
Tweak types, reduce casting

paldepind:
I took a stab at removing some of the casting in the TypeScript port.

Mostly by replacing `isDef(foo)` with `foo !== undefined` because the TS compiler don't understand that `isDef` does `undefined`-checking.

I've removed some other ones by tweaking the types slightly.

I managed to get rid of 19 casts and a few `any`s.

### DIFF
--- a/src/is.ts
+++ b/src/is.ts
@@ -1,4 +1,4 @@
 export const array = Array.isArray;
-export function primitive(s: any): boolean {
+export function primitive(s: any): s is (string | number) {
   return typeof s === 'string' || typeof s === 'number';
-};
+}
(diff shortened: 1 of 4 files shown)


src/is.ts
```diff
@@ -1,4 +1,4 @@
 export const array = Array.isArray;
-export function primitive(s: any): boolean {
+export function primitive(s: any): s is (string | number) {
```
TylorS: Wasn't even aware of this TS syntax! Really nice :+1:

paldepind: It's not super common. But can be very handy 😄  It's called a type predicate.

staltz:
Yep, this is what I would do too.

paldepind:
@mightyiam Yep 😄

PR #205, opened 2016-12-17
Use `typeof`

paldepind:
This PR uses two nifty new features from TypeScript 2.1. `keyof` and mapped types to typecheck the stored module hooks in `cbs`.

I also changed the type of the `data` argument to `h` from `any` to `VNodeData`. I think that is better as it assures that "known" properties on the data object are type checked.

A few other things are tweaked as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "gulp-uglify": "^1.5.3",
     "knuth-shuffle": "^1.0.1",
     "testem": "^1.0.2",
-    "typescript": "^2.0.6",
+    "typescript": "^2.1.0",
     "xyz": "0.5.x"
   },
   "scripts": {
--- a/src/h.ts
+++ b/src/h.ts
@@ -1,4 +1,4 @@
-import {vnode, VNode} from './vnode';
+import {vnode, VNode, VNodeData} from './vnode';
 import * as is from './is';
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "gulp-uglify": "^1.5.3",
     "knuth-shuffle": "^1.0.1",
     "testem": "^1.0.2",
-    "typescript": "^2.0.6",
+    "typescript": "^2.1.0",
```
staltz: Because TypeScript 2.1 is a breaking change from 2.0, I recommend we first publish the existing TypeScript-based snabbdom version, then later merge this PR and publish it as a breaking change (yes, bumping `major`).

staltz: Clarifying, because of using `keyof`, snabbdom users and frameworks still based on TypeScript v2.0 *won't* be able to compile their projects if this PR is in snabbdom.

staltz: Other than that, I'm 👍 for this PR.

paldepind: To make sure I understand correctly. You think we should first publish a version using TypeScript 2.0 and then afterwards one using 2.1? I don't understand the issue? Upgrading from 2.0 to 2.1 should be easy for people?

PR #210, opened 2016-12-23
use parentNode instead of parentElement

aronallen:
In IE11 children and parentElement are undefined for SVG elements. 
Instead childNodes and parentNode should be used.

### DIFF
--- a/src/helpers/attachto.ts
+++ b/src/helpers/attachto.ts
@@ -16,7 +16,7 @@ function post(_: any, vnode: VNode): void {
 
 function destroy(vnode: VNode): void {
   // Remove placeholder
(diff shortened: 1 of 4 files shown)


paldepind:
Does this happen only in IE11? Would you maybe add a test as well for the problem?

aronallen:
I believe it is IE11 only. I don't have access to my IE11 machine right now.
It should be quite simple to write a test for the HTMLDOM API.

aronallen:
@paldepind I wrote a test now. This is quite critical for us, as our application breaks in SVG when drawing charts.

paldepind:
Great! Thanks for the PR.

I plan on doing the next release in ~ 1 week.

aronallen:
@paldepind det var så lidt.

PR #212, opened 2016-12-26
Fix compile issues

### LINKED ISSUE #211: `master` fails to compile
I get this error when running `npm install` or `npm run compile` on a freshly cloned copy of the repo:
```
src/helpers/attachto.ts(19,16): error TS2531: Object is possibly 'null'.
src/htmldomapi.ts(39,10): error TS2322: Type 'HTMLElement | null' is not assignable to type 'HTMLElement'.
  Type 'null' is not assignable to type 'HTMLElement'.
src/htmldomapi.ts(43,10): error TS2322: Type 'Node | null' is not assignable to type 'Node'.
  Type 'null' is not assignable to type 'Node'.
```

whmountains:
Fixes #211


paldepind:
Thank you @whmountains. It looks good! However, you've accidentally committed a `yarn.lock` file as well and made changes to `tsconfig.json`.

#205 fixes this as well. But if we want to do a release using TypeScript 2.0 then I think we should merge this.

paldepind:
Thank you for the PR. But the compiler issues has now been fixed by #205.

Issue #192, opened 2016-11-08
[Question] selector=svg check is weird


bogas04:
https://github.com/snabbdom/snabbdom/blob/master/h.js#L30

```javascript
  if (sel[0] === 's' && sel[1] === 'v' && sel[2] === 'g') {
    addNS(data, children, sel);
}
```

Why would you do that ? Perf benefits?

paldepind:
How would you do it? :smile: 

I wrote this reply in [another thread](https://github.com/snabbdom/snabbdom/pull/213):

> I don't think it matters. But I like the current code. It's simple, takes constant time and bails out early. `indexOf` would always scan the entire selector looking for `"svg"` and therefore take time proportional to the length of the selector string.

mightyiam:
```js
if (sel.slice(0, 3) === 'svg') {
}
```

Hypnosphi:
if (/^svg/.match(sel))

Hypnosphi:
direct char comparison is faster anyway (https://jsperf.com/string-startswith), but it would be nice to abstract it into a helper function to make this part of code cleaner

bogas04:
How about using [startsWith](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) ?

paldepind:
Using `String#slice` has the disadvantage that it creates a new string. A regex is slightly more complex IMO. `startsWith` is nice but it is not supported in all the browsers that Snabbdom otherwise work in.

staltz:
In general, it's ok for a library like snabbdom which focuses on size and speed to hurt code readability here and there. I'm 👍  for the char comparisons.

nunocastromartins:
The only issue I see with the current solution is that there can't exist another tag that starts with `svg` without the element getting the SVG namespace. This means that I cannot create a custom element named `svgx`; and that, in the unlikely future where a new tag with such a name is introduced in HTML, this will be broken.

I'd maybe go with something like:
```js
if (sel[0] === 's' && sel[1] === 'v' && sel[2] === 'g' &&
    (sel.length === 3 || sel[3] === '#' || sel[3] === '.')) {
    addNS(data, children, sel);
}
```

paldepind:
@nunocastromartins That has been fixed by #213 and is included in the latest release.

nunocastromartins:
You're right, I didn't see it, sorry! :-)

865881900:
[...]

PR #213, opened 2016-12-28
Fix Custom Element creation when tag name begins with 'svg'

tdumitrescu:
The check at https://github.com/snabbdom/snabbdom/blob/f79af828/src/h.ts#L37
is too loose, and ends up adding the svg namespace to elements with
tagNames such as `svg-icon`, which prevents web components lifecycle
callbacks from firing.

Added test cases for the custom element case as well as longer svg selectors
(e.g. `svg#my-id`) which are currently untested.

### DIFF
--- a/src/h.ts
+++ b/src/h.ts
@@ -34,7 +34,10 @@ export function h(sel: any, b?: any, c?: any): VNode {
       if (is.primitive(children[i])) children[i] = (vnode as any)(undefined, undefined, undefined, children[i]);
     }
   }
-  if (sel[0] === 's' && sel[1] === 'v' && sel[2] === 'g') {
+  if (
+    sel[0] === 's' && sel[1] === 'v' && sel[2] === 'g' &&
+    (sel.length === 3 || sel[3] === '.' || sel[3] === '#')
+  ) {
     addNS(data, children, sel);
   }
   return vnode(sel, data, children, text, undefined);
--- a/test/core.js
+++ b/test/core.js
@@ -89,15 +89,23 @@ describe('snabbdom', function() {
       elm = patch(vnode0, h('div', [h('div', {ns: SVGNamespace})])).elm;
       assert.equal(elm.firstChild.namespaceURI, SVGNamespace);
 
+      // verify that svg tag automatically gets svg namespace
       elm = patch(vnode0, h('svg', [
         h('foreignObject', [
           h('div', ['I am HTML embedded in SVG'])
         ])
       ])).elm;
-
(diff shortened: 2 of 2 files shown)


tdumitrescu:
Also, further to https://github.com/snabbdom/snabbdom/issues/192, I'd be happy to refactor the first line of that check to `sel.indexOf('svg') === 0` unless there's a reason not to.

paldepind:
Yes. You're right. That is certainly a bug. The fix and test cases looks good to me :+1:

What do you think about either creating a new test for this bug? Or alternatively adding comments to the added test lines to describe the bug being tested for?

> I'd be happy to refactor the first line of that check to sel.indexOf('svg') === 0 unless there's a reason not to.

I don't think it matters. But I like the current code. It's simple, takes constant time and bails out early. `indexOf` would always scan the entire selector looking for `"svg"` and therefore take time proportional to the length of the selector string.

mightyiam:
Why would indexOf scan the entire string? You have references for that?

tdumitrescu:
For cases where the `searchValue` argument to `str.indexOf(searchValue)` is not present in `str`, the method will have to scan `str.length - searchValue.length` chars to determine that there is no match (assuming the string object hasn't been pre-indexed by the runtime, which AFAIK doesn't happen in any current JS implementations). For instance, `'blockquote#my-quote.class1.class2.extra-super-long-class'.indexOf('svg')` will have to scan 53 chars before returning -1, whereas the existing code will scan at most 3 chars. `str.startsWith` would take care of that, but it's still got [pretty limited browser support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith#Browser_compatibility) (apparently including _no_ IE implementation). So from a performance perspective, it makes sense to keep the existing implementation, at least inasmuch as it affects the creation of many elements at once.

@paldepind I'm happy to clarify the test cases, will update shortly, thanks!

paldepind:
@mightyiam So, @tdumitrescu has already given a great answer to your question. Per its specification `indexOf` has to look for occurrences _anywhere_. But we're only interested in whether of not `indexOf` would return `0`. So it wastes resources trying to find occurrences that are of no interest. A since most selectors do not contain the string `"svg"` `indexOf` would in most cases have to check every possible location where `"svg"` could be.

@tdumitrescu Looks good :+1: I agree with your reasoning. Comments are sufficient :smile: Thanks a lot for the PR!

tdumitrescu:
Thanks @paldepind. Any idea when it will get published to npm? Looks like it's been a few months since a release...

Issue #189, opened 2016-11-02
EventListener interface in eventlisteners.js


hville:
The `eventlistener` module assigns a [property to a listener function](https://github.com/snabbdom/snabbdom/blob/master/modules/eventlisteners.js#L36-L40).

However, the [EventListener interface](https://developer.mozilla.org/en-US/docs/Web/API/EventListener) is already defined so that the listener is an object with a `.handleEvent` method.

This would allow the following simplification:

### Before
```
function createListener() {
  return function handler(event) {
    handleEvent(event, handler.vnode);
  }
}
```

### After
```
function createListener() {
  return {handleEvent: handleEvent}
}
```
The `vnode` is still accessible in `handleEvent` through `this.vnode`.
The benefits are marginal but the code is simpler and there is one less function call on every event for every element.

I can do the PR if there is any interest

hville:
Closing since it has gone 1 month without feedback

PR #219, opened 2017-01-06
call handleEvent if handler is object

Reon90:
According to the spec, we can pass object with handleEvent method to addEventlistener.

### DIFF
--- a/src/modules/eventlisteners.ts
+++ b/src/modules/eventlisteners.ts
@@ -5,7 +5,7 @@ function invokeHandler(handler: any, vnode?: VNode, event?: Event): void {
   if (typeof handler === "function") {
     // call function handler
     handler.call(vnode, event, vnode);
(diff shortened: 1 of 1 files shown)


AlexGalays:
True, although I never saw anyone use this form. Is it really that useful when using a VDOM lib ?
Also, you can create your own modules if the default ones are not exactly how you would want them.

Reon90:
Otherwise, I have to use the bind(this). It looks ugly. 
It's a bad practice in some companies.

AlexGalays:
Agreed about bind(this). You're not using ES6/babel/typescript by any chance? Then the `=>` notation would replace bind(this).

hville:
I sugested the same in #189 and closed it due to lack of interest. I still think it is a better pattern.

mightyiam:
Closing this because #189 is closed. But if you'd like to reopen that issue, feel free to do so.

PR #221, opened 2017-01-11
[perf-optimization] skip unnecessary work in modules when possible

caridy:
Modules are already applying some opt-out rules, but when the specific piece of data used by the module is the same, a couple of unnecessary `for-in`, plus few conditions were evaluated. This is particularly useful if any kind of memoization is used during the creation of the vnode, which is user-land code.

### DIFF
--- a/src/modules/attributes.ts
+++ b/src/modules/attributes.ts
@@ -23,6 +23,7 @@ function updateAttrs(oldVnode: VNode, vnode: VNode): void {
       attrs = (vnode.data as VNodeData).attrs, namespaceSplit: Array<string>;
 
   if (!oldAttrs && !attrs) return;
+  if (oldAttrs === attrs) return;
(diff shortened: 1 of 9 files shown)


staltz:
Looks like a pretty good PR in my opinion. 👍

paldepind:
I agree. This is a very cheap check that can potentially have nice performance benefits. Especially with values that are static.

But I don't understand the added tests. Do they actually test anything new? It looks to me like they should pass both before and after this change?

caridy:
@paldepind yeah, tests are always complicated when there is memoization involved. They certainly pass before and after, I decided to add them to make sure that future changes will honor the behavior when the exact same object is passed.

paldepind:
@caridy Ok. Great. I just wanted to make sure I was understanding the tests correctly :smile: 

Thanks a lot for the great PR! :+1:

PR #224, opened 2017-01-16
Fix hooks file to be TypeScript source

staltz:
In order for other TypeScript code to consume snabbdom's TypeScript
correctly, we must "export" the hooks interfaces just like we export
other types.


staltz:
This was a trivial fix, I'm releasing it asap just to help fix TypeScript projects.

paldepind:
Great 👍

PR #228, opened 2017-01-17
Fix compatibility issue of String.prototype.startsWith

zhulongzheng:
`String.prototype.startsWith` is in ES2015 specification and is not ready for some devices.

### DIFF
--- a/src/modules/style.ts
+++ b/src/modules/style.ts
@@ -21,7 +21,7 @@ function updateStyle(oldVnode: VNode, vnode: VNode): void {
 
   for (name in oldStyle) {
     if (!style[name]) {
-      if (name.startsWith('--')) {
+      if (name.indexOf('--') === 0) {
         (elm as any).style.removeProperty(name);
       } else {
         (elm as any).style[name] = '';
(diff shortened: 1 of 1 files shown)


mightyiam:
This kind of thing is easily solved in the end-product using a polyfill, right?

nunocastromartins:
This is similar to the discussion in #192. Even though it shouldn't matter much performance-wise, `indexOf` does search through the whole string.
Maybe do: `name[0] === '-' && name[1] === '-'`?

zhulongzheng:
But you must read the source code to find what causes this problem, I don't think it's normal users' job.

paldepind:
Oh. That is a mistake. Would you change the PR so the check is done like this:

```js
name[0] === "-" && name[1] === "-"
```

In #192 and #213 a similar thing was discussed.

jbucaran:
@paldepind Why not `name.substr(0, 2) === "--"`?

zhulongzheng:
@paldepind Updated!

staltz:
> Why not name.substr(0, 2) === "--"?

Performance

staltz:
This should be good to merge.

paldepind:
Looks good. Thank you @zhulongzheng :smile:

Issue #142, opened 2016-07-26
Support for comment nodes


r24y:
I'm interested in using snabbdom server-side, but as far as I can tell it lacks the ability to handle comments. This is useful for (among other things) adding browser-conditional comments:

```
<!--[if lte IE 8]><script src="..."></script><!endif-->
```

Other than emitting a `<div>` with `innerHTML` set for a comment, how can I add comments?

paldepind:
Are you using snabbdom-to-html?

I don't think there currently is a way to achieve what you want.

How whould you like for this to work ideally?

katyo:
I have same issue also. In any case we must provide support for comment nodes in html because the DOM supports it too (see [document.createComment()]).

[document.CreateComment()]: https://www.w3.org/TR/REC-DOM-Level-1/level-one-core.html#method-createComment

r24y:
@katyo great point! I'm changing this issue to be a bit more general; once comment nodes are supported then browser conditionals should "just work" by putting text into comment nodes.

```diff
  Issue #142
- `if lte IE 8` and similar comments
+  Support for comment nodes
```

r24y:
I'm not picky on the API, but for the sake of giving a strawman:

```javascript
h('!', ['comment content'])
```

Of course this may end up needing all sorts of special cases; comments don't have attributes, a user might try to put a hook on a comment node, etc. It would be pretty easy to implement comment nodes, but tough to guard against the ways people can break them in the real world!

paldepind:
I think your proposed API looks very fine :+1: 

I am not too concerned that users can in theory put attributes, hooks, etc. on comment nodes.

paldepind:
This was closed by #229.

PR #229, opened 2017-01-17
Add ability to create comment nodes

pedrosland:
Like some other users in #142, I need to be able to create comment nodes.

This PR adds this feature using similar syntax as proposed before:

```js
h('!', 'comment content')
```

Is a better place for the `vnode.text = ''` in `createElm`?

### DIFF
--- a/src/htmldomapi.ts
+++ b/src/htmldomapi.ts
@@ -2,6 +2,7 @@ export interface DOMAPI {
   createElement: (tagName: any) => HTMLElement;
   createElementNS: (namespaceURI: string, qualifiedName: string) => Element;
   createTextNode: (text: string) => Text;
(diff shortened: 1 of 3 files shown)


staltz:
Looks like a good PR in my opinion.

katyo:
I also found this feature is pretty useful.

staltz:
There has been only positive reviews for this, so I'm merging it.

Thanks!

fix-fix:
I think there should be documentation for this feature, at least as an example in the README, otherwise it will be really hard to discover this syntax.

AlexGalays:
None of the hooks work on a comment node though? Is that on purpose?

Use case:  A virtual component like a popup, returning null locally but rendering some element somewhere else in the proper layer.

PR #230, opened 2017-01-17
Speed-up checking boolean attributes

katyo:
I tried speed-up checking boolean attributes using RegExp instead of lookup key in object.

You can see performance difference for both lookup methods [here](https://gist.github.com/katyo/4c7c469c28b5976e16ca87d89bac382a).

Also I exported Regexp to make it possible use it in other modules (I have one).

*I don't know the generic way to generate regular expression using list of matched words as source. Because I use emacs, so I simply created a script to do that directly in editor.*

### DIFF
--- a/src/modules/attributes.ts
+++ b/src/modules/attributes.ts
@@ -5,17 +5,21 @@ const NamespaceURIs = {
   "xlink": "http://www.w3.org/1999/xlink"
 };
 
-const booleanAttrs = ["allowfullscreen", "async", "autofocus", "autoplay", "checked", "compact", "controls", "declare",
-                "default", "defaultchecked", "defaultmuted", "defaultselected", "defer", "disabled", "draggable",
-                "enabled", "formnovalidate", "hidden", "indeterminate", "inert", "ismap", "itemscope", "loop", "multiple",
-                "muted", "nohref", "noresize", "noshade", "novalidate", "nowrap", "open", "pauseonexit", "readonly",
-                "required", "reversed", "scoped", "seamless", "selected", "sortable", "spellcheck", "translate",
-                "truespeed", "typemustmatch", "visible"];
+/* Boolean attributes regexp generator (elisp)
+   You can edit list of properties and re-evaluate expression
 
-const booleanAttrsDict = Object.create(null);
-for (let i = 0, len = booleanAttrs.length; i < len; i++) {
-  booleanAttrsDict[booleanAttrs[i]] = true;
-}
+   (replace-regexp "\\(booleanAttrsRegex = \\)[^;]*;"
+   (concat "\\1/^" (replace-regexp-in-string "\\\\" "" (regexp-opt '(
+   "allowfullscreen" "async" "autofocus" "autoplay" "checked" "compact" "controls" "declare"
+   "default" "defaultchecked" "defaultmuted" "defaultselected" "defer" "disabled" "draggable"
+   "enabled" "formnovalidate" "hidden" "indeterminate" "inert" "ismap" "itemscope" "loop" "multiple"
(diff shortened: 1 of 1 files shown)


acstll:
I found this:

```js
var regexFromMyArray = new RegExp(myArray.join("|"), 'gi');
```

http://stackoverflow.com/a/28280976

katyo:
@acstll okay, but we need yet to add optimization :)

jbucaran:
@katyo I left you a comment https://gist.github.com/katyo/4c7c469c28b5976e16ca87d89bac382a#gistcomment-1974181

katyo:
It may looks surprising, but testing string using regex really faster than hash(?)/array lookup. At least all mainstream JS engines, which I tested, confirm this. But real increasing of speed is differs from engine to engine. Look at my results below:

| ECMAScript Environment | Speed increase comparing with hash lookup |
| ---- | ---- |
| Firefox: 45.6.0 | 15% |
| Chromium: 55.0.2883.75 | 56% |
| Node: v4.3.1 | 80% |

Of course, the bundle size also decrease when we use regex. It's so because regex is short itself, but optimized regex is yet more short than array with strings or object with keys.

nunocastromartins:
I'd suggest switching to [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), possibly polyfilling it. It's the most semantic option and, as @katyo's test shows, quite performant in the engines that support it. :-)

paldepind:
That is very cool. I would never have guessed that the regexp is faster than using an object.

But using a Set as @nunocastromartins suggests would be a good idea as well. Should the fallback be the current object based solution or something else?

AlexGalays:
How about less magic and just enumarate all the attributes in the regexp without trying to reuse sub words? Would that affect performances much? It's quite hard to read/maintain at the moment.

caridy:
Yes, we need optimizations (micro and micros), yes, they normally come with a price-tag, so we need to be very careful. I usually favor incremental optimizations that can pass a simple tests:

1. does it move the needle?
2. is this the recommended approach or takes us close to the recommended approach?
3. is the code still readable by authors?
4. is the code still debuggable by consumers?

Take this with a grain of salt since it really depend on the project. In this case, I think this PR will not pass the test 2, or 3 or 4, knowing that the `Set` will continue improving in evergreen browsers.

I don't know how many dictionaries do we need in Snabbdom, but if there is more that this one, we could probably just have a very dummy polyfill of Set's basic features, and go with that. Another alternative here is to try to identify why is that object de-optimized by the engine.

acstll:
Since (correct me if I'm wrong) we only need to do a check, I think something like this could work

```js
const Set = window.Set || PseudoSet

function PseudoSet (array) {
  this.data = array
}

PseudoSet.prototype.has = function (value) {
  if (typeof Array.prototype.find === 'function') {
    return this.data.find(x => x === value) !== void 0
  }
  return this.data.indexOf(value) > -1
}
```

It's been some time since I used this so I don't remember exactly, but I think `Array#find` was faster than `indexOf`.

caridy:
@acstll something like that... we can iterate on it, but first lets find out if there are other areas in which a Set should be used in the entire library. As for the forking logic on fallback implementations, in general, if you are hitting the fallback mechanism, you're already in the slow-path ;) where "compat" is the main goal, there is not need to optimize the slow-paths considering that the fast-path is already implemented in all major browsers.

paldepind:
In #254 a [new way of handling boolean attributes](https://github.com/snabbdom/snabbdom/issues/254#issuecomment-284693378) was proposed. Essentially whenever an attribute is `true` or `false` it would be handled like a boolean attribute. Please share your opinion over in #254.

paldepind:
The technique in this PR is really cool :+1:. But https://github.com/snabbdom/snabbdom/pull/314 removed boolean attribute checking altogether which I think is even better.

Issue #226, opened 2017-01-16
Filter out null/undefined children


staltz:
In Cycle.js we roll our own hyperscript just for one reason: we filter out falsy (null or undefined or empty string) children. This is the only modification we do compared to snabbdom's hyperscript:

```diff
 h(sel: string, b?: any, c?: any): VNode {
  let data = {};
  let children: Array<VNode | string> | undefined;
  let text: string | undefined;
  if (arguments.length === 3) {
    data = b;
    if (is.array(c)) {
      children = c;
    } else if (is.primitive(c)) {
      text = c as string;
    }
  } else if (arguments.length === 2) {
    if (is.array(b)) {
      children = b;
    } else if (is.primitive(b)) {
      text = b as string;
    } else {
      data = b;
    }
  }
  if (is.array(children)) {
+   children = children.filter(x => !!x);
    for (let i = 0; i < children.length; ++i) {
      if (is.primitive(children[i])) {
        children[i] = vnode(undefined as any, undefined, undefined, children[i] as any, undefined);
      }
    }
  }
  if (sel[0] === 's' && sel[1] === 'v' && sel[2] === 'g') {
    addNS(data, children, sel);
  }
  return vnode(sel, data, children, text, undefined);
};
```

We're considering moving this change to snabbdom itself. So, question: are there any strong reasons not to do this?

nunocastromartins:
I'm just going to raise a few questions here, but I'm actually leaning against your proposal in favour of something else. Consider:
```js
let children = [h('i', 0), null, h('i', 2)];
let parent = h('div', children);
parent.children.length === children.length; // I'd say this should hold

// I'd also like to be able to do this:
patch(parent.children[1], h('i', 1));
```
I'm not sure whether Snabbdom is _supposed_ to be used like this, but this would allow you to patch a single child without patching the whole `parent`. I'd prefer if Snabbdom rendered `null`/`undefined` values as some "placeholder" node that could afterwards be patched (I'm not actually sure, but I don't think it currently does this) - I'd say render them as empty text nodes.

Your approach does, however, have a thing I appreciate. I don't know if it's done on purpose but, by calling `children.filter`, you're actually creating a new array. This is good because afterwards the original array isn't changed!

```js
let children = [0, 1, 2];
let node = h('div', children);
console.log(children[1]); // What should this print?
```
[...]

PR #232, opened 2017-01-19
Support for null/undefined children

nunocastromartins:
Changed the patching algorithm to support `null`/`undefined` children as I described in #226.

This should make the core algorithm slightly more resilient, but maybe changing the `h` function would be simpler (as was originally proposed). :-)

Tell me your thoughts. Maybe this needs more tests? Lazily copied the one @staltz wrote.

### DIFF
--- a/test/core.js
+++ b/test/core.js
@@ -232,7 +232,9 @@ describe('snabbdom', function() {
     });
     describe('updating children with keys', function() {
       function spanNum(n) {
-        if (typeof n === 'string') {
+        if (n == null) {
+          return n;
+        } else if (typeof n === 'string') {
           return h('span', {}, n);
         } else {
           return h('span', {key: n}, n.toString());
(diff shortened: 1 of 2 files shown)


test/core.js
```diff
@@ -230,6 +230,14 @@ describe('snabbdom', function() {
...
       assert.equal(elm.src, undefined);
     });
+    it('supports some null children', function() {
+      var vnode1 = h('span', [h('span', 'hi'), h('span', 'snabbdom')]);
+      var vnode2 = h('span', [h('span', 'hi'), null, h('span', 'snabbdom')]);
+      elm = patch(vnode0, vnode1).elm;
+      assert.equal(elm.children.length, 2);
+      elm = patch(vnode1, vnode2).elm;
+      assert.deepEqual(map(inner, elm.children), ['hi', 'snabbdom']);
+    });
```
mightyiam: I think there should be a test for `undefined` as well.

staltz:
Might be a good opportunity to replace all `isDef` and `isUndef` cases.

nunocastromartins:
> I think there should be a test for `undefined` as well.

I agree! I'll commit something in a bit.

> Might be a good opportunity to replace all `isDef` and `isUndef` cases.

Just so we're on the same page, you mean replace all occurrences of `isDef` and `isUndef` with `!= null` and `== null` checks? If so, then I agree.

staltz:
We need @paldepind's opinion about that.

paldepind:
I think this is a good idea and I think this PR looks like a good approach 👍

I think adding a few more to the tests may be nice. For instance in the edge case where all child values are `null`/undefined`. And maybe also a case where an element is toggled on and off. I.e. first it is a real vnode, then `undefined` and then a real vnode again.

staltz:
By the way `foo === void 0` might be better than `foo === undefined`. A bit shorter.

nunocastromartins:
> I think adding a few more to the tests may be nice. For instance in the edge case where all child values are `null`/`undefined`. And maybe also a case where an element is toggled on and off. I.e. first it is a real vnode, then `undefined` and then a real vnode again.

Agree.

> I personally prefer `foo === undefined` when just `undefined` is expected and to only use `foo == null` when both `undefined` and `null` are expected.

Just to be clear, the only place where `null` is expected is as a child, correct?

> By the way `foo === void 0` might be better than `foo === undefined`. A bit shorter.

I don't see a problem with this. `void 0` should also be safer, though I'm not sure that matters nowadays.

nunocastromartins:
> And maybe also a case where an element is toggled on and off. I.e. first it is a real vnode, then `undefined` and then a real vnode again.

@paldepind I _think_ this is being tested. Tell me if you'd like some other test.

Regarding the change from `undefined` to `void 0`, I abstained from doing so in this latest commit; if you'd like me to change it, tell me so. :-)

paldepind:
I'm fairly certain that Uglify transforms `undefined` to `void 0`. But with regards to code size I think keeping `isUndef` is best since the function call can be made really short.

nunocastromartins:
So you'd rather keep the `isDef`/`isUndef` checks?

staltz:
@nunocastromartins could you add back the `isDef`/`isUndef`? We could discuss that in another issue/PR, but it would be nice to get this PR merged for the purpose of supporting null children.

nunocastromartins:
@staltz, readded `isDef`/`isUndef` and rebased PR.

paldepind:
Thank you @nunocastromartins :+1:

PR #234, opened 2017-01-23
Create element workflow should align with update element workflow

caridy:
I have noticed that there is a slightly different between the create and update workflows, specifically, during the update process, the element itself is updated first (by calling the update callback), then the children collection is updated, but during the creation, this process is the inverse, and all children are created before the element itself is updated (by calling the create callback).

Considering that for the majority of the modules, the create and update hooks are the same, it makes it difficult to predict if the children collection can still be used in any way. Think of a vnode with a hook to validate, and sanitize the children collection, in which case the process will work only when patching against an existing tree, but not when a new tree is created.

The change itself seems to be trivial, and all tests are still passing. Not sure how to test this in a reliable way.

### DIFF
--- a/src/snabbdom.ts
+++ b/src/snabbdom.ts
@@ -92,6 +92,7 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
                                                                                : api.createElement(tag);
       if (hash < dot) elm.id = sel.slice(hash + 1, dot);
       if (dotIdx > 0) elm.className = sel.slice(dot + 1).replace(/\./g, ' ');
+      for (i = 0; i < cbs.create.length; ++i) cbs.create[i](emptyNode, vnode);
       if (is.array(children)) {
         for (i = 0; i < children.length; ++i) {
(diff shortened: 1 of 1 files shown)


diervo:
I run into this issue as well some time back and workaround it. I think normalizing the lifecycle and hooks make things more predictable specially while make recursion. And we have some perf benefits as a side effect.

caridy:
@paldepind did you get a chance to look at this PR?

caridy:
@paldepind @staltz can you guys look at this? do you need more info? what else can we do to get this in?

staltz:
I tested out this change in Cycle.js tests (they are rather extensive) and everything still worked. I think it also makes sense to do this change, so I'm favorable.

caridy:
@staltz that's great news. I'm eager to get this patch in, it will help us to remove some of the hacks we have in place today to accomodate this small discrepancy.

staltz:
Yeah, seems okay to merge, it's quite old PR and everything looks green.

PR #239, opened 2017-02-02
Fix dataset module to support old browsers

staltz:
At least IE10 does not support the elm.dataset property, so we must use
elm.removeAttribute and elm.setAttribute instead.

I found this issue when testing Cycle.js against IE10.

### DIFF
--- a/src/modules/dataset.ts
+++ b/src/modules/dataset.ts
@@ -1,8 +1,10 @@
 import {VNode, VNodeData} from '../vnode';
 import {Module} from './module';
 
+const CAPS_REGEX = /[A-Z]/g;
+
 function updateDataset(oldVnode: VNode, vnode: VNode): void {
-  var elm: HTMLElement = vnode.elm as HTMLElement,
+  let elm: HTMLElement = vnode.elm as HTMLElement,
     oldDataset = (oldVnode.data as VNodeData).dataset,
     dataset = (vnode.data as VNodeData).dataset,
     key: string;
@@ -11,15 +13,24 @@ function updateDataset(oldVnode: VNode, vnode: VNode): void {
   if (oldDataset === dataset) return;
   oldDataset = oldDataset || {};
   dataset = dataset || {};
+  const d = elm.dataset;
 
   for (key in oldDataset) {
     if (!dataset[key]) {
-      delete elm.dataset[key];
+      if (d) {
+        delete d[key];
+      } else {
+        elm.removeAttribute('data-' + key.replace(CAPS_REGEX, '-$&').toLowerCase());
+      }
     }
   }
   for (key in dataset) {
     if (oldDataset[key] !== dataset[key]) {
-      elm.dataset[key] = dataset[key];
+      if (d) {
+        d[key] = dataset[key];
+      } else {
+        elm.setAttribute('data-' + key.replace(CAPS_REGEX, '-$&').toLowerCase(), dataset[key]);
+      }
     }
   }
 }


TylorS:
:+1:

src/modules/dataset.ts
```diff
@@ -14,12 +14,20 @@ function updateDataset(oldVnode: VNode, vnode: VNode): void {
 
   for (key in oldDataset) {
     if (!dataset[key]) {
-      delete elm.dataset[key];
+      if (elm.dataset) {
+        delete elm.dataset[key];
+      } else {
+        elm.removeAttribute('data-' + key);
```
caridy: keys used via `elm.dataset` are not reflective to the data attribute keys, they need transformation according to the spec. Here is an example:

```js
$0.setAttribute('data-foo-bar', 3);
// yield: DOMStringMap {fooBar: "3"}
$0.setAttribute('data-fooBar', 3);
// yield: DOMStringMap {fooBar: "3", foobar: "3"}
```

The proper transformation here is to apply a regexp to the key to make it a data key, e.g.:

```js
const CAPS_REGEX = /[A-Z]/g;
elm.removeAttribute('data-' + key.replace(CAPS_REGEX, (match: string): string => '-' + match.toLowerCase()));
```

something along those lines should be enough. yes, it will be a little bit slower for some old browsers, but the fast path is what really matters.

src/modules/dataset.ts
```diff
@@ -14,12 +14,20 @@ function updateDataset(oldVnode: VNode, vnode: VNode): void {
...
+      }
     }
   }
   for (key in dataset) {
     if (oldDataset[key] !== dataset[key]) {
-      elm.dataset[key] = dataset[key];
+      if (elm.dataset) {
+        elm.dataset[key] = dataset[key];
+      } else {
+        elm.setAttribute('data-' + key, dataset[key]);
```
caridy: same as above.

src/modules/dataset.ts
caridy: keys used via `elm.dataset` are not reflective to the data attribute keys, they need transformation according to the spec. Here is an example:

```js
$0.setAttribute('data-foo-bar', 3);
// yield: DOMStringMap {fooBar: "3"}
$0.setAttribute('data-fooBar', 3);
// yield: DOMStringMap {fooBar: "3", foobar: "3"}
```

The proper transformation here is to apply a regexp to the key to make it a data key, e.g.:

```js
const CAPS_REGEX = /[A-Z]/g;
elm.removeAttribute('data-' + key.replace(CAPS_REGEX, (match: string): string => '-' + match.toLowerCase()));
```

something along those lines should be enough. yes, it will be a little bit slower for some old browsers, but the fast path is what really matters.

nunocastromartins: Slightly shorter (but possibly less efficient): `key.replace(CAPS_REGEX, '-$&').toLowerCase()`

caridy: less efficient on the non-critical path is just fine, I like yours better @nunocastromartins.

staltz:
Updated to make data attribute keys computed from the dataset keys.

src/modules/dataset.ts
```diff
@@ -2,10 +2,11 @@ import {VNode, VNodeData} from '../vnode';
 import {Module} from './module';
 
 function updateDataset(oldVnode: VNode, vnode: VNode): void {
-  var elm: HTMLElement = vnode.elm as HTMLElement,
+  let elm: HTMLElement = vnode.elm as HTMLElement,
     oldDataset = (oldVnode.data as VNodeData).dataset,
     dataset = (vnode.data as VNodeData).dataset,
-    key: string;
+    key: string,
+    CAPS_REGEX = /[A-Z]/g;
```
TylorS: You could lift this out of the function so it's only ever constructed one time.

caridy: yeah, this regexp should be a `const` binding at the module level.

src/modules/dataset.ts
```diff
@@ -14,12 +15,22 @@ function updateDataset(oldVnode: VNode, vnode: VNode): void {
...
     }
   }
   for (key in dataset) {
     if (oldDataset[key] !== dataset[key]) {
-      elm.dataset[key] = dataset[key];
+      if (elm.dataset) {
+        elm.dataset[key] = dataset[key];
+      } else {
+        key = key.replace(CAPS_REGEX, '-$&').toLowerCase();
+        elm.setAttribute('data-' + key, dataset[key]);
```
caridy: This is incorrect, the transformed `key` should only we used as the attribute's key, you can't use it for `dataset[key]`. I think we need tests for this to make sure we are doing the right thing.

caridy: `elm.setAttribute('data-' + key.replace(CAPS_REGEX, '-$&').toLowerCase(), dataset[key]);`

staltz: Right, my bad there

staltz:
Fixed usage of `key` as @caridy mentioned.

caridy:
Another small optimization for the critical path will be to allocate the value of `elm.dataset` into a variable, then check the variable, and remove or add from that variable rather than doing thru `elm.dataset` a second time:

```js
       const d = elm.dataset;
       if (d) {
         delete d[key];
       } else {
         elm.removeAttribute('data-' + key.replace(CAPS_REGEX, '-$&').toLowerCase());
       }
```

That will probably be closer to what we have today in terms of perf, and DOM getters/accessors have been known for how slow they are :).

staltz:
Good point, I updated it again.

staltz:
Thanks folks for reviewing.

PR #240, opened 2017-02-02
Fix style module to support old browsers

staltz:
Avoid using String.prototype.startsWith because it is not supported in
browsers like IE10. Also, this commit uses a similar technique already
present in snabbdom (e.g. in the hyperscript implementation, checking
for svg).

Found this issue when testing Cycle.js against IE10.


TylorS:
:+1:

zhulongzheng:
Duplicate with [#228](https://github.com/snabbdom/snabbdom/pull/228).

staltz:
Thanks for telling @zhulongzheng

Issue #167, opened 2016-09-29
Patching should clear out previous root element content


amirouche:
Given the following html:

``` html
  <div id="app">should disappear</div>
```

And the following js code

``` javascript
var patch = snabbdom.init([
    snabbdom_style,
    snabbdom_class,
    snabbdom_props,
    snabbdom_attributes,
    snabbdom_eventlisteners
]);

var vnode = h('div', [" but it", " doesn't"]);
var container = document.getElementById('app');

patch(container, vnode);
```

`patch` will generate the following html:

``` html
<div id="app">should disappear but it doesn't</div>
```

One can work around this by setting an empty id for instance:

``` javascript
var vnode = h('div#', [" but it", " doesn't"]);
```

amirouche:
Here is a jsbin http://jsbin.com/sovesitade/1/edit?html,output

amirouche:
Also the `id` is wrong

amirouche:
It's not possible to reproduce in the test suite. The output of patch is correct but the DOM is not.

amirouche:
Here is another bug with similar behavior. With the same html as above, patching the div with the following `vnode`:

```javascript
var vnode = h('div', {attrs: {id: "app"}}, [" but it", " doesn't"]);
```

Won't work.

But the same call with `div#` seem to give the good result.

WorldMaker:
This seems to be a regression between 0.5.0 and 0.5.3, if that helps narrow down the cause.

paldepind:
Thank you for reporting this @amirouche. Also thanks to @WorldMaker for the additional information.

This is probably related to some recent changes to how `patch` with an element as it's first argument is done.

AlexGalays:
Could @staltz help perhaps, since the regression appears to come from that change?

staltz:
I'll look into it, but probably after the TypeScript migration.

staltz:
Done and released in v0.6.4

avesus:
Thank you!

PR #242, opened 2017-02-06
Introduce toVNode() to reconstruct root element as vnode 

staltz:
This commit addresses issue #167. 

Previously, in snabbdom v0.5.0, patch(element, vnode) would always create a new element for the root. This resulted in problems with custom elements (web components), and was fixed in commit c091c59.

However, that commit resulted in bug #167. This meant that snabbdom would have bugs with server-side rendering, where the root element would be non-empty (it has many children, rendered on the server-side as HTML), and the client-side rendering should reuse those existing children (or clear all the children and recreate them again in patch()).

This commit introduces the function toVNode(elm) that deep-converts an element (and its tree) to a VNode (and its tree), that is separately imported and used before calling patch(). toVNode(elm) will look at the element's attributes and gather those as data for the vnode.

Overall, this commit is important for fixing #167 and enabling client/server-side rendering in an efficient manner (destruction/recreation client-side is probably too expensive).

**Would appreciate reviews as soon as possible since this is a critical bug that is blocking development of a new version in Cycle.js**

### DIFF
--- a/test/attributes.js
+++ b/test/attributes.js
@@ -44,6 +44,18 @@ describe('attributes', function() {
     elm = patch(vnode0, vnode1).elm;
     assert.strictEqual(elm.getAttributeNS('http://www.w3.org/1999/xlink', 'href'), '#foo');
   });
+  it('should not touch class nor id fields', function() {
+    elm = document.createElement('div');
+    elm.id = 'myId';
+    elm.className = 'myClass';
+    vnode0 = elm;
+    var vnode1 = h('div#myId.myClass', {attrs: {}}, ['Hello']);
+    elm = patch(vnode0, vnode1).elm;
+    assert.strictEqual(elm.tagName, 'DIV');
+    assert.strictEqual(elm.id, 'myId');
+    assert.strictEqual(elm.className, 'myClass');
+    assert.strictEqual(elm.textContent, 'Hello');
+  });
   describe('boolean attribute', function() {
     it('is present if the value is truthy', function() {
       var vnode1 = h('div', {attrs: {required: true, readonly: 1, noresize: 'truthy'}});
@@ -62,16 +74,12 @@ describe('attributes', function() {
   });
   describe('Object.prototype property', function() {
     it('is not considered as a boolean attribute and shouldn\'t be omitted', function() {
-      var vnode1 = h('div', {attrs: {valueOf: true, toString: 1, constructor: 'truthy'}});
+      var vnode1 = h('div', {attrs: {constructor: true}});
       elm = patch(vnode0, vnode1).elm;
-      assert.strictEqual(elm.getAttribute('valueOf'), 'true');
-      assert.strictEqual(elm.getAttribute('toString'), '1');
-      assert.strictEqual(elm.getAttribute('constructor'), 'truthy');
-      var vnode2 = h('div', {attrs: {valueOf: false, toString: 0, constructor: null}});
+      assert.strictEqual(elm.getAttribute('constructor'), 'true');
+      var vnode2 = h('div', {attrs: {constructor: false}});
       elm = patch(vnode0, vnode2).elm;
-      assert.strictEqual(elm.getAttribute('valueOf'), 'false');
-      assert.strictEqual(elm.getAttribute('toString'), '0');
-      assert.strictEqual(elm.getAttribute('constructor'), 'null');
+      assert.strictEqual(elm.getAttribute('constructor'), 'false');
     })
   });
 });
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,9 @@ snabbdom.js.map
 thunk.d.ts
 thunk.js
 thunk.js.map
+tovnode.d.ts
+tovnode.js
+tovnode.js.map
 vnode.d.ts
 vnode.js
 vnode.js.map
--- a/src/htmldomapi.ts
+++ b/src/htmldomapi.ts
@@ -10,6 +10,10 @@ export interface DOMAPI {
   nextSibling: (node: Node) => Node;
   tagName: (elm: Element) => string;
   setTextContent: (node: Node, text: string | null) => void;
+  getTextContent: (node: Node) => string | null;
+  isElement: (node: Node) => node is Element;
+  isText: (node: Node) => node is Text;
+  isComment: (node: Node) => node is Comment;
 }
 
 function createElement(tagName: any): HTMLElement {
@@ -56,6 +60,22 @@ function setTextContent(node: Node, text: string | null): void {
   node.textContent = text;
 }
 
+function getTextContent(node: Node): string | null {
+  return node.textContent;
+}
+
+function isElement(node: Node): node is Element {
+  return node.nodeType === 1;
+}
+
+function isText(node: Node): node is Text {
+  return node.nodeType === 3;
+}
+
+function isComment(node: Node): node is Comment {
+  return node.nodeType === 8;
+}
+
 export const htmlDomApi = {
   createElement,
   createElementNS,
@@ -68,6 +88,10 @@ export const htmlDomApi = {
   nextSibling,
   tagName,
   setTextContent,
+  getTextContent,
+  isElement,
+  isText,
+  isComment,
 } as DOMAPI;
 
 export default htmlDomApi;
--- a/src/tovnode.ts
+++ b/src/tovnode.ts
@@ -1,0 +1,39 @@
+import vnode, {VNode} from './vnode';
+import htmlDomApi, {DOMAPI} from './htmldomapi';
+
+export function toVNode(node: Node, domApi?: DOMAPI): VNode {
+  const api: DOMAPI = domApi !== undefined ? domApi : htmlDomApi;
+  let text: string;
+  if (api.isElement(node)) {
+    const id = node.id ? '#' + node.id : '';
+    const cn = node.getAttribute('class');
+    const c = cn ? '.' + cn.split(' ').join('.') : '';
+    const sel = api.tagName(node).toLowerCase() + id + c;
+    const attrs: any = {};
+    const children: Array<VNode> = [];
+    let name: string;
+    let i: number, n: number;
+    const elmAttrs = node.attributes;
+    const elmChildren = node.childNodes;
+    for (i = 0, n = elmAttrs.length; i < n; i++) {
+      name = elmAttrs[i].nodeName;
+      if (name !== 'id' && name !== 'class') {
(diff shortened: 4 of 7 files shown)


src/snabbdom.ts
```diff
@@ -57,10 +57,27 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
+      attrs[elmAttrs[i].nodeName] = elmAttrs[i].nodeValue;
+    }
+    delete attrs.id;
+    delete attrs.class;
+    const children: Array<VNode | string> = [];
+    for (let i = 0, elmChildren = elm.childNodes, n = elmChildren.length; i < n; i++) {
+      const node = elmChildren[i];
+      if (node.nodeType === 1) {
+        children.push(toVNode(elmChildren[i] as Element));
+      } else if (node.nodeType === 2) {
```
nunocastromartins: Did you test this? `Text` nodes should have a `nodeType` of `3`.
https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType#Node_type_constants

Also, I'd consider creating a `nodeType` function in `htmldomapi.ts` instead, to be consistent with how the `tagName` is accessed. Maybe also declare the `1` and `3` as constants there?

staltz: 👍

staltz:
Fixed now with @nunocastromartins 's suggestions.

src/modules/attributes.ts
```diff
@@ -26,28 +26,31 @@ function updateAttrs(oldVnode: VNode, vnode: VNode): void {
...
+  const lowAttrs: any = {};
 
   // update modified attributes, add new attributes
   for (key in attrs) {
     cur = attrs[key];
-    old = oldAttrs[key];
+    const lowkey = key.toLowerCase();
+    lowAttrs[lowkey] = cur;
+    old = oldAttrs[lowkey];
     if (old !== cur) {
```
AlexGalays: For subsequent patchings (a VNode against a VNode), won't that compare the lowerCased current attribute with the non lowerCased previous attribute and thus do a wasteful setAttribute() ?

src/snabbdom.ts
```diff
@@ -57,10 +57,29 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
+    delete attrs.class;
+    const children: Array<VNode> = [];
+    for (let i = 0, elmChildren = elm.childNodes, n = elmChildren.length; i < n; i++) {
+      const node = elmChildren[i];
+      if (api.isElement(node)) {
+        children.push(toVNode(node));
+      } else if (api.isText(node)) {
+        children.push(
+          (vnode as any)(undefined, undefined, undefined, node.textContent)
+        );
```
nunocastromartins: Now that Snabbdom supports comment nodes, maybe you should also check for them (`nodeType` is `8`)?

src/snabbdom.ts
```diff
@@ -57,10 +57,29 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
     }
   }
 
-  function emptyNodeAt(elm: Element) {
+  function toVNode(elm: Element): VNode {
```
nunocastromartins: Why not make this a bit more generic and accept `Node` instead of `Element`? This would make it possible to patch a virtual node against a `Text` node (and since a recent PR, against `Comment` nodes too).

src/snabbdom.ts
```diff
@@ -57,10 +57,29 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
+    delete attrs.id;
+    delete attrs.class;
+    const children: Array<VNode> = [];
+    for (let i = 0, elmChildren = elm.childNodes, n = elmChildren.length; i < n; i++) {
+      const node = elmChildren[i];
+      if (api.isElement(node)) {
+        children.push(toVNode(node));
+      } else if (api.isText(node)) {
+        children.push(
+          (vnode as any)(undefined, undefined, undefined, node.textContent)
```
nunocastromartins: I'd also put a `getTextContent` function in the API, in conformity with `setTextContent`.

src/snabbdom.ts
```diff
@@ -57,10 +57,29 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
-    const c = elm.className ? '.' + elm.className.split(' ').join('.') : '';
-    return vnode(api.tagName(elm).toLowerCase() + id + c, {}, [], undefined, elm);
+    const cn = elm.getAttribute('class');
+    const c = cn ? '.' + cn.split(' ').join('.') : '';
+    const sel = api.tagName(elm).toLowerCase() + id + c;
+    const attrs: any = {};
+    for (let i = 0, elmAttrs = elm.attributes, n = elmAttrs.length; i < n; i++) {
+      attrs[elmAttrs[i].nodeName] = elmAttrs[i].nodeValue;
+    }
+    delete attrs.id;
```
caridy: `delete` might have undesirable "perf" consequences (lead to deoptimization) in dictionary-like objects like `attrs`. Instead of adding the keys and then delete them immediately after, just opt-out of the addition of the key in the first place (maybe an `if` statement in line 67). Btw, I'm not sure this is a critical path for the engine where the `delete` will have an impact.

src/modules/attributes.ts
```diff
@@ -26,28 +26,31 @@ function updateAttrs(oldVnode: VNode, vnode: VNode): void {
   if (oldAttrs === attrs) return;
   oldAttrs = oldAttrs || {};
   attrs = attrs || {};
+  const lowAttrs: any = {};
 
   // update modified attributes, add new attributes
   for (key in attrs) {
     cur = attrs[key];
-    old = oldAttrs[key];
+    const lowkey = key.toLowerCase();
```
caridy: why lowercasing the keys in `attrs` when attribute keys are suppose to be in lowercase in the first place? snabbdom is a low-level API, and whoever relies on `attrs` must ensure that attribute keys (per spec) are lowercased. Seems to me that we are slowing down the critical path to accomodate a very edge case in which someone uses snabbdom incorrectly.

caridy: if this is because `toVNode()` and not really about folks using snabbdom incorrectly, then the lowercasing should happen during `toVNode()` routine instead instead of doing it during the critical path.

AlexGalays: Agreed. We can always add the most common attributes to the typings in lowercase form too, to limit errors (and people usually know that custom data attributes are in lower case). Personally, I will never use server-side rendering but with this change I would still have to pay the price of lowercasing every single attribute of my app, on every render (twice).

staltz: My problem was this test https://github.com/snabbdom/snabbdom/blob/master/test/attributes.js#L65 which uses snabbdom "incorrectly" as you mentioned.

caridy: then, lets clarify the api, where `attrs` is suppose to be key value pairs where the key is suppose to be a valid html attribute name (per spec), and let's fix the test to provide the right form.

staltz:
- Changed `toVNode(elm: Element): VNode` to `toVNode(elm: Node): VNode` as @nunocastromartins requested
- Made `toVNode` check nodes of type Comment too, @nunocastromartins 
- Created `DOMAPI.getTextContent` to match `setTextContent`, @nunocastromartins 
- Replaced `delete` with `if` blacklist, as @caridy requested

staltz:
Also, I reverted changes to the Attributes Module. I update the attributes test to avoid using camel-cased attributes.

test/attributes.js
```diff
@@ -62,15 +74,11 @@ describe('attributes', function() {
   });
   describe('Object.prototype property', function() {
     it('is not considered as a boolean attribute and shouldn\'t be omitted', function() {
```
staltz: For the purpose of just checking Object.prototype properties, it's enough to test for the `constructor` property. Attributes like `valueOf` are invalid since they contain an uppercase letter.
Requesting @fix-fix to review this since they wrote this test.

TylorS:
From a more general stance of type-safety, I think it'd be even better to have this function be required for the user to make use of, which in turn will allow patch to have a monomorphic type signature. For example

```typescript
const element = document.querySelector('#root') as Element

const rootVNode: VNode = toVNode(element)

patch(rootVNode, h('div', {}, [ ... ])
```

If anyone agrees with me, this *is* a breaking change, I think the types can be greatly improved from there as well, but I'll avoid going into that right now.

staltz:
Nice idea @TylorS. I would discuss that in a separate issue only after this PR is merged, so we don't delay merging this too much, but I like it.

TylorS:
Fain enough, opening a separate issue.

src/snabbdom.ts
```diff
@@ -60,10 +60,38 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
     }
   }
 
-  function emptyNodeAt(elm: Element) {
-    const id = elm.id ? '#' + elm.id : '';
-    const c = elm.className ? '.' + elm.className.split(' ').join('.') : '';
-    return vnode(api.tagName(elm).toLowerCase() + id + c, {}, [], undefined, elm);
+  function toVNode(elm: Node): VNode {
+    let text: string;
+    if (api.isText(elm)) {
```
AlexGalays: Nitpicking but I suspect `toVNode` will be called with an Element rather than a text or comment, most of the time. Wouldn't it make more sense to start with `if (api.isElement(elm))`?

staltz: That's smart, I'll do that

paldepind:
This is definitely an improvement when using server side rendering 😄 

I think I prefer keeping `emptyNodeAt` and moving `toVNode` to a seperate file. If one does not use server side rendering then `toVNode` isn't very useful. Furthermore it relies on the attributes module. Both are reasons for having it in a seperate file in my opinion.

Besides the above I think the PR is wonderful and a very welcome feature addition.

staltz:
Good point, I can move it to a separate file. However, it needs the `api` object. How do you recommend that `toVNode` depend on that? 'Optional with default', like `init` has?

test/attributes.js
fix-fix: @staltz LGTM, even testing for single property is enough. Though it might be better to check using `true`/`false` instead of `'truthy'`/`null` for this single case to make it more obvious.

paldepind:
'Optional with default' sounds good to me 👍

staltz:
I made `toVNode` a separate file, @paldepind.
Updated the attributes test to use true/false, @fix-fix.

staltz:
Thanks :)

PR #246, opened 2017-02-13
h() should accept an Array<VNode | string | undefined> as its children

naruaway:
From TypeScript's side, it seems that ```h()``` only accepts an ```Array<VNode>``` as its ```children```.
But its actual implementation can deal with ```string```s and some tests are covering these cases.
Because these tests are written in JavaScript, they cannot detect type mismatch.

For example,
```typescript
h('div', [ 'Hello,', h('br'), 'World!' ])
```
is invalid in TypeScript but can be correctly executed in JavaScript.
This PR fixes the type mismatch so that ```h()``` takes ```children``` as an ```Arrray<VNode | string>```.

By the way, this kind of type mismatch may suggest that tests should also be written in TypeScript...

### DIFF
--- a/src/h.ts
+++ b/src/h.ts
@@ -15,10 +15,12 @@ function addNS(data: any, children: Array<VNode> | undefined, sel: string | unde
 
 export function h(sel: string): VNode;
 export function h(sel: string, data: VNodeData): VNode;
+export function h(sel: string, child: VNode): VNode;
 export function h(sel: string, text: string): VNode;
-export function h(sel: string, children: Array<VNode | string>): VNode;
+export function h(sel: string, children: Array<VNode | string | undefined>): VNode;
+export function h(sel: string, data: VNodeData, child: VNode): VNode;
 export function h(sel: string, data: VNodeData, text: string): VNode;
-export function h(sel: string, data: VNodeData, children: Array<VNode | string>): VNode;
+export function h(sel: string, data: VNodeData, children: Array<VNode | string | undefined>): VNode;
 export function h(sel: any, b?: any, c?: any): VNode {
   var data: VNodeData = {}, children: any, text: any, i: number;
   if (c !== undefined) {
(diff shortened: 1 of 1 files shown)


staltz:
LGTM

nunocastromartins:
Can't it also deal with `number`s and `null`/`undefined`?

naruaway:
Thank you for your immediate responses :)
I love snabbdom and its community very much!

I modified my PR to be more correct.
Again I did not touch on any implementation and I just changed ```h()```'s type signature to match its actual implementation.
# For ```null``` / ```undefined```
I just forgot to even think about that...
I think these types should also be accepted in TypeScript.

# For ```number```s
In the current test cases, there is no execution path that ```children``` is containing any ```number```s.
I checked this by inserting below code
```typescript
  is.array(children) && children.forEach((x: any) => {
    if (typeof x === 'number') {
      throw 'number is used in children';
    }
  })
```
just before the ```return``` statement in ```h()``` and running the test.

But code like ```h('span', ['The number is ', 123])``` can be used as expected.
I think this implicit type coercion behavior is sometimes useful.
But from TypeScript's perspective, explicit typings are preferred.
So I personally think ```h()``` should NOT accept ```number```s in TypeScript
but it is OK to accept them in JavaScript.

# For a single child```VNode```
In a test case, there is a code like ```h('div', h('span#hello'))``` , 
But there is no corresponding definition in ```h()```'s type signature.
So I think this case should also be considered.

caesarsol:
Wouldn't it make sense to have the tests written in Typescript?

mightyiam:
@naruaway, thank you for this PR. I apologize that it did not receive the proper attention.

It seems that your desired functionality was introduced:

https://github.com/snabbdom/snabbdom/blob/a42c31a300b7d689ad8a0b424afac67636da54d7/src/h.ts#L5-L25

So I will close this PR, if you don't mind. I welcome your participation in the future.

@caesarsol the tests have been converted to TypeScript since.

PR #247, opened 2017-02-14
[WIP] Fixes #225: removing default exports alltogether

### LINKED ISSUE #225: [Discussion] ambiguity between default export and named exports
I'm not sure about reason to have everything exported as a default export and a corresponding named export (I could not find a discussion around that decision), but I hope that can be revisited. In general, the way we designed the ESM was to avoid such as ambiguity. In theory, there is no problem with exporting the same thing using multiple names (including default), but in practical terms, there are:

* any consumer trying to do detection may fail (this is the case of `rollup`, if you try to `import` or `require` `snabbdom` from a project that relies on the popular rollup bundler, it will fail because rollup applies detection to normalize the exports by testing against the `default` export, assuming regular CJS modules will not have them, and transpiled CJS will have them but with the namespace object.
* we have enough issues with CJS and modules (which we are trying to solve via few actives initiatives at TC29 and Node CTC), we have arrived to the conclusion that the lower common denominator is to let transpilers (babel and ts) to compiled ESM to a simple CJS module. Last time we spoke to TS folks, they were onboard with this idea, I'm not sure why snabbdom needs the complexity of the duplicated entries after adopting TS.
* Last, but not least, it is very confusing for the consumer, having a CJS that contains two values, default and named export with the same thing (`require('snabbdom/h').default === require('snabbdom/h').h`.

Can someone explains the reasons of the duplication? I see a few options:

* refactor to either use the default export only or the named export only (I think named exports are just fine for snabbdom).
* try to work with TS folks and rollup folks to make these ambiguous modules compiled with TS to work with rollup.

@DanielRosenwasser, @Rich-Harris can you guys chime on this?

caridy:
## TL;DR

* This library now contains a single module that you can import or require: `"snabbdom"`.
* Lockdown API for this package: `init()`, `h()`, `thunk()`, `toVNode()`. 
* Convenience public submodules are available via `"snabbdom/modules/*"`, which contains a default export with the Module object.

## Snabbdom Public API

```js
const {init, h, thunk, toVNode} = require('snabbdom');
```

This is equivalent to:

```js
import {init, h, thunk, toVNode} from "snabbdom";
```

_Note: documentation and README.md will have to be updated if this gets merged._ 

## Snabbdom Modules API

```js
var snabbdom = require('snabbdom');
var patch = snabbdom.init([
  require('snabbdom/modules/class').default,
  require('snabbdom/modules/props').default,
  require('snabbdom/modules/style').default,
  require('snabbdom/modules/eventlisteners').default,
]);
```

This is equivalent to:

```js
import {init} from 'snabbdom';
import classModule from 'snabbdom/modules/class';
import propsModule from 'snabbdom/modules/props';
import styleModule from 'snabbdom/modules/style';
import eventModule 'snabbdom/modules/eventlisteners';
var patch = init([
  classModule,
  propsModule,
  styleModule,
  eventModule,
]);
```

_Note: No API changes here based to the documentation._

## Changes

* Remove all the default exports from `src/*.ts` files.
* Rename `src/snabbdom.ts` to `src/init.ts`.
* Add a new main file `src/snabbdom.ts` that re-exports all the commonly used modules (init, h, thunk and toVNode).
* Remove all explicit named exports from `src/snabbdom/modules/*.ts` files.
* remove the unnecessary submodules from the build script.

## Impact

* This change might breaks folks that were importing from anything other than a module name `"snabbdom"` or `snabbdom/moduels/*`. E.g., if you are using `require('snabbdom/h')`, you should now use `require('snabbdom').h`.
* Implementation details should be just that, implementation details, and people should not assume that every file in this package is just part of the public API of the library. With this change, the API is locked down to what is described above in this issue.
* In old versions of this library the assumption was that you could use the pieces of it that you need, instead of having to bring the whole library (which in itself is already pretty small). Today, tree-shaking is standard equipment in bundlers like webpack, rollup, etc. When bundling your project that uses snabbdom, the bundler or minifier can cut those pieces that are not really needed (e.g.: `toVNode()` method). This means that we now can have two main pieces, the library itself, and a set of snabbdom submodules that you can use during the initialization.

## Open questions

* There are many inconsistencies with the names, specifically in modules, e.g.: modules/attributes.ts vs modules/props.ts, which exports some different names as well. Should we clean that up?


caridy:
@paldepind I tried to collect as much info as possible from issue #225, and tried to stick to the plan. This seems to be a big endeavor in terms of the impact on folks using snabbdom today, and a good communication in terms of that will be needed if we want to carry on with these changes. Nevertheless, I'm willing to get this to completion if that's the case.

caridy:
@paldepind @staltz, this is probably not going to get any traction, it is too disruptive. I will accomodate this to be BC, or almost BC, with a little bit a overhead to upgrade existing projects, and will update this PR.

caridy:
Ok guys, finished the second round, updated the details of the PR, and it seems less corrosive. I still need to fix the build scripts for TS and browserify, but I can do that after I get the thumbs up from @paldepind, @staltz and co. on this.

paldepind:
I'm sorry for the late response. I've been on a small vacation away from email.

Thank you a lot for working on this :clap: : I think this is very important work. And one of the things we need to get done before a `1.0.0` release.

I can't find anything to put a finger on in the PR. It all looks great :+1: I'm usually not a fan of using default exports but I think using them in the modules is quite sensible.

I have a question regarding this:

> This change might breaks folks that were importing from anything other than a module name `"snabbdom"` or `snabbdom/moduels/*`. E.g., if you are using `require('snabbdom/h')`, you should now use `require('snabbdom').h`.

As far as I can tell from the changes in the PR using `require('snabbdom/h')` is still supported. And I think it might still make sense for people who wants to cherry pick stuff and who don't have tree-shaking set up.

> There are many inconsistencies with the names, specifically in modules, e.g.: modules/attributes.ts vs modules/props.ts, which exports some different names as well. Should we clean that up?

I'm not sure I understand what you mean here. Is it that `props` is abbreviated and `attributes` is not?

caridy:
> I have a question regarding this:

>> This change might breaks folks that were importing from anything other than a module name "snabbdom" or snabbdom/moduels/*. E.g., if you are using require('snabbdom/h'), you should now use require('snabbdom').h.

> As far as I can tell from the changes in the PR using require('snabbdom/h') is still supported. And I think it might still make sense for people who wants to cherry pick stuff and who don't have tree-shaking set up.

If that's the case, then we will have to make `snabbdom/h` to have a default export. With this PR, we eliminate default exports on all those files, which means they will have to do `require('snabbdom/h').h` which is not pretty, and a breaking change. Honestly, I don't think cherry picking on those very small pieces is important, if it is important for the consumer, they can do tree-shaking, and considering that the whole thing is already pretty small, I don't think it is really needed.

>> There are many inconsistencies with the names, specifically in modules, e.g.: modules/attributes.ts vs modules/props.ts, which exports some different names as well. Should we clean that up?

> I'm not sure I understand what you mean here. Is it that props is abbreviated and attributes is not?

Yes, I'm referring to the name of the modules, we might want to clean that up too, but not a big deal.

mightyiam:
Sorry if I missed it, but why not have an index that re-exports all the submodules?

Compare
```js
import {init} from 'snabbdom';
import classModule from 'snabbdom/modules/class';
import propsModule from 'snabbdom/modules/props';
import styleModule from 'snabbdom/modules/style';
import eventModule 'snabbdom/modules/eventlisteners';
var patch = init([
  classModule,
  propsModule,
  styleModule,
  eventModule,
]);
```

With
```js
import {init} from 'snabbdom';
import { classModule, propsModule, styleModule, eventListenersModule } from 'snabbdom/modules'
var patch = init([
  classModule,
  propsModule,
  styleModule,
  eventListenersModule,
]);
```

Or are we supporting non-tree-shaking bundlers?

caridy:
@mightyiam yes, we can have an index for all default modules, but that is orthogonal since we don't do that today, and this PR doesn't introduce that either. Most people will pick and match those modules with their own, so I don't think that feature as critical.

mightyiam:
Is there any thought on type exports? I'm using `VNode`, for example.

kuraga:
@caridy , @paldepind ,
Will it be compatible with ESM? Will be snabbdom's files `.mjs`?
Thanks!

mightyiam:
I am trying to review this but there's mental overhead from #601. Let's get that merged, prior.

mightyiam:
Hi, @caridy. Would you like to give it another go?

PR #249, opened 2017-02-23
Update `h` signature; children nodes can now be `undefined` or `null`

ornicar:
Required now that #226 is implemented.


staltz:
I'd rather drop `MaybeVNode` and just write it straight inline

ornicar:
Ok, squashed it in.

ornicar:
This is still required to make use of #226. Anything blocking the merge?

staltz:
Looks all good. Thanks!

Issue #252, opened 2017-02-25
ToVNode incorrectly handles text nodes


Steelfish:
Currently ToVNode creates text nodes for an element by returning a vnode with undefined for everything with text. This causes the following type error to be thrown when snabbdom tries to update the DOM by removing the nodes. 

```
TypeError: Failed to execute 'removeChild' on 'Node': parameter 1 is not of type 'Node'.
 at Object.removeChild (htmldomapi.js:18)
 at removeVnodes (snabbdom.js:153) 
 at updateChildren (snabbdom.js:233) 
 at patchVnode (snabbdom.js:256) 
 at patch (snabbdom.js:28)
```

This is the cause for https://github.com/cyclejs/cyclejs/issues/531

Creating the vnode with (undefined, undefined, undefined, text, node) ensures that snabbdom is properly able to remove the nodes.

staltz:
Fix will be released in v0.6.5

PR #253, opened 2017-02-25
fix tovnode to ensure text elements can be removed

Steelfish:
I've added the change to fix https://github.com/snabbdom/snabbdom/issues/252 and added a test that threw the type error before the change and now passes.

This should also fix https://github.com/cyclejs/cyclejs/issues/531

### DIFF
--- a/src/tovnode.ts
+++ b/src/tovnode.ts
@@ -27,7 +27,7 @@ export function toVNode(node: Node, domApi?: DOMAPI): VNode {
     return vnode(sel, {attrs}, children, undefined, node);
   } else if (api.isText(node)) {
     text = api.getTextContent(node) as string;
-    return vnode(undefined, undefined, undefined, text, undefined);
+    return vnode(undefined, undefined, undefined, text, node);
   } else if (api.isComment(node)) {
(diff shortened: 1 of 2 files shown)


staltz:
Superb PR. Thanks

(For others: I'll merge this since I tested that it does what it should do, and the bug is very important to fix, while the fix itself is tiny)

nunocastromartins:
I just checked and comment nodes have the same problem, would you please fix that as well? :-)

src/tovnode.ts
```diff
@@ -27,7 +27,7 @@ export function toVNode(node: Node, domApi?: DOMAPI): VNode {
     return vnode(sel, {attrs}, children, undefined, node);
   } else if (api.isText(node)) {
     text = api.getTextContent(node) as string;
-    return vnode(undefined, undefined, undefined, text, undefined);
+    return vnode(undefined, undefined, undefined, text, node);
   } else if (api.isComment(node)) {
     text = api.getTextContent(node) as string;
     return vnode('!', undefined, undefined, text, undefined);
```
nunocastromartins: I believe this should have `node` as well.

Issue #254, opened 2017-03-04
Boolean attributes should not have a value


PerWiklander:
The current boolean attribute handling feels only half done. The attributes are added and removed in a correct way (depending on if the value is truthy or not), but they shouldn't have values set.

I suggest that values set on [boolean attributes](https://github.com/snabbdom/snabbdom/blob/master/src/modules/attributes.ts#L8) should only be used to determine if the attribute is rendered or not and not added as value of the attribute.

According to http://w3c.github.io/html/infrastructure.html#sec-boolean-attributes

"A number of attributes are boolean attributes. The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

If the attribute is present, its value must either be the empty string or a value that is an ASCII case-insensitive match for the attribute’s canonical name, with no leading or trailing white space.

NOTE
The values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether."

mightyiam:
Sounds correct. Similar to the class module, isn't it?

mightyiam:
But the API isn't clear to me. Example?

PerWiklander:
The suggested new API would given `hidden = true` output only `hidden`. Given `hidden = false` it would remove a previous attribute with the name `hidden` (this works today).

Today, from what I can see, a value of `hidden = "whatever"` would output `hidden = "whatever"` in the dom.

caridy:
I favor this request as well. Setting the value to empty string for boolean attributes should be enough, we can effectively ignore the provided value. More details here: https://html.spec.whatwg.org/multipage/infrastructure.html#boolean-attributes

PerWiklander:
Yeah, that is the same spec as I referred to above.

paldepind:
Thank you for reporting this issue @PerWiklander :smile: Please take a look at #255. It should fix the problem. If it looks good to you I will merge and do a patch release.

PerWiklander:
The implementation looks good to me, except the fact that having a `const booleanAttrs` is not very future proof.

PerWiklander:
It looks like it is missing `buffered` and `contenteditable` already.
Wouldn't it work if it checks if the attribute value is strictly `true` or `false`? Attributes that might look like they are boolean, like `aria-atomic` should actually have the `string` value `"true"` or `"false"`, so they would (correctly) not be affected.

paldepind:
[...]

PR #257, opened 2017-03-10
Performance improvement for attribute patching:

diervo:
- Avoid call to split and use a object lookup instead
- Simplify the code for svg namespaced attributes

### DIFF
--- a/src/modules/attributes.ts
+++ b/src/modules/attributes.ts
@@ -10,20 +10,7 @@ const booleanAttrs = ["allowfullscreen", "async", "autofocus", "autoplay", "chec
 
 const xlinkNS = 'http://www.w3.org/1999/xlink';
 const xmlNS = 'http://www.w3.org/XML/1998/namespace';
-
-const namespaces = Object.create(null);
-
-namespaces['xlink:href'] = xlinkNS;
-namespaces['xlink:arcrole'] = xlinkNS;
-namespaces['xlink:actuate'] = xlinkNS;
-namespaces['xlink:role'] = xlinkNS;
-namespaces['xlink:title'] = xlinkNS;
-namespaces['xlink:show'] = xlinkNS;
-namespaces['xlink:type'] = xlinkNS;
-namespaces['xml:base'] = xmlNS;
-namespaces['xml:lang'] = xmlNS;
-namespaces['xml:space'] = xmlNS;
-
+const colonChar = 58;
 const booleanAttrsDict: {[attribute: string]: boolean} = Object.create(null);
 
 for (let i = 0, len = booleanAttrs.length; i < len; i++) {
@@ -52,9 +39,12 @@ function updateAttrs(oldVnode: VNode, vnode: VNode): void {
           elm.removeAttribute(key);
         }
       } else {
-        const ns = namespaces[key];
-        if (ns) {
-          elm.setAttributeNS(ns, key, cur);
+        if (key.charCodeAt(3) === colonChar) {
+          // Assume xml namespace
+          elm.setAttributeNS(xmlNS, key, cur);
+        } else if (key.charCodeAt(5) === colonChar) {
+          // Assume xlink namespace
+          elm.setAttributeNS(xlinkNS, key, cur);
         } else {
           elm.setAttribute(key, cur);
         }
--- a/src/modules/attributes.ts
+++ b/src/modules/attributes.ts
@@ -11,6 +11,7 @@ const booleanAttrs = ["allowfullscreen", "async", "autofocus", "autoplay", "chec
 const xlinkNS = 'http://www.w3.org/1999/xlink';
 const xmlNS = 'http://www.w3.org/XML/1998/namespace';
 const colonChar = 58;
+const xChar = 120;
 const booleanAttrsDict: {[attribute: string]: boolean} = Object.create(null);
 
 for (let i = 0, len = booleanAttrs.length; i < len; i++) {
@@ -39,7 +40,9 @@ function updateAttrs(oldVnode: VNode, vnode: VNode): void {
           elm.removeAttribute(key);
         }
       } else {
-        if (key.charCodeAt(3) === colonChar) {
+        if (key.charCodeAt(0) !== xChar) {
+          elm.setAttribute(key, cur);
+        } else if (key.charCodeAt(3) === colonChar) {
           // Assume xml namespace
           elm.setAttributeNS(xmlNS, key, cur);
         } else if (key.charCodeAt(5) === colonChar) {
--- a/src/modules/attributes.ts
+++ b/src/modules/attributes.ts
@@ -18,10 +18,10 @@ namespaces['xlink:arcrole'] = xlinkNS;
 namespaces['xlink:actuate'] = xlinkNS;
 namespaces['xlink:role'] = xlinkNS;
 namespaces['xlink:title'] = xlinkNS;
+namespaces['xlink:show'] = xlinkNS;
 namespaces['xlink:type'] = xlinkNS;
 namespaces['xml:base'] = xmlNS;
 namespaces['xml:lang'] = xmlNS;
-namespaces['xml:show'] = xmlNS;
 namespaces['xml:space'] = xmlNS;
 
 const booleanAttrsDict: {[attribute: string]: boolean} = Object.create(null);
--- a/src/modules/attributes.ts
+++ b/src/modules/attributes.ts
@@ -17,10 +17,11 @@ namespaces['xlink:href'] = xlinkNS;
 namespaces['xlink:arcrole'] = xlinkNS;
 namespaces['xlink:actuate'] = xlinkNS;
 namespaces['xlink:role'] = xlinkNS;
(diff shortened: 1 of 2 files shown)


src/modules/attributes.ts
```diff
@@ -40,9 +51,9 @@ function updateAttrs(oldVnode: VNode, vnode: VNode): void {
           elm.removeAttribute(key);
         }
       } else {
-        namespaceSplit = key.split(":");
-        if (namespaceSplit.length > 1 && NamespaceURIs.hasOwnProperty(namespaceSplit[0])) {
-          elm.setAttributeNS((NamespaceURIs as any)[namespaceSplit[0]], key, cur);
+        const ns = namespaces[key];
```
caridy: good call, we were doing all these string computations for all attributes :/

src/modules/attributes.ts
```diff
@@ -1,17 +1,28 @@
...
                 "muted", "nohref", "noresize", "noshade", "novalidate", "nowrap", "open", "pauseonexit", "readonly",
                 "required", "reversed", "scoped", "seamless", "selected", "sortable", "spellcheck", "translate",
                 "truespeed", "typemustmatch", "visible"];
 
+const xlinkNS = 'http://www.w3.org/1999/xlink';
+const xmlNS = 'http://www.w3.org/XML/1998/namespace';
+
+const namespaces = Object.create(null);
+
+namespaces['xlink:href'] = xlinkNS;
```
caridy: @diervo we are missing `xlink:show`. this is the major difference. This PR makes the list very restrictive, which I'm ok with, but we will have to maintain it if more attributes are added at some point.

src/modules/attributes.ts
```diff
@@ -1,17 +1,28 @@
...
+namespaces['xlink:arcrole'] = xlinkNS;
+namespaces['xlink:actuate'] = xlinkNS;
+namespaces['xlink:role'] = xlinkNS;
+namespaces['xlink:titlef'] = xlinkNS;
```
caridy: typo, should be `xlink:title`

caridy:
For reference, here is the official list of svg attributes:

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute

caridy:
@diervo thanks for making the changes.

I'm supportive of this change, even though it is more restrictive than the current code, but it should improve perf considerable.

paldepind:
Great PR. Doing a `split` for every attribute is definitely very bad.

However, the approach in this PR got me thinking.

The most important thing is to make the common case, when there is no namespace, as fast as possible.

So what about doing something like the following:

```js
if (key[0] !== "x") { // fast path
  elm.setAttribute(key, cur);
} else {
  if (key[4] === ":") {
    // Assume xml namespace
    elm.setAttributeNS(xmlNS, key, cur);
  } else if (key[6] === ":") {
    // Assume xlink namespace
    elm.setAttributeNS(xlinkNS, key, cur);
  }
}
```

Benefits

* When there is no prefix the check `key[0] !== "x"` is _very_ cheap
* Simply looking at index 4 and 6 for a `:` should be faster than doing a lookup in the `namespaces` map.
* No need to maintain `namespaces`

Cons

* Assumes that if `:` is found it implies the specific namespace. However, if that is an issue more checking could easily be done after the `x` and the `:` is found.

diervo:
@paldepind couple of things: 

String comparison is around an order of magnitude slower than a map lookup, in fact we should avoid string comparison in favor of integer/boolean comparison as much as we can everywhere in the library). Your code will still run much slower at runtime.

Moreover I think the code you proposed IMO is very error prone and hard to read and maintain going forward.

I really think for this case explicitly adding the properties outweighs everything else from a performance, maintainability, etc. The only downside is that you added some extra bytes on the payload.

I have checked in all other vDOM implementations and most of them choose speed and maintainability over everything else, so I really think the trade-offs here are very worth it given that is one of the critical performance hotspots.

diervo:
@caridy @Pmdartus Let's provide some perf numbers if we can soon for this.

caridy:
Here is the benchmark:
https://jsperf.com/snabbdom-attribute-type

I did a little bit more optimizations on @paldepind's code, to use numeric comparison, which is faster, e.g.:

```js
  if (key.charCodeAt(3) === 58) {
    // Assume xml namespace
    elm.setAttributeNS(xmlNS, key, cur);
  } else if (key.charCodeAt(5) === 58) {
    // Assume xlink namespace
    elm.setAttributeNS(xlinkNS, key, cur);
  } else {
    elm.setAttribute(key, cur);
  }
```

Overall, both solutions are statistically equal, so either way is fine, but the `charCodeAt()` solution is simpler, smaller, and more flexible since we don't have to keep the map for new SVG attributes.

Also, it align with the other idea that we have been debating, snabbdom's role is not to validate, it is to act fast with the smaller footprint, and validation can happen at the higher level (static errors when compiling JSX or HTML) on the compiler of templates, etc. So, let's go with the `charAtCode()` solution. @diervo can you update the PR?

diervo:
@caridy @paldepind let me know if you guys are good with the changes.

paldepind:
> String comparison is around an order of magnitude slower than a map lookup, in fact we should avoid string comparison in favor of integer/boolean comparison as much as we can everywhere in the library).

I think you have a good point that string comparison can be slow. But object lookups aren't particularly fast either.

> Your code will still run much slower at runtime.

[This JSperf](https://jsperf.com/snabbdom-attributes-namespace) shows that my code runs significantly faster than the object lookup based approach. In Chrome my code is about 35% faster and in Firefox the gap is even bigger.

The key observation is this: Most attributes that the code will encounter in the real world will not be namespaced attributes. The vast majority will be normal attributes.

So the key thing is to optimize the handling of non-namespaced attributes. If we can make a fast-path for those we will get better performance. And `key[0] !== "x"` is faster than `namespaces[attrName] === undefined` at quickly figuring out that something is not a namespace attribute.

In the jsperf @caridy posted 50% of the attributes tested are namespaced attributes. This does not accurately represent the real world. Thus it gives an unrealistic disadvantage to my code. In the jsperf I posted only 10% of the attributes are namespaced. In the real world the number is probably even smaller.

> Moreover I think the code you proposed IMO is very error prone and hard to read and maintain going forward.

I definitely agree that it isn't the most obvious thing in the world. But I think some helpful comments will solve that.

@caridy Good idea to use `charCodeAt`. It does indeed provide a small but nice performance improvement.

I think the PR should be updated to use `attrName.charCodeAt(0) !== 120` as in the jsperf I linked.

caridy:
@diervo the code is good, can you just double check that we have tests for this?

@paldepind never got the link from your perf, but I totally agree with your assessment, attributes with charCode 58 are rare, or should be. I think we are all in agreement, let's move forward.

paldepind:
> never got the link from your perf, but I totally agree with your assessment, attributes with charCode 58 are rare, or should be.

What do you mean? The link is in the comment above.

The PR currently does not include the important performance improvement of checking `attrName.charCodeAt(0) !== 120` first.

diervo:
Will do the change, add some tests if appropriate and merge it.

diervo:
@paldepind @caridy Let me know if this is ready for merge!

caridy:
I think this is good enough. There is another refactor around booleans (https://github.com/snabbdom/snabbdom/pull/230 and https://github.com/snabbdom/snabbdom/issues/254) that might touch this logic, and we can generalize it a little bit more.

Just double check for the tests, if we have good coverage of this, we can go ahead and merge this, if not, add a couple of tests just to be sure.

diervo:
Will do, I think once we settle on the boolean, we can come back to this code.

diervo:
@paldepind @caridy Added a couple of tests. Please do the honors of squash and merge :)

nunocastromartins:
Before merging this, consider a completely different approach: remove this code altogether; create a new module to handle namespaced attributes.
Something that'd make you use, say:
```js
h('div', {nsAttrs: {xlink: ['http://www.w3.org/1999/xlink', {href: 'something'}]}}, [])
```
You could perhaps allow something like that for generic namespaces but recognize `xml`/`xlink` ones, so you'd write (instead of the above):
```js
h('div', {nsAttrs: {xlink: {href: 'something'}}}, [])
```
To me this just seems much simpler: I think Snabbdom should strive for more generic solutions.

diervo:
@nunocastromartins I don't disagree with your sentiment about simplicity. However your proposed API will be a breaking change, where this PR is not, since this is just a performance optimization rather than a functional change.

So, I'm going to merge this PR since will give immediate perf benefits to all users, and you can create a separate issue/PR to introduce the new module. We could introduce a new module as part of a mayor version update, if we all feel that approach is superior.

That being said, I would love to see the performance difference between both approaches as part of any PR that adds code.  We are working on having some perf automation that we can run to see the differences per commit.

paldepind:
@nunocastromartins I think that is a very nice idea :+1: 

@diervo 

> We are working on having some perf automation that we can run to see the differences per commit.

That sounds pretty awesome.

nunocastromartins:
I don't mind doing it. It'll have to wait until next week, though! 😄 
Feel free to beat me to it!

PR #260, opened 2017-03-14
use regex to parse selector

ricokahler:
I think this pull is pretty self explanatory. The only issue you might with this is that if the parsed selector couldn't find a tag, then it [defaults to `div`](https://github.com/ricokahler/snabbdom/blob/robust-selector/src/snabbdom.ts#L112). Maybe you'd like to `throw` instead?

### DIFF
--- a/src/snabbdom.ts
+++ b/src/snabbdom.ts
@@ -45,6 +45,25 @@ const hooks: (keyof Module)[] = ['create', 'update', 'remove', 'destroy', 'pre',
 export {h} from './h';
 export {thunk} from './thunk';
 
+/**
+ * Robust selector parser.
+ * Parses a selector into a `tag`, `id`, and `className`.
+ * The `className` is a single string with all the classes of the selector seperated by spaces.
+ */
+export function parseSelector(selector: string) {
+  const tagMatch = selector.match(/^[^#.]+?(?=[#.]|$)/);
+  const tag = (tagMatch && tagMatch[0] && tagMatch[0].trim()) || undefined;
+  const idMatch = selector.match(/#([^.#]+?(?=\.|$))/);
+  const id = (idMatch && idMatch[1]) || undefined; // prefer to return `undefined`
+  const classNameMatch = selector.match(/\.([^.#]+?(?=[.#]|$))/g);
+  const className = (Array
+    .from(classNameMatch || [])
+    .map(match => match.substring(1).trim())
+    .join(' ')
+  ) || undefined;
+  return { tag, id, className };
+}
+
 export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
   let i: number, j: number, cbs = ({} as ModuleHooks);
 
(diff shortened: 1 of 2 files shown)


paldepind:
Thank you for the PR @ricokahler.

This PR has a lot in common with #118.

I must admint that I am personally quite happy with the current selector parsing.

* The code is simple and easy to understand
* I guessing it performs better than the one in this PR. For instance it avoids creating arrays. This parser allocates at least two arrays. One in `Array.from` and another one when calling `map`.
* It enforces that the id must come first in the string which I think is good for consistency.

But, that is just my opinion. As #118 shows you're not the first who has wanted more features and flexibility from the parser.

One could take a parser even further and support CSS-selector like stuff like this:

```js
h("a.btn[href=foo]", "Click me");
```

Maybe a fancier `h`-function should be created that has features like the one in this PR?

ricokahler:
I actually like the current selector too. I also had the task of parsing a selector so the first thing I did was copy snabbdom's :smile: .  Later down the road though, I wanted the selector to be more robust so I ended up using regex to create my own and the result was a pretty robust parser. So I thought why not do a pull here?

Anyway, if you're really considering performance, the regex parsing might also have an effect but I don't think it'll make a difference.

Maybe the biggest difference between this and #118 is that I add tests :D ??

nunocastromartins:
@paldepind I agree with you and, to be quite fair, I'd rather see this going the opposite direction: remove all of this logic from the core algorithm and accept only a tag in the `sel`; all of the "fancy" stuff could go into the `h` function.

However, this would make the `sameVNode` function only compare tags rather than the composition of "tag, id, classes": meaning that `sameVNode` would have to access `data.id` and `data.class(Name?)` instead of only the `sel` in order to keep the same behaviour.

Just my two cents. :-)

staltz:
Agreed with paldepind too. Things we need to keep an eye on are performance and size. Ideally we should have in snabbdom automatic performance change reports and size change reports, so it's easy to know "this PR made the code 1kB smaller and 10% slower". So far we need to do that assessment manually. @ricokahler can you say how the code size changed in total?

caridy:
I will say the same I have said in other issues, the role of snabbdom is not to normalize, that can happen at the higher level (compilers, app, etc.), the goal here is to have a super-fast diffing algo with the minimum complexity and memory footprint, that aligns with the @nunocastromartins' comment. It is true that for this particular case, `sameVNode` is a problem, maybe we can open that up so a custom function can be provided when constructing the patch, so people can create their own `sameVNode()` with whatever logic they want.

nunocastromartins:
@caridy I had thought of that exactly but didn't want to propose too many changes in a single comment - especially (possibly) breaking ones. :-)

ricokahler:
revisiting this, i would have to agree with all of you, the current selector is faster is enforces better structure.

maybe this selector could be used for development only? like react, we could log warnings when we see a selector being formatted incorrectly but still render it. then in a production build we could [uglify the dev selector out](https://github.com/mishoo/UglifyJS#use-as-a-code-pre-processor) and replace it with a production one.

there's a lot of opportunity with doing something like that but i feel like we need more use cases than incorrectly formatted selectors to bring in that extra level of build complexity. i'd be happy to contribute for something like that but my bandwidth is limited at the time being so this selector will be closed for now.

anyway thanks for the great vdom library!

PR #266, opened 2017-03-24
Fix toVNode() for comment nodes, lacking some fields

staltz:
Comment nodes when converted to VNode were lacking `vnode.data` and `vnode.elm`, which in turn would break use cases with modules such as the style module (Which assumed `vnode.data` to be truthy). Since `toVNode` is used as an initial step (e.g. for server-side rendered content), there is no reason to not provide `vnode.elm` since that vnode may be used as the "oldvnode" during patch. Also, because comment vnodes have the `'!'` sel, most logic in snabbdom that detects a truthy `sel` will also assume a truthy `data` field. It is easier to build the comment vnode with those fields than to modify all the logic elsewhere in snabbdom. Probably also better for performance.

This fix in snabbdom is important for this issue in Cycle.js: https://github.com/cyclejs/cyclejs/issues/561

### DIFF
--- a/src/tovnode.ts
+++ b/src/tovnode.ts
@@ -30,7 +30,7 @@ export function toVNode(node: Node, domApi?: DOMAPI): VNode {
     return vnode(undefined, undefined, undefined, text, node);
   } else if (api.isComment(node)) {
     text = api.getTextContent(node) as string;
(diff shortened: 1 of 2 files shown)


staltz:
This is an old and useful bug fix PR with apparently no downsides, so I'll just merge.

PR #270, opened 2017-03-28
dont lose listener

Reon90:
We have vnode with listener, after that we update vnode, then vnode lose listener.
If vnode lose listener, it will not be able to pass the checking:
`if (oldOn && oldListener) {`

### DIFF
--- a/src/modules/eventlisteners.ts
+++ b/src/modules/eventlisteners.ts
@@ -52,6 +52,9 @@ function updateEventListeners(oldVnode: VNode, vnode?: VNode): void {
 
   // optimization for reused immutable handlers
   if (oldOn === on) {
+    if (vnode) {
+      vnode.listener = oldVnode.listener;
+    }
     return;
   }
 
--- a/src/modules/eventlisteners.ts
+++ b/src/modules/eventlisteners.ts
@@ -53,7 +53,7 @@ function updateEventListeners(oldVnode: VNode, vnode?: VNode): void {
   // optimization for reused immutable handlers
   if (oldOn === on) {
     if (vnode) {
-      vnode.listener = oldVnode.listener;
+      (vnode as any).listener = (oldVnode as any).listener;
     }
     return;
   }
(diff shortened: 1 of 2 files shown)


Reon90:
It causes serious issues. Why do you want to fix it?

caridy:
@Reon90 we do want to fix any issue. We just need to be sure. Can you provide an example, hopefully in a form of a test that fails for this case? At first glance, this seems very weird because restoring `vnode.listener` does not guarantee that the element has the proper event listeners added, which will be a major problem.

Reon90:
@caridy, test was added.

Reon90:
@caridy could you check it?

test/eventlisteners.js
```diff
@@ -169,4 +169,18 @@ describe('event listeners', function() {
     elm.firstChild.click();
     assert.equal(3, result.length);
   });
+  it('Rewrite listener on the same node', function() {
+    var result = [];
+    function onClick(ev) {  };
+    function onClick2(ev) { result.push(ev) };
+    var obj = {on: {click: onClick }};
```
caridy: this is the main issue here IMO, you're reusing `data` instead of reusing `on`. I don't think you want `data` to be memoizable because you loose the opportunity to do other optimizations. What if you do: `var on = { click: onClick }` and then everytime you construct `h` you pass `{on}` as data?

Also, I don't see the `result.push(ev)` in `onClick`, should we add it to make sure that we are not adding the same listeners over and over again?

src/modules/eventlisteners.ts
```diff
@@ -52,6 +52,9 @@ function updateEventListeners(oldVnode: VNode, vnode?: VNode): void {
 
   // optimization for reused immutable handlers
   if (oldOn === on) {
+    if (vnode) {
+      (vnode as any).listener = (oldVnode as any).listener;
```
caridy: I think you're right, and there is a problem here, but this patch is not sufficient. Reusing listeners is tricky, because the listener has a backpointer to the corresponding vnode, and you will still be loosing that, which is covered by `listener.vnode = vnode;` (line 84).

mightyiam:
I will close this PR for clean-up purposes. I did not find an issue. If anyone feels there's an issue, could you please report an issue?

PR #271, opened 2017-04-05
type VNode props (fixes #264)

### LINKED ISSUE #264: Various props of VNode are typed `any`
https://github.com/snabbdom/snabbdom/blob/v0.6.7/src/vnode.ts#L16-L23

Would it be beneficial to type those?

mightyiam:
Is this any kind of reasonable?

### DIFF
--- a/src/modules/attributes.ts
+++ b/src/modules/attributes.ts
@@ -1,6 +1,8 @@
 import {VNode, VNodeData} from '../vnode';
 import {Module} from './module';
 
+export type Attrs = Record<string, string | number | boolean>
+
 const booleanAttrs = ["allowfullscreen", "async", "autofocus", "autoplay", "checked", "compact", "controls", "declare",
                 "default", "defaultchecked", "defaultmuted", "defaultselected", "defer", "disabled", "draggable",
                 "enabled", "formnovalidate", "hidden", "indeterminate", "inert", "ismap", "itemscope", "loop", "multiple",
@@ -40,16 +42,19 @@ function updateAttrs(oldVnode: VNode, vnode: VNode): void {
           elm.removeAttribute(key);
         }
       } else {
+        // because those in TypeScript are too restrictive: https://github.com/Microsoft/TSJS-lib-generator/pull/237
+        type SetAttribute = (name: string, value: string | number | boolean) => void;
+        type SetAttributeNS = (namespaceURI: string, qualifiedName: string, value: string | number | boolean) => void;
         if (key.charCodeAt(0) !== xChar) {
-          elm.setAttribute(key, cur);
+          (elm.setAttribute as SetAttribute)(key, cur);
         } else if (key.charCodeAt(3) === colonChar) {
           // Assume xml namespace
-          elm.setAttributeNS(xmlNS, key, cur);
+          (elm.setAttributeNS as SetAttributeNS)(xmlNS, key, cur);
         } else if (key.charCodeAt(5) === colonChar) {
           // Assume xlink namespace
-          elm.setAttributeNS(xlinkNS, key, cur);
+          (elm.setAttributeNS as SetAttributeNS)(xlinkNS, key, cur);
         } else {
-          elm.setAttribute(key, cur);
+          (elm.setAttribute as SetAttribute)(key, cur);
         }
       }
     }
--- a/src/modules/eventlisteners.ts
+++ b/src/modules/eventlisteners.ts
@@ -1,6 +1,12 @@
 import {VNode, VNodeData} from '../vnode';
 import {Module} from './module';
 
+export type On = {
+  [N in keyof HTMLElementEventMap]: (ev: HTMLElementEventMap[N]) => void
+} & {
+  [event: string]: EventListener
+};
+
 function invokeHandler(handler: any, vnode?: VNode, event?: Event): void {
   if (typeof handler === "function") {
     // call function handler
--- a/src/modules/style.ts
+++ b/src/modules/style.ts
@@ -1,10 +1,9 @@
 import {VNode, VNodeData} from '../vnode';
 import {Module} from './module';
 
-export interface VNodeStyle {
-  delayed: { [prop: string]: string }
-  remove: { [prop: string]: string }
-  [prop: string]: string | { [prop: string]: string }
+export type VNodeStyle = Record<string, string> & {
+  delayed: Record<string, string>
+  remove: Record<string, string>
 }
 
 var raf = (typeof window !== 'undefined' && window.requestAnimationFrame) || setTimeout;
--- a/src/modules/style.ts
+++ b/src/modules/style.ts
@@ -1,6 +1,12 @@
 import {VNode, VNodeData} from '../vnode';
 import {Module} from './module';
 
+export interface VNodeStyle {
+  delayed: { [prop: string]: string }
+  remove: { [prop: string]: string }
+  [prop: string]: string | { [prop: string]: string }
+}
+
 var raf = (typeof window !== 'undefined' && window.requestAnimationFrame) || setTimeout;
 var nextFrame = function(fn: any) { raf(function() { raf(fn); }); };
 
@@ -15,8 +21,8 @@ function updateStyle(oldVnode: VNode, vnode: VNode): void {
 
   if (!oldStyle && !style) return;
   if (oldStyle === style) return;
-  oldStyle = oldStyle || {};
-  style = style || {};
+  oldStyle = oldStyle || {} as VNodeStyle;
+  style = style || {} as VNodeStyle;
   var oldHasDel = 'delayed' in oldStyle;
 
   for (name in oldStyle) {
--- a/src/vnode.ts
+++ b/src/vnode.ts
@@ -1,4 +1,6 @@
 import {Hooks} from './hooks';
+import {AttachData} from './helpers/attachto'
+import {VNodeStyle} from './modules/style'
 
 export type Key = string | number;
 
@@ -12,22 +14,32 @@ export interface VNode {
 }
 
 export interface VNodeData {
-  // modules - use any because Object type is useless
-  props?: any;
-  attrs?: any;
-  class?: any;
-  style?: any;
-  dataset?: any;
-  on?: any;
-  hero?: any;
-  attachData?: any;
+  props?: {
+    [prop: string]: any
+  };
+  attrs?: {
+    [attr: string]: any
+  };
+  class?: {
+    [name: string]: boolean
+  };
+  style?: VNodeStyle;
+  dataset?: {
+    [name: string]: string
+  };
+  on?: {
+    [event: string]: EventListener | EventListener[]
+  };
+  hero?: {
+    id: string
+  };
+  attachData?: AttachData;
   hook?: Hooks;
   key?: Key;
   ns?: string; // for SVGs
   fn?: () => VNode; // for thunks
   args?: Array<any>; // for thunks
   [key: string]: any; // for any other 3rd party module
-  // end of modules
 }
 
 export function vnode(sel: string | undefined,
--- a/src/vnode.ts
+++ b/src/vnode.ts
@@ -1,6 +1,12 @@
 import {Hooks} from './hooks';
 import {AttachData} from './helpers/attachto'
 import {VNodeStyle} from './modules/style'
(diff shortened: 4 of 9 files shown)


mightyiam:
@paldepind I'll look into the magic stuff that I'm not familiar with that you described in [this comment](https://github.com/snabbdom/snabbdom/issues/264#issuecomment-288535381) regarding the types of `style` and `on`. In the mean time, I'll be happy for other input.

src/vnode.ts
```diff
@@ -12,22 +14,32 @@ export interface VNode {
...
-  // modules - use any because Object type is useless
-  props?: any;
-  attrs?: any;
-  class?: any;
-  style?: any;
-  dataset?: any;
-  on?: any;
-  hero?: any;
-  attachData?: any;
+  props?: {
```
caridy: probably generalizing this with a new type, something like:

```js
declare interface HashTable<T> {
    [key: string]: T;
}
```

then just:

```js
props?: HashTable<any>;
on?: HashTable<EventListener|EventListener[]>
```

TylorS:
`props` in particular could be improved quite a bit further. Here is an interface containing all Html properties I could find on MDN.

```typescript
export interface HtmlProperties {
  // Node properties
  nodeValue?: string | null;
  textContent?: string | null;

  // Element properties
  className?: string;
  id?: string;
  innerHTML?: string;
  scrollLeft?: number;
  scrollTop?: number;
  slot?: string;

  // HTML Element specific
  abbr?: string;
  axis?: string;
  accessKey?: string;
  accept?: string;
  acceptCharset?: string;
  action?: string;
  align?: string;
  aLink?: any;
  allowFullscreen?: boolean;
  alt?: string;
  altHtml?: string;
  archive?: string;
  async?: boolean;
  autocomplete?: string;
  autofocus?: boolean;
  autoplay?: boolean;
  background?: string;
  behavior?: string;
  bgColor?: any;
  bgProperties?: string;
  border?: string;
  borderColor?: any;
  caption?: HTMLTableCaptionElement;
  cellPadding?: string;
  cells?: HTMLCollectionOf<HTMLTableDataCellElement | HTMLTableHeaderCellElement>;
  cellSpacing?: string;
  charset?: string;
  checked?: boolean;
  cite?: string;
  clear?: string;
  code?: string;
  codeBase?: string;
  codeType?: string;
  cols?: string;
  colSpan?: number;
  compact?: boolean;
  content?: string;
  coords?: string;
  contentEditable?: string;
  controls?: boolean;
  crossOrigin?: string;
  currentTime?: number;
  data?: string;
  dateTime?: string;
  declare?: boolean;
  default?: boolean;
  defaultChecked?: boolean;
  defaultMuted?: boolean;
  defaultPlaybackRate?: number;
  defaultSelected?: boolean;
  defaultValue?: string;
  defer?: boolean;
  dir?: string;
  direction?: string;
  disabled?: boolean;
  download?: string;
  draggable?: boolean;
  encoding?: string;
  enctype?: string;
  event?: string;
  face?: string;
  formAction?: string;
  formEnctype?: string;
  formMethod?: string;
  formNoValidate?: string;
  formTarget?: string;
  frame?: string;
  frameBorder?: string;
  frameSpacing?: any;
  hash?: string;
  headers?: string;
  height?: any;
  hidden?: boolean;
  hideFocus?: boolean;
  high?: number;
  host?: string;
  hostname?: string;
  href?: string;
  hreflang?: string;
  hspace?: number;
  htmlFor?: string;
  httpEquiv?: string;
  indeterminate?: boolean;
  innerText?: string;
  isMap?: boolean;
  kind?: string;
  label?: string;
  lang?: string;
  link?: any;
  longDesc?: string;
  loop?: number;
  low?: number;
  lowsrc?: string;
  marginHeight?: string;
  marginWidth?: string;
  max?: string;
  maxLength?: number;
  method?: string;
  media?: string;
  min?: string;
  minLength?: number;
  multiple?: boolean;
  msAudioCategory?: string;
  msAudioDeviceType?: string;
  msHorizontalMirror?: boolean;
  msKeySystem?: string;
  msPlayToDisabled?: boolean;
  msPlayToPreferredSourceUri?: string;
  msPlayToPrimary?: boolean;
  msRealTime?: boolean;
  msStereo3DPackingMode?: string;
  msStereo3DRenderMode?: string;
  msZoom?: boolean;
  muted?: boolean;
  name?: string;
  noHref?: boolean;
  noResize?: boolean;
  noShade?: boolean;
  noWrap?: boolean;
  noValidate?: boolean;
  object?: string | null;
  optimum?: number;
  outerHTML?: string;
  outerText?: string;
  pathname?: string;
  pattern?: string;
  placeholder?: string;
  playbackRate?: number;
  port?: string;
  preload?: string;
  protocol?: string;
  profile?: string;
  readOnly?: boolean;
  readyState?: number;
  rel?: string;
  required?: boolean;
  rev?: string;
  rows?: string | HTMLCollectionOf<HTMLTableRowElement>;
  rowSpan?: number;
  rules?: string;
  scope?: string;
  scrolling?: string;
  scheme?: string;
  search?: string;
  selected?: boolean;
  selectedIndex?: number;
  selectionDirection?: string;
  selectionEnd?: number;
  selectionStart?: number;
  shape?: string;
  size?: number;
  sizes?: string;
  scrollAmount?: number;
  scrollDelay?: number;
  src?: string;
  srclang?: string;
  srcObject?: MediaStream | null;
  srcset?: string;
  spellcheck?: boolean;
  standby?: string;
  start?: number;
  status?: any;
  step?: string;
  summary?: string;
  target?: string;
  tabIndex?: number;
  tBodies?: HTMLCollectionOf<HTMLTableSectionElement>;
  text?: string;
  tFoot?: HTMLTableSectionElement;
  tHead?: HTMLTableSectionElement;
  title?: string;
  trueSpeed?: boolean;
  type?: string;
  units?: string;
  url?: string;
  urn?: string;
  useMap?: string;
  vspace?: number;
  valueAsDate?: Date;
  valueAsNumber?: number;
  value?: string;
  version?: string;
  volume?: number;
  vAlign?: string;
  vLink?: any;
  webkitdirectory?: boolean;
  width?: number;
  wrap?: string;
}
```

mightyiam:
@paldepind regarding [your suggestion](https://github.com/snabbdom/snabbdom/issues/264#issuecomment-288535381) of using `HTMLElementEventMap` in the type of the `on` property; wouldn't this make attempting to listen to custom events an error?

mightyiam:
Similar question regarding using `CSSStyleDeclaration` and what it would mean for some out-of-spec style properties. If we provide index signature, and all the values are strings, then it might as well be only an index signature. There is no value in using `CSSStyleDecleration` in this case, right?

src/vnode.ts
mightyiam: That might make it a bit less readable for users, in their IDE/tooling whatever, I imagine. But not tested this.

mightyiam:
If this is an improvement with no adverse side effects, then perhaps it could be already merged.

mightyiam:
No time for little ol' snabbdom?

paldepind:
Sorry 😿 I'll take a look at it tomorrow or the day after.

src/vnode.ts
paldepind: I think the current version is fine.

In any case defining a custom `HashTable` is not necessary as TypeScript ships with the following type out of the box:

```ts
type Record<K extends string, T> = {
    [P in K]: T;
}
```

Using this type `props` could be defined to have the type `Record<string, any>` which would be equivalent to the current type.

src/vnode.ts
```diff
@@ -12,22 +14,32 @@ export interface VNode {
...
+    [prop: string]: any
+  };
+  attrs?: {
+    [attr: string]: any
```
paldepind: I think we can restrict this one to `string | number | boolean`. Attributes are always strings, numbers coerce to strings, and booleans are for boolean attributes.

paldepind:
> regarding [your suggestion](https://github.com/snabbdom/snabbdom/issues/264#issuecomment-288535381) of using `HTMLElementEventMap` in the type of the `on` property; wouldn't this make attempting to listen to custom events an error?

Yes, you are right. If the `on` property is given the `On` type I wrote then custom events will not be allowed. To fix that we can tweak the type slightly to allow any event name while still getting most of the benefits.

```ts
type On = {[N in keyof HTMLElementEventMap]: (ev: HTMLElementEventMap[N]) => void};

export interface VNodeData {
  ...
  on: On | {[event: string]: EventListener}
  ...
}
```

In the union type for `on` TypeScript will first try and match the first type and afterward the "catch-all' type. This means that the user can give an object with any key. But, if the user gives an object with known event names we can use that to calculate the type that the event type the event handler will receive.

Here is a small self-contained example showing what I mean.

image

In the first call to `test` the property in the object is unknown so the type of `ev` is `EventListener`. In the second invocation `click` is a known event so the argument to the event handler is correctly found to be of type `MouseEvent`. This provides extra help and safety inside the event handler.

So the type I gave above for `on` will be just as permissive but it will provide benefits as it gives better typing to the event handler.

src/vnode.ts
```diff
@@ -12,22 +14,32 @@ export interface VNode {
...
+  };
+  class?: {
+    [name: string]: boolean
+  };
+  style?: VNodeStyle;
+  dataset?: {
+    [name: string]: string
+  };
+  on?: {
+    [event: string]: EventListener | EventListener[]
```
paldepind: This type is not correct. When given an array only the first element has to be an event listener. Changing `EventListener[]` to `any[]` would probably be the easiest fix. I at least can't think of a better way right now.

paldepind:
I definitely think this is an improvement over the current types. I think the two lines I left comments on should be tweaked. Besides that :+1: :tada:

src/vnode.ts
```diff
@@ -1,4 +1,12 @@
 import {Hooks} from './hooks';
+import {AttachData} from './helpers/attachto'
+import {VNodeStyle} from './modules/style'
+import {On} from './modules/eventlisteners'
+import {Attrs} from './modules/attributes'
+import {Classes} from './modules/class'
+import {Props} from './modules/props'
+import {Dataset} from './modules/dataset'
+import {Hero} from './modules/hero'
```
mightyiam: Now pretty much all of the types are exported in their own modules and imported to be used here. I find this cleaner, I guess.

src/modules/attributes.ts
```diff
@@ -1,6 +1,8 @@
 import {VNode, VNodeData} from '../vnode';
 import {Module} from './module';
 
+export type Attrs = Record<string, string | number | boolean>
```
mightyiam: @paldepind this is a change you asked for.

src/modules/eventlisteners.ts
```diff
@@ -1,6 +1,12 @@
 import {VNode, VNodeData} from '../vnode';
 import {Module} from './module';
 
+export type On = {
+  [N in keyof HTMLElementEventMap]: (ev: HTMLElementEventMap[N]) => void
+} & {
+  [event: string]: EventListener
+};
```
mightyiam: @paldepind is this fine? Notice that on line 6, while your example had `|` (union), I've found that `&` (intersection) works here.

src/modules/style.ts
```diff
@@ -1,6 +1,11 @@
 import {VNode, VNodeData} from '../vnode';
 import {Module} from './module';
 
+export type VNodeStyle = Record<string, string> & {
+  delayed: Record<string, string>
+  remove: Record<string, string>
+}
```
mightyiam: This is an improvement over the last commit, because it is slightly more specific, in that only `delayed` and `remove` can have `Record`s as values.

src/vnode.ts
```diff
@@ -12,22 +20,20 @@ export interface VNode {
...
-  hero?: any;
-  attachData?: any;
+  props?: Props;
+  attrs?: Attrs;
+  class?: Classes;
+  style?: VNodeStyle;
+  dataset?: Dataset;
+  on?: On;
+  hero?: Hero;
+  attachData?: AttachData;
```
mightyiam: This looks :sparkles:, to me.

src/modules/eventlisteners.ts
paldepind: I think that is fine as well :+1:

paldepind:
Looks good to me! Unless there are any objections I'll go ahead and merge it.

PR #273, opened 2017-04-13
Bugfix: thunk function called two times

caesarsol:
In some particular combination of events it was possible that the thunk function was called two times, wasting resources.

See the introduced [failing test](https://github.com/caesarsol/snabbdom/blob/bugfix-thunk-called-two-times/test/thunk.js#L62-L78) for details.

In addition, I noticed some files were git-ignored incorrectly, and fixed the issue.

A couple of side questions:
- Are you following a certain styleguide or linting configuration?
- I had some friction in running tests, are you interested on trying a new testing stack? I'm not an expert but I've used AVA with profit.
- I haven't understood how to run perf stats, do you have any suggestion? 

Thanks 😄


caesarsol:
@staltz friendly ping 😄

staltz:
Hi @caesarsol  :)
Thanks for submitting this PR. Before merging this PR, can I ask whether any of the tests added catches the case of thunk fn called twice? I mean, is there any test that fails if we remove the `return` that you added to prepatch?

> I had some friction in running tests, are you interested on trying a new testing stack? I'm not an expert but I've used AVA with profit.

What kind of friction?

> Are you following a certain styleguide or linting configuration?

Not a specific one, but usually in the test files we don't have empty lines in between. I could remove them myself after your PR, but if you could remove them that would be a nice bonus.

caesarsol:
Sure, sorry for the non-explicitness. 

In the test-failing commit the `it('calls render function on data-length change')` is failing, because the thunk function is called twice. Then after the fixing commit (the one adding the `return`) the test passes.

I'm correcting the blank lines right now! (I realised too late it was a coherent style choice...)

caesarsol:
About the friction in testing, my problems were:

- the tests aren't performed on all browsers present on my machine
- every failing test has a double output (maybe every test is performed two times?)
- the failing tests are difficult to identify due to compiled line number (example: `browserified.js:3985:5`)
- most importantly, the code is not automatically recompiled and browserified before every test run in watch mode! apparently only the bundle is watched.

Let me know if you'd like any help on these issues! 
(Or if you find them not to be issues...)

staltz:
I think for this PR it seems ready to merge. But I'm typing this on my phone so I'll check carefully later on the computer. Just to say we don't need to block this PR further, about those testing friction issues

caesarsol:
That's right, sorry for the OT!
I opened an issue on it if you're interested: https://github.com/snabbdom/snabbdom/issues/288

staltz:
Looks good, thanks!

PR #276, opened 2017-04-19
Fix error when class is set by vdom selector in SVG

### LINKED ISSUE #217: Unable to create svg elements with CSS selectors
Something like this never works —

```js
h('line.form-field-underline-default', {attrs: {x1: '0', y1: '0', x2: '100', y2: '0'}})
```

**ERROR:**
```
Uncaught TypeError: Cannot assign to read only property 'className' of object '#<SVGLineElement>'
    at createElm (snabbdom.js:75)
    at createElm (snabbdom.js:78)
    at createElm (snabbdom.js:78)
    at createElm (snabbdom.js:78)
    at createElm (snabbdom.js:78)
    at updateChildren (snabbdom.js:172)
    at patchVnode (snabbdom.js:212)
    at updateChildren (snabbdom.js:151)
    at patchVnode (snabbdom.js:212)
    at updateChildren (snabbdom.js:151)
```

caesarsol:
Fixes #217 

Use `elm.setAttribute('class', ...)` instead of `elm.className = ...` to make it work in SVGs.

Not absolutely sure it does not have implications, but `setAttribute` is in scope here since we are talking about Elements creation (and not mutation, in which className is recommended).

Another option would be to write in the attribute `className.baseVal` when in SVG namespace.

### DIFF
--- a/src/snabbdom.ts
+++ b/src/snabbdom.ts
@@ -98,8 +98,8 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
       const tag = hashIdx !== -1 || dotIdx !== -1 ? sel.slice(0, Math.min(hash, dot)) : sel;
       const elm = vnode.elm = isDef(data) && isDef(i = (data as VNodeData).ns) ? api.createElementNS(i, tag)
                                                                                : api.createElement(tag);
-      if (hash < dot) elm.id = sel.slice(hash + 1, dot);
-      if (dotIdx > 0) elm.className = sel.slice(dot + 1).replace(/\./g, ' ');
+      if (hash < dot) elm.setAttribute('id', sel.slice(hash + 1, dot));
+      if (dotIdx > 0) elm.setAttribute('class', sel.slice(dot + 1).replace(/\./g, ' '));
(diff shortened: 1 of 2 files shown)


caesarsol:
I will need to update the docs, in case your response is positive!

staltz:
Hi :)
Thanks for submitting this PR. This looks good from a functionality perspective, but before merging the PR, what is the implications for performance? I'm not saying it will have, but it's always important to check if it could possibly be slower/faster. Ideally we would have automated perf benchmarks instead of making this type of question.

staltz:
Great, let's merge it then. :)

caesarsol:
Thanks! The benchmark is actually very interesting. Is that result valid in all browsers?

I tried to do the the benchmarking in the repo, bit I didn't get how does it work.

I'll update the README section in a few days, to remove the alert and make it reflect the change.

caesarsol:
I'd like to try to automate testing and perf benckmark for this prpject!

staltz:
> I'd like to try to automate testing and perf benckmark for this prpject!

This is super needed. Please do!

PR #280, opened 2017-04-23
Tiny loop perf optimization

dmitriz:
Don't know how much it matters, but still.

### DIFF
--- a/src/h.ts
+++ b/src/h.ts
@@ -4,7 +4,7 @@ import * as is from './is';
 function addNS(data: any, children: Array<VNode> | undefined, sel: string | undefined): void {
   data.ns = 'http://www.w3.org/2000/svg';
   if (sel !== 'foreignObject' && children !== undefined) {
-    for (let i = 0; i < children.length; ++i) {
+    for (let i = 0, l = children.length; i < l; ++i) {
       let childData = children[i].data;
       if (childData !== undefined) {
         addNS(childData, (children[i] as VNode).children as Array<VNode>, children[i].sel);
--- a/src/h.ts
+++ b/src/h.ts
@@ -33,7 +33,7 @@ export function h(sel: any, b?: any, c?: any): VNode {
     else { data = b; }
   }
   if (is.array(children)) {
-    for (i = 0; i < children.length; ++i) {
+    for (var i = 0, l = children.length; i < l; ++i) {
       if (is.primitive(children[i])) children[i] = (vnode as any)(undefined, undefined, undefined, children[i]);
     }
   }


fix-fix:
First of all, any performance optimizations are worthless without measurements. So if the changeset's proposition is not clear from the code, it's difficult to act upon, and it just makes reviewers either spend more time on it or just ignore it. You shouldn't take any wild guesses in form of PRs before evaluating them first.

Second, this kind of change wouldn't make any significant difference even if it was 10x times slower, because it's not related to the core logic and computations. As pointed out in [Rules Of Optimization](http://wiki.c2.com/?RulesOfOptimization) you should profile your program first and look for what's really slowing it down, and only then try to optimize it.

As for this specific change, caching a length of arrays stopped being relevant to performance a long time ago (you can run [this benchmark ](https://jsperf.com/caching-array-length/4) in any modern browser), and may even hurt (as seen in [this article ](http://mrale.ph/blog/2014/12/24/array-length-caching.html) from former V8 engineer).

Thanks for looking into the code and trying to squeeze even more performance though, any perf optimizations suggestions are welcome here.

dmitriz:
Many thanks @fix-fix for your detailed comments and the links with interesting information.
Ironically

dmitriz:
Many thanks @fix-fix for your detailed comments and the links with interesting information. I am always happy to learn new things and different points of view. Ironically, that benchmark site actually does show the cached loop to be the fastest 😄 

But even if it didn't, in my humble opinion, there is hardly a reliable way to compare performance for a general purpose library, such as snabbdom. There are so many different ways the user can employ the library in many different situations, with many other details affecting the performance difference. At least I would see it for a complete stable application, and even then features can change, new versions updated and so on.

The Rules of Optimisation are totally correct when applied to optimisations requiring time and effort. But that particular one is so cheap that I'd argue they don't.  Also I have yet to see any perf comparison where a caching would score considerably worse. ;) 

In that article, if I understand it correctly, the author argues that the compiler already might possibly take care  of caching (or maybe not?) and I certainly know too little about that subject. But even if it did, it would need possibly expensive tracking of all possible changes of all possible variables, including deep nested objects, as the compiler can't have any idea what can ever change. Plus also compilers, browser versions, vendors, implementations etc. can change any time (or they might have dropped it already since the author left).

In contrast, in my opinion, the programmer knows exactly which variables can be cached, and so can make often cheap improvements that often also result in more expressive and readable code (by making more explicit variables that don't mean to change) and also preventing possible bugs. Which was actually my main reason, together with improving the code consistency (since there are other loops where the `.length` is actually cached), rather than performance, which in this case can often be tiny (as per title ;). For that reason, it looked more like a typo to me, and if it doesn't immediately look so to the maintainers, I'd rather leave it at that, as who am I to argue?

fix-fix:
> that benchmark site actually does show the cached loop to be the fastest

That's the hard truth about microbenchmarks: they almost always don't give enough information applicable to the real world scenarios. If we can't actually measure which way is faster, how can we decide whether this change speeds things up or not?

I just think that any perf change of this type (refining code, and not changing algorithms or general structure) should be taken a with grain of salt and should be measured first, because it's really hard to know the outcome without making additional assumptions.

You are definitely right about not assuming how this lib used and in this case this change can't go wrong, and catching inconsistencies is good thing too. These are actually good reasons for this PR, but they were not included in the original description and it was harder to review it without (restricted) context (and maybe that's why I commented on the broader issues).

I'm not against this change, since it doesn't increase complexity (most devs are familiar with this trick) and won't likely cause any performance issues in the future. But in all modern browsers and runtimes that I tested the differences were negligible, so that's why I started discussion not about the actual change, but about cost of the extra contributions without significant effects and extra code cost.

However I'm not a member of this project and was just passing by, so you should wait the contributors' stance on this issue and ignore my thoughts.

diervo:
Most of engines will optimize and memoize the length property if the code is side-effect free (no mutations on the reachable code). For this case we will have to look at some of the V8 traces to see what is the result is to even know if is an optimization at all. 

That being said, even then as @fix-fix pointed about, the benefits are negligible. Moreover, if we were to do this optimization, we should do it throughout the whole project.

Since we are talking about perf, note that for example the very first line `sel !== 'foreignObject'` is already more expensive that all the .lenghts you can find (strings comparisons are very expensive for perf), so if you guys are looking for perf improvements, there is a lot of low hanging fruits waiting to be fixed :)

I will try to do a manual analysis and close this accordingly.
Also will have perf benchmark framework integrated soon.

paldepind:
I'm closing this. This should be optimized by JS engines. If we are to make this change it should be backed up by benchmarks showing performance benefits.

PR #287, opened 2017-04-30
Update and improve docs section "Using Classes in SVG Elements"

caesarsol:
Docs updates taking #276 into account.

Also improved the explanation on the SVG classList IE bug, with link to caniuse.com, and with new example on a workaround. (BTW, I'm going to fix that classList problem also, in the next weeks)

I assumed the next version would be 0.6.7, and wrote
> The classes in selectors for SVG elements work fine from version 0.6.7.

Please correct me if you are going to make it 0.7.0!

### DIFF
--- a/README.md
+++ b/README.md
@@ -487,22 +487,28 @@ var vnode = h('div', [
 
 See also the [SVG example](./examples/svg) and the [SVG Carousel example](./examples/carousel-svg/).
 
-#### Using Classes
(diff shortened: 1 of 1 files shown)


paldepind:
Thank you for this PR 👍 It's always nice when a great PR is followed up with matching documentation 😄

caesarsol:
It was a pleasure! 😉

Issue #262, opened 2017-03-17
Can't generate draggable elements using attributes module


sergkop:
When I try to generate a draggable element using `h('div', {attrs: {draggable: 'true'}}, 'text')` I get `<div draggable>text</div>` instead of `<div draggable="true">text</div>`. The former element is not draggable in browsers.

The problem is that the [attributes module](https://github.com/snabbdom/snabbdom/blob/master/src/modules/attributes.ts) treats "draggable" as a boolean attribute, while it isn't (according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable) "true" value must be provided).

Otherwise, I love this library! It was easy to integrate into an existing project and performance is amazing. Thanks for the great work!

caridy:
Related to the discussion on #230 and #254.

In principle we can just remove `draggable` from that list since it is `Enum` instead of `Boolean`, according to the docs. A PR will be welcome :)

caridy:
@sergkop make sure you pass the string value "true", "false" or "auto" to be forward-compatible :)

astroK8:
PR created.

GitHubStig:
I had an issue with `draggable` attribute not set properly when using attribute module to add it, after switching to the property module, it works fine.

astroK8:
@GitHubStig this is a good workaround that can be used until modification of attributes module.

There is however a conceptual difference between attributes and properties. Attributes are strings in the html code. Theses html attributes are parsed by the browser and added as properties to the DOM node object. If you manually set a property on the DOM node, you just won't see it in the html code in the debugger (not an issue in many cases ;-)).

caridy:
>  If you manually set a property on the DOM node, you just won't see it in the html code in the debugger (not an issue in many cases ;-)).

@GitHubStig most attributes are reflective to a property, and they will appear in the DOM either way.

astroK8:
@caridy I did the test for draggable, it is effectively reflective attribute.

pme123:
when will this fix be released?

paldepind:
This has been fixed.

PR #289, opened 2017-05-05
PR for issue #262 (remove draggable of boolean attributes)

astroK8:
Remove draggable from boolean attributes enum.

"This attribute is an enumerated one and not a Boolean one. This means that the explicit usage of one of the values true or false is mandatory and that a shorthand like \<label draggable>Example Label\</label> is not allowed. The correct usage is "\<label draggable="true">Example Label\</label>."


paldepind:
That is true. Thank you.

PR #290, opened 2017-05-09
VNode key property can be undefined

yarom82:
Because actually it can be undefined:
https://github.com/snabbdom/snabbdom/blob/4adbd971cba738c38d6158b38fdb81a6509d1553/src/vnode.ts#L44

@mightyiam


staltz:
I can't see any CI results, so just to make sure, can you run the tests? I see the PR was made with the branch `patch-1` which is from the github editor.

yarom82:
I ran the tests and they passed.

staltz:
Thanks, seems correct.

PR #307, opened 2017-06-16
Delete props only when they become undefined

nornagon:
This allows setting a prop to `0` or `false`.


fix-fix:
This is a breaking change, isn't it?

nornagon:
It doesn't seem to break the tests :)

caridy:
yes, this is a breaking change. Also, typeof is slow for critical path. I don't fully understand the rational of this change though.

staltz:
Well, I see the utility of allowing prop values like `0` and `false`, just need to use `if (props[key] == null)`.

mightyiam:
#624

PR #308, opened 2017-06-19
handle vnode with undefined data

yarom82:
There was a run time error when used with vnode with undefined data.
API allows to provide undefined data.

### DIFF
--- a/test/eventlisteners.js
+++ b/test/eventlisteners.js
@@ -169,4 +169,13 @@ describe('event listeners', function() {
     elm.firstChild.click();
     assert.equal(3, result.length);
   });
+  it('handle vnode with undefined data', function () {
+    assert.doesNotThrow(function() {
+        var elm1 = document.createElement('span');
+        var elm2 = document.createElement('div');
+        var vnode1 = {sel:'span', elm: elm1, data: undefined};
+        var vnode2 = {sel:'div', elm: elm2, data: undefined};
+        patch(vnode1, vnode2);
+      });
+  });
 });
--- a/test/style.js
+++ b/test/style.js
@@ -116,6 +116,15 @@ describe('style', function() {
     fakeRaf.step();
     assert.equal(elm.style.fontSize, '20px');
   });
+  it('handle vnode with undefined data', function() {
+    assert.doesNotThrow(function() {
+      var elm1 = document.createElement('span');
+      var elm2 = document.createElement('div');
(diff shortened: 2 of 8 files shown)


test/eventlisteners.js
```diff
@@ -169,4 +169,13 @@ describe('event listeners', function() {
     elm.firstChild.click();
     assert.equal(3, result.length);
   });
+  it('handle vnode with undefined data', function () {
```
yarom82: Could be redundant as we have this test at test/core.js

test/style.js
```diff
@@ -116,6 +116,15 @@ describe('style', function() {
     fakeRaf.step();
     assert.equal(elm.style.fontSize, '20px');
   });
+  it('handle vnode with undefined data', function() {
```
yarom82: Could be redundant as we have this test at test/core.js

caridy:
Can you explain the rationale about this? Why will you want to have a vnode without data (even if it is empty)? Snabbdom is built for speed, `h()` is just an artifact for people to build an object that represent a vnode, but you can build your own `h` with your own logic. For example, I normally have my own `h()` that returns the same empty object (`Object.create(null)`) for data.

mightyiam:
@caridy either some variant of this PR or make [`data`](https://github.com/snabbdom/snabbdom/blob/7b45afea49a32e2f4b0c5f12ca7df0040fd1dcf7/src/vnode.ts#L15) always `VNodeData`.

caridy:
@mightyiam I'm reluctant to merge this PR because it includes many checks that must be performed during the critical path, when the same thing can be achieve in user-land by adding a simple abstraction around `h()` to produce valid vnodes. I don't think snabbdom should be in the business of validation, which is something that I have mentioned many times before.

mightyiam:
Then would it be fine to type the `data` property of `VNode` as always `VNodeData` instead of `VNodeData | undefined`? Otherwise, what seems to be a valid `VNode`, with `data: undefined`, causes a runtime error.

caridy:
@mightyiam yes, that seems fine. we can probably adjust `h()` to enforce that data cannot be `undefined`, to eliminate all kind of critical path checks.

staltz:
I agree with @caridy. Let's not merge it (or not merge it *yet*).

yarom82:
Closing in favor of PR #329

PR #313, opened 2017-06-28
fix(dataset): check if property exist before deleting

### LINKED ISSUE #309: deleting a dataset property that doesn't exist throws an error
Found out that Safari automatically deletes the dataset property when you do `use strict;`

So when [this line of code](
https://github.com/snabbdom/snabbdom/blob/7b45afea49a32e2f4b0c5f12ca7df0040fd1dcf7/src/modules/dataset.ts#L23) run, it may error. So maybe an extra checking? `if (d && d[key]) {`

I have a PR ready

FeliciousX:
in Safari, when it is `use strict`, if the dataset property is removed in DOM
deleting it again will cause an error

closes #309

### DIFF
--- a/src/modules/dataset.ts
+++ b/src/modules/dataset.ts
@@ -19,7 +19,7 @@ function updateDataset(oldVnode: VNode, vnode: VNode): void {
 
   for (key in oldDataset) {
     if (!dataset[key]) {
-      if (d) {
+      if (d && d[key]) {
         delete d[key];
       } else {
         elm.removeAttribute('data-' + key.replace(CAPS_REGEX, '-$&').toLowerCase());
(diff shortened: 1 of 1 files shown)


src/modules/dataset.ts
```diff
@@ -19,7 +19,7 @@ function updateDataset(oldVnode: VNode, vnode: VNode): void {
 
   for (key in oldDataset) {
     if (!dataset[key]) {
-      if (d) {
+      if (d && d[key]) {
```
caridy: this is not enough. this will fail if `elm.dataset.x = 0` or any other falsy value. you will have to change this to `if (d && key in d) {` to only delete it if the key exists.

caridy: ... but if you look at the other code path with the `removeAttribute`, we don't want to invoke that one if dataset is available (which is for HTMLElement but not for Element). So, I guess you just need to protect the delete invocation only. something like:

```
if (d) {
   if (key in d) {
       delete d[key];
   }
} else {
   ...
}
```

FeliciousX:
done :D @caridy

paldepind:
Great! Thank you 😄

Issue #306, opened 2017-06-13
Adding the AMP boolean attributes to the attributes module


zkochan:
I am trying to render an [AMP HTML](https://www.ampproject.org/) page using snabbdom. AMP has some custom boolean attributes: amp, ⚡, amp-boilerplate (might be more, I don't know yet).

Would it be fine to add those attributes to the list of boolean attributes [here](https://github.com/snabbdom/snabbdom/blob/master/src/modules/attributes.ts#L6)

Note: currently snabbdom renders those as `amp="true"` and that does not pass the AMP validator.

* * *

I can work on a fork and try to make all the necessary changes that need to be done for snabbdom to be able to render AMP-compliant pages

zkochan:
This might be an issue with [snabbdom-to-html](https://github.com/snabbdom/snabbdom-to-html). I am not sure yet.

fix-fix:
I think the only clear way (without exposing attributes module internals) of making this possible is to define a separate module for custom handling of VNode's attributes from AMP.

However, the main problem is that AMP is non-compliant with HTML standard and those attributes are not valid at all, and it's should be avoided.

zkochan:
So a separate module has to be created and used instead of the core attributes module, right?

fix-fix:
Yes, but I don't think that module should replace attributes module, but instead there should be a special way for providing AMP (and maybe some other non-standard) attributes to VNodes (not in the `data` property), that can be used alongside default attributes module.

caridy:
There is an issue somewhere to remove the list of booleans, and instead rely on the value to determine if the attribute is a boolean or not and act accordingly. It will be a mostly backward compatible solution, but there are cases in which it will break some folks. We need this for custom elements as well, and I suspect we are just waiting for someone to send a PR :)

paldepind:
Thank you for reporting this issue @zkochan. The latest release now handles custom boolean attributes seamlessly.

`h("div", { attrs: { amp: true} }` results in `<div amp>`.

zkochan:
Awesome! Thanks!

PR #314, opened 2017-06-28
fixes issue #306 and others: boolean attributes alignment with HTML

caridy:
_note: this change is not backward compatible_

* if the attribute `foo` is set to `true` in the `vnode.data.attrs`, it will appear in the dom as `<div foo>`
* if the attribute `foo` is set to `false` in the `vnode.data.attrs`, it will not appear in the dom as `<div>` 
* snabbdom does not provide any validation, nor a whitelist of boolean, that's a responsibility of the creator of the vnode.
* falsy values will be considered as regular attributes, and the value will be casted to string, just like html.
   * if the attribute `foo` is set to `"false"` in the `vnode.data.attrs`, it will appear in the dom as `<input readonly="false">`, which implies that the input is readonly. 
* perf optimization (this change should improve the perf a little bit)

### DIFF
--- a/src/modules/attributes.ts
+++ b/src/modules/attributes.ts
@@ -3,22 +3,10 @@ import {Module} from './module';
 
 export type Attrs = Record<string, string | number | boolean>
 
-const booleanAttrs = ["allowfullscreen", "async", "autofocus", "autoplay", "checked", "compact", "controls", "declare",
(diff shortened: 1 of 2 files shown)


paldepind:
Fantastic! I noticed the exact same thing when I saw #306—that this privously discussed change would fix it.

The breaking change seems to be for people who currently use truthy/falsy values to toggle boolean attributes. In my opinion that is bad practice and I assume it's rare.

PR #320, opened 2017-07-17
Support DocumentFragment in toVNode()

staltz:
This PR is justified on a general level and a specific level.

In general: I believe we don't have a reason to set `vnode.elm` to undefined inside `toVNode()`, no matter what the node is.

In specific: this PR is necessary to support patching inside a DocumentFragment.


paldepind:
@staltz 

> In general: I believe we don't have a reason to set `vnode.elm` to undefined inside `toVNode()`, no matter what the node is.

Good point. I agree. Always setting `elm` seems more consistent.

PR looks good 👍  I'll merge and do a release since there's some other changes I'd like to get out as well.

PR #321, opened 2017-07-20
Wrong example name "Reorder Animation" to "SVG"

ChangJoo-Park:


paldepind:
Nice. Thank you :smile:

PR #325, opened 2017-07-29
Fix VNodeStyle

whitecolor:
This allows using `delayed`, `remove`, `destroy` properties, the previous variant gives the typing error.

image

### DIFF
--- a/src/modules/style.ts
+++ b/src/modules/style.ts
@@ -1,9 +1,13 @@
 import {VNode, VNodeData} from '../vnode';
 import {Module} from './module';
 
-export type VNodeStyle = Record<string, string> & {
-  delayed?: Record<string, string>
-  remove?: Record<string, string>
+export type VNodeStyle = {
+  delayed?: { [prop: string]: string };
+  remove?: { [prop: string]: string };
+  destroy?: { [prop: string]: string };
+  [prop: string]: string | {
+    [prop: string]: string
+  } | undefined
 }
 
 var raf = (typeof window !== 'undefined' && window.requestAnimationFrame) || setTimeout;


src/modules/style.ts
```diff
@@ -1,9 +1,13 @@
...
-export type VNodeStyle = Record<string, string> & {
-  delayed?: Record<string, string>
-  remove?: Record<string, string>
+export type VNodeStyle = {
+  delayed?: { [prop: string]: string };
+  remove?: { [prop: string]: string };
+  destroy?: { [prop: string]: string };
+  [prop: string]: string | {
+    [prop: string]: string
+  } | undefined
```
mightyiam: Thank you for this. You're right about there being an error in the type. It is my code. My error 🐛 . There is a simpler, smaller and more correct fix, I believe. It consists only of replacing the `&` (intersection) with a `|` (union).

```
type VNodeStyle = Record<string, string> | {
  delayed?: Record<string, string>
  remove?: Record<string, string>
}
```

Here is a [live example](http://www.typescriptlang.org/play/index.html#src=type%20VnodeStyleA%20%3D%20Record%3Cstring%2C%20string%3E%20%7C%20%7B%0D%0A%20%20%20%20delayed%3F%3A%20Record%3Cstring%2C%20string%3E%0D%0A%20%20%20%20remove%3F%3A%20Record%3Cstring%2C%20string%3E%0D%0A%7D%0D%0A%0D%0Alet%20styleA%3A%20VnodeStyleA%20%3D%20%7B%0D%0A%20%20%20%20'foo'%3A%20'bar'%2C%0D%0A%20%20%20%20delayed%3A%20%7B%0D%0A%20%20%20%20%20%20%20%20foo%3A%20'bar'%0D%0A%20%20%20%20%7D%0D%0A%7D%0D%0A%0D%0Atype%20VNodeStyleB%20%3D%20%7B%0D%0A%20%20delayed%3F%3A%20%7B%20%5Bprop%3A%20string%5D%3A%20string%20%7D%3B%0D%0A%20%20remove%3F%3A%20%7B%20%5Bprop%3A%20string%5D%3A%20string%20%7D%3B%0D%0A%20%20destroy%3F%3A%20%7B%20%5Bprop%3A%20string%5D%3A%20string%20%7D%3B%0D%0A%20%20%5Bprop%3A%20string%5D%3A%20string%20%7C%20%7B%0D%0A%20%20%20%20%5Bprop%3A%20string%5D%3A%20string%0D%0A%20%20%7D%20%7C%20undefined%0D%0A%7D%0D%0A%0D%0Alet%20styleB%3A%20VNodeStyleB%20%3D%20%7B%0D%0A%20%20%20%20'foo'%3A%20'bar'%2C%0D%0A%20%20%20%20delayed%3A%20%7B%0D%0A%20%20%20%20%20%20%20%20foo%3A%20'bar'%0D%0A%20%20%20%20%7D%0D%0A%7D).

mightyiam: Oh, and I don't know what that `destroy` property is, so no comment on that.

whitecolor: I tried this too, the problem with that variant is that it allows any properties also:
image

mightyiam: Oh! Sorry. Could you please point me at the right direction understanding why any value is possible like that?

whitecolor: I don't know, this is how TS handles it.

jvanbruegge: @mightyiam this is because of subtyping
```
{ delayed: { /* ... */ }, 10: true }
```
is a subtype of `{ delayed: { /* ... */ }}`. Think about class inheritance if the child class has additional properties it *is* still of type base class

jvanbruegge:
Can this be merged? Just noticed this again when trying to add better typings to hyperscript helpers in cyclejs

jvanbruegge:
@paldepind ping

Austaras:
since this pr ain't going nowhere and a decent fix won't happen unless typescript has some improvement, what about use CSSType like react instead?

kay999:
Bump... Very annoying and easy to fix, please already merge it.

mightyiam:
Closing in favor of #339.

PR #326, opened 2017-08-09
Add package.json/module

AlexGalays:
Modern module bundlers (e.g webpack 3) can generate a more efficient output when working with ES6 modules.

This PR adds the package.json `module` field, which is not standard (yet?) but is now broadly used by the community. See [here](https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md) for detailed info.

This `module` field points to a new directory where the module syntax is ES6 but everything else is ES5. I usually like to separate the two builds into `commonjs` and `es` folders but this would break backward compatibility when importing modules like `snabbdom/modules/props`. I kept the root file structure for the commonjs version so that existing users using commonjs modules are not impacted.

I didn't touch the README as it's been needing some love for a while and that could be done on the side (it still mentions using require(module).default)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.7.0",
   "description": "A virtual DOM library with focus on simplicity, modularity, powerful features and performance.",
   "main": "snabbdom.js",
+  "module": "es/snabbdom.js",
   "typings": "snabbdom.d.ts",
   "directories": {
     "example": "examples",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.7.0",
   "description": "A virtual DOM library with focus on simplicity, modularity, powerful features and performance.",
   "main": "snabbdom.js",
+  "module": "es/snabbdom.js",
```
caridy: `"jsnext:main": "es/snabbdom.js",` should also be used, some tools are still using that one.

paldepind:
Since `jsnext:main` is deprecated I think not having it is ok. As far as I understand tools that use it should use `module` instead.

PR #327, opened 2017-08-20
allows symbol keys (fix #124)

### LINKED ISSUE #124: Allow symbol keys
Snabbdom allows numbers or strings as keys.
It would also be really powerful if it could open for ES6 Symbols as keys.

guri:


test/core.js
```diff
@@ -615,6 +619,16 @@ describe('snabbdom', function() {
           assert.deepEqual(map(inner, elm.children), arr.filter(function(x) {return x != null;}));
         }
       });
+      it('accepts symbol as key', function() {
+        var vnode1 = h('span', [Symbol()].map(spanNum));
+        var vnode2 = h('span', [Symbol('1'), Symbol('2'), Symbol('3')].map(spanNum));
+        elm = patch(vnode0, vnode1).elm;
+        assert.equal(elm.children.length, 1);
```
caridy: we don't really care about children that much, but about properties set on the element.

caridy: e.g.: 

```js
const s = Symbol();
var vnode1 = h('span', { s: 100 });
...
assert.equal(elm[s], 100);
```

mightyiam: I'm sorry. I'm the culprit. I have guided @guri through this PR. I was certain that this is about having symbols as `key`s. @aronallen, what did you mean in #124, please?

caridy: I understand now! but still this test is not enough! if what we want to test is that key could be a symbol, then we need at least 2 tests:

1. test that patching two similar vnodes with the same symbol as a key, reuses the element
2. test that patching two similar vnodes with different symbol as key, does not reuse the element

paldepind:
As far as I can tell this PR does not achieve what it intends to. The primary way that Snabbdom uses keys is to build the `oldKeyToIdx` map [here](https://github.com/snabbdom/snabbdom/blob/e203f711c60fde72fa391f0a8e927df4a376204b/src/snabbdom.ts#L226). But you can only use strings as keys in plain objects. So in order to make symbols work we should use a `Map`. Using `Map` also has the benefit that it's faster.

It seems to me that the best solution to the problem is to use `Map` in browsers that supports it. Then everything would work as keys and we would get better performance.

mightyiam:
@paldepind 

> But you can only use strings as keys in plain objects

You can use [symbols as property keys](http://2ality.com/2014/12/es6-symbols.html#symbols-as-property-keys).

caridy:
I think we are all confused by this PR :) ... I see two things:

* the special `data.key` could be a symbol (should not be too problematic since it is just a primitive)
* two similar elements with keys set to symbols should be compared correctly (symbols has identity, so they must be triple equal)
* DOM properties can be symbols, and the diffing algo should diff those symbols accordingly, and set them and remove them from the DOM element when needed.

Is there anything else missing?

paldepind:
@mightyiam 

> You can use symbols as property keys.

You're absolutely right. That was a blunder on my part.

I would like to merge this PR but the current test doesn't make sense. I think changing it to one of these would be fine:

1. Remove the added tests altogether. Since this PR essentially only changes the types and not the code not adding any tests would be acceptable.
2. Add a test that actually uses symbols as keys and check that reordering of shuffled elements happens correctly.

Once this PR is merged I'd like to take a stab at implementing the idea I mentioned above https://github.com/snabbdom/snabbdom/pull/327#issuecomment-324672615.

mreinstein:
resolved via https://github.com/snabbdom/snabbdom/commit/ad80c6ea530a058501465cfcb45698e380ceb96a

PR #328, opened 2017-08-28
Tiny type check in call to `vnode` at `h.ts`

yarom82:


yarom82:
This is just a tiny and obvious change.
Please review.

staltz:
Seems like a harmless fix. Thanks :)

PR #329, opened 2017-08-28
`VNodeData` property is mandatory

yarom82:


src/snabbdom.ts
```diff
@@ -138,15 +136,13 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
+    if (isDef(vnode.sel)) {
       if (isDef(i = data.hook) && isDef(i = i.destroy)) i(vnode);
       for (i = 0; i < cbs.destroy.length; ++i) cbs.destroy[i](vnode);
       if (vnode.children !== undefined) {
         for (j = 0; j < vnode.children.length; ++j) {
           i = vnode.children[j];
-          if (i != null && typeof i !== "string") {
-            invokeDestroyHook(i);
-          }
+          invokeDestroyHook(i);
```
paldepind: I don't think you can leave this `if` statement out. Children can potentially be strings or `undefined`.

mightyiam: At least in TypeScript, [children is](https://github.com/snabbdom/snabbdom/blob/e203f711c60fde72fa391f0a8e927df4a376204b/src/vnode.ts#L16) `Array<VNode | string> | undefined`.

So yes, the condition here should be changed accordingly.

Is the reuse of `i` here really helping with performance? it makes it hard for the typing. Perhaps that is why it is `any`. But this code might have been converted to TypeScript, and not originally written with TypeScript types in mind. Would it be welcome to "explode" this reuse of `i`, so that we would have better types and so the IDE would actually help in development with this variable (preventing such an error as this)?

src/snabbdom.ts
```diff
@@ -138,13 +136,13 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
     let i: any, j: number, data = vnode.data;
-    if (data !== undefined) {
+    if (isDef(vnode.sel)) {
       if (isDef(i = data.hook) && isDef(i = i.destroy)) i(vnode);
       for (i = 0; i < cbs.destroy.length; ++i) cbs.destroy[i](vnode);
       if (vnode.children !== undefined) {
         for (j = 0; j < vnode.children.length; ++j) {
           i = vnode.children[j];
-          if (i != null && typeof i !== "string") {
+          if (typeof i !== "string") {
```
yarom82: Hey @paldepind ,
As you mentioned, checking if child is string.
According to child type, it can only be VNode | string.

caridy:
I'm supportive of this considering that in all my projects, I do enforce this rule on top of the snabbdom, and pushing it down to snabbdom seems like a good option here because it could be a nice perf improvement (to be validated). @yarom82 I will add few more notes to the diff.

The concern here is that this is a non-backward compatible change for folks creating their own vnodes today (skipping the h provided by this library), but I think this is something that we can document well, and handle via semver.

src/h.ts
```diff
@@ -34,7 +34,7 @@ export function h(sel: any, b?: any, c?: any): VNode {
   }
   if (is.array(children)) {
     for (i = 0; i < children.length; ++i) {
-      if (is.primitive(children[i])) children[i] = vnode(undefined, undefined, undefined, children[i], undefined);
```
caridy: creating a new object everytime while we were using `undefined` before will be a perf penalty.

snabbdom never mutates the VNodeData structure itself, in which case reusing an empty object should be ok (and favored), but this object might need to be marked as non-extensible so 3rd party modules might not attempt to mutate it, this itself might be problematic (non-backward compatible).

yarom82: @caridy making `data` "non-extensible" (immutable?) would not only be a breaking change, it would be undesirable.
Users might want to mutate the `data` of VNodes returned from `h`. I know because I do it.

Oh, but this line is only for text VNodes.
I guess there is no foreseeable reason to mutate a text VNode's `data`. Is there?

So, a top scope empty object that will be reused as `data` of all text VNodes?

Should I `Object.freeze` it?

As far as TypeScript, I don't see a reasonable way to type the `data` of a text VNode as an empty object. The only way I can see is that we have a new interface, `TextVNode` and that seems kind of overkill to me. So how about just document this caveat and `Object.freeze` to prevent mistakes?

Is there a standard way of running benchmarks, so I could test the possible perf penalty of this new object, at all? Who knows—perhaps engine optimizations make this a non-issue.

paldepind:
I'm closing this PR due to its age. Personally, I'm not necessarily opposed to this, but I'm also not sure what the benefits are either. Perhaps this is less needed in an age with the `.?` operator? If anyone is still interested in this then feel free to open a new issue or PR with motivation included.

PR #334, opened 2017-10-03
stencil uses snabbdom.

zigomir:


README.md
```diff
@@ -656,7 +656,8 @@ Here are some approaches to building applications with Snabbdom.
 * [Cycle.js](https://cycle.js.org/) –
   "A functional and reactive JavaScript framework for cleaner code"
   uses Snabbdom
-* [Vue.js](http://vuejs.org/) use a fork of snabbdom.
+* [Vue.js](http://vuejs.org/) uses a fork of snabbdom.
+* [stencil](https://stenciljs.com/) uses a [fork](https://github.com/ionic-team/stencil/blob/6193e74c9cca2ed7038c7e5fdf1fd925f8eba831/src/core/renderer/patch.ts) of snabbdom.
```
caridy: I don't think we need to link to the file, just specified that this is a fork.

caridy: I will love to see why do they need to fork it, if anyone has more info about it, and see if we can provide the APIs/features that they need to avoid the fork.

blikblum:
> I will love to see why do they need to fork it, if anyone has more info about it

They are [manually](https://github.com/ionic-team/stencil/blob/master/src/core/renderer/patch.ts#L27-L65) handling the [slot](https://developers.google.com/web/fundamentals/web-components/shadowdom#slots) projection. 

There's also some adaptation to handle server side rendering.

paldepind:
@zigomir Why did you close the PR? We would like to have it on the list with the requested changes 😄

caridy:
@blikblum is there anything that we can do in snabbdom directly to accomodate those two features? One of my projects is doing something similar with slots without having to patch snabbdom, maybe we could document how to do so. As for server side rendering, we will need more info about what was patched, to see if we can accomodate that too.

Please, reopen the PR, update the link according to our request, so we can merge it :)

zigomir:
Hey, sorry for closing this, I just didn't bother editing :P. Now I can't reopen it because of: `The patch-1 branch was force-pushed or recreated`.

blikblum:
Sorry, i don't have enough info to answer that, i'm just a snabbdom user that also follows stencil development.

Incidentally i'm interested in a snabbdom implementation of slot projection for custom elements without shadowDOM

caridy:
>  Incidentally i'm interested in a snabbdom implementation of slot projection for custom elements without shadowDOM

me too! /cc @diervo 

## Rant

The main situation today is that as part of the patching process (going thru modules) does not allow you to rearrange the children list, because that reference is cached before patching the node. As a result, you have to do some gymnastics, e.g.:

```js
// in your snabbdom module, update/create method:
const { data: { slots }, children } = vnode;
const newChildrenCollection = rearrangeChildrenAndSlots(children, slots); 
vnode.children.length = 0;
Array.prototype.push.apply(vnode.children, newChildrenCollection);
```

__note: the code above mutates the original children list provided to the vnode, which means it might have side effects, not good.__

This is somehow horrendous, and error prompt. Maybe we should not cache that value, and just collect it after the patching of the element finishes, and before patching the children, so modules can actually replace the `vnode.children` with a new computed list of children.

PR #337, opened 2017-10-08
Improve VNode typings

StreetStrider:
I've noticed that `h` [can accept single VNode](https://github.com/snabbdom/snabbdom/blob/56577183e2352fce720c8a993cdc821afa039633/src/h.ts#L32) in position for `children`. I use this feature from time to time, but if I do this in TS environment, `tsc` nags that this is not correct. This PR fixes typings to allow such invocation. Also, I've created a couple of aliases to make code more readable and self-descriptive.

### DIFF
--- a/src/h.ts
+++ b/src/h.ts
@@ -1,6 +1,6 @@
 import {vnode, VNode, VNodeData} from './vnode';
 export type VNodes = Array<VNode>;
-export type VNodesSparse = Array<VNode | undefined | null>;
+export type VNodesSparse = VNode | Array<VNode | undefined | null>;
 import * as is from './is';
 
 function addNS(data: any, children: VNodes | undefined, sel: string | undefined): void {
--- a/src/h.ts
+++ b/src/h.ts
@@ -1,13 +1,15 @@
 import {vnode, VNode, VNodeData} from './vnode';
+export type VNodes = Array<VNode>;
+export type VNodesSparse = Array<VNode | undefined | null>;
 import * as is from './is';
 
-function addNS(data: any, children: Array<VNode> | undefined, sel: string | undefined): void {
+function addNS(data: any, children: VNodes | undefined, sel: string | undefined): void {
   data.ns = 'http://www.w3.org/2000/svg';
   if (sel !== 'foreignObject' && children !== undefined) {
     for (let i = 0; i < children.length; ++i) {
       let childData = children[i].data;
       if (childData !== undefined) {
-        addNS(childData, (children[i] as VNode).children as Array<VNode>, children[i].sel);
+        addNS(childData, (children[i] as VNode).children as VNodes, children[i].sel);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


staltz:
Looks good to me :)

caridy:
I actually will like us to move into the opposite direction. Eliminating the possibility of passing a VNode in position for children, and be a little bit more restrictive about it in upcoming versions. It is a breaking change, but will help you to eliminate an extra check in our code. @StreetStrider is this something that you do manually? or is this something that is produced by a compiler?

mightyiam:
I would prefer to not be bothered by the possibility of single vs. multiple children when writing my `h` calls.

StreetStrider:
@mightyiam your types would not be violated, check my definitions, they're compatible

mightyiam:
I wasn't clear, sorry. I meant that I prefer the monomorphic approach, like @caridy. Then, I have no choice so I'm not concerned with a choice.

StreetStrider:
@caridy 🤔  I do this by myself in case when node has only child, it makes code form more concise.

I'm already outnumbered here, but I feel responsible for my own opinion, so I put it anyway.

@mightyiam if you're feel afraid of such kind of choices, consider switching to another language (if it is possible). You need more simple and straightforward one. JS is all about choices and alternative ways. It is also very dynamic, in fact current behavior of `h` for children mimics `Array.prototype.concat` and that technic is almost common in cases of heteroheous input of type `T | T[]`. Overloading arguments is also a common. You are free to not use it and develop your own style where you use only comfortable JS subset, which fits to you. Other people can use other approaches. No compability is breaken. That's the sense of flexibility.

@caridy 
> It is a breaking change, but will help you to eliminate an extra check in our code

If you concerned about performance, I would say that in my practice generating vdom is not a bottleneck, rather than *applying* it on dom. The difference may be slight or considerable and grow with structure grow. On small datasets this extra check won't saving anything, putting JIT on table eliminates the rest of the difference. On big sets morphing dom is a perf issue and not vdom itself.
 👍 for striving to maximum performance, but you may be in wrong field too.

paldepind:
Personally, I would prefer to keep the current one-child-shorthand. It think it's quite a bit nicer to write

```ts
h("li", h("a", "Click me"))
```

than

```ts
h("li", [h("a", ["Click me"])])
```

Why would you like to remove this @caridy? It is completely handled in the `h` function so it's not something that add complexity in the core.

caridy:
I'm usually opposed to any sort of condition that will have to happen in the critical path. And in this case, just because it is very nice (which I totally agree), it means that at least one more condition (usually type check) on the creation of the vnodes of the tree to determine if `children` should be wrapped or not. Maybe, what I can do is to create a new pkg or file that only exports something call `fast-h` or something like that, which has a very rigid API for those who really care about speed.

Disclaimer: I already use that fast-h everywhere, and it is lightning fast! :)

paldepind:
I definitely see the advantage of a fast `h` with no bells and whistles. I think having both a convenient and a fast `h` would be great. Both are appropriate in different situations. We could include it either in the main repository or as a separate package.

If we agree to keep the current convenience in `h` then this PR should definitely be merged.

StreetStrider:
You can rid of that functionality in following releases if you decide to, in discussion; but for now (imo) this PR is a fix and should be merged anyways, since it better reflects the current API.

paldepind:
I'm going to merge this PR as we all agree that it represents the way the current API actually is.

StreetStrider:
@paldepind glad to help. Thanks for presense in your initialives, like snabbdom 👍

PR #346, opened 2017-11-28
Fixed missing domApi argument in recursive toVNode() call

katyo:


staltz:
Looks good to me

katyo:
I have no ideas how to test it.

staltz:
Hints:

- There are tests for HTML DOM API: https://github.com/snabbdom/snabbdom/blob/master/test/htmldomapi.js
- There are tests for toVNode: https://github.com/snabbdom/snabbdom/blob/2271b7bdf15577eabd8de961f4e5bba5bd1515fe/test/core.js#L266

How about creating a fake `domApi` object where some functions have asserts inside them, then use toVNode in a way that will trigger the recursive calls.

test/core.js
```diff
@@ -337,6 +337,42 @@ describe('snabbdom', function() {
...
         assert.strictEqual(elm.childNodes[0].textContent, 'Hello');
       })
+      it('can work with domApi', function () {
+        var domApi = {
+            tagName: function(elm) { return elm.tagName.toUpperCase(); },
+            getTextContent: function(elm) { return elm.text; },
+            isElement: function(elm) { return !!elm.tagName; },
+            isText: function(elm) { return !!elm.text; }
+        };
+        var Elm = function(tagName) {
```
caridy: I don't think u need `Elm` constructor, just create a DOM structure like described here: 

https://github.com/snabbdom/snabbdom/blob/master/test/core.js#L268-L273

Then call `toVNode()` on that structure and analyze the returned vnode structure to make sure that it is correct.

you can provide a domApi where tagName returns a custom tagname for any element processed, e.g.:

```js
tagName: function(elm) { return 'X-' + elm.tagName.toUpperCase(); },
```

which means that both, the parent a child vnodes should have a `sel` that contains the `x-` prefix.

katyo:
It seems I amended last commit

staltz:
Thank you @katyo !

PR #347, opened 2017-12-02
Using innerHTML prop to patch non-empty node results in TypeError (null)

tokichie:
I'm using snabbdom through cycle.js.
Recently, I'm stuck on handling server-side-rendered content.
I want to refresh server-side-rendered element with raw html returned from server asynchronously, but it raises a TypeError.

Checking existence of parent can avoid this (see changes), but I couldn't trace why parent is null in this case.

Here is html and js code to reproduce:
```html
<html>
  <head></head>
  <body>
    <div id="container"><h2>Server-side-rendered content</h2></div>
    <script src="./index.js"></script>
  </body>
</html>
```

```js
var snabbdom = require('snabbdom');
var patch = snabbdom.init([
  require('snabbdom/modules/class').default,
  require('snabbdom/modules/props').default,
  require('snabbdom/modules/style').default,
  require('snabbdom/modules/eventlisteners').default,
]);
var h = require('snabbdom/h').default;
var toVNode = require('snabbdom/tovnode').default;
 
var container = document.getElementById('container');
var html= '<h2>asynchronously returned content</h2>';
var vnode = h('div#container', { props: { innerHTML: html } });
patch(toVNode(container), vnode);
```

Here is the error (on Chrome 62.0.3202.94)

### DIFF
--- a/src/snabbdom.ts
+++ b/src/snabbdom.ts
@@ -70,7 +70,9 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
     return function rmCb() {
       if (--listeners === 0) {
         const parent = api.parentNode(childElm);
-        api.removeChild(parent, childElm);
+        if (parent) {
+          api.removeChild(parent, childElm);
+        }
       }
     };
   }
--- a/test/core.js
+++ b/test/core.js
@@ -336,7 +336,25 @@ describe('snabbdom', function() {
         assert.strictEqual(elm.childNodes.length, 1);
         assert.strictEqual(elm.childNodes[0].nodeType, 1);
         assert.strictEqual(elm.childNodes[0].textContent, 'Hello');
-      })
+      });
+      it('can replace non-empty node with innerHTML prop', function() {
+        var h2 = document.createElement('h2');
+        h2.textContent = 'Hello'
+        var prevElm = document.createElement('div');
+        prevElm.id = 'id';
+        prevElm.className = 'class';
+        prevElm.appendChild(h2);
+        var html = '<span>Hi</span>';
+        var nextVNode = h('div#id.class', { props: { innerHTML: html } });
+        elm = patch(toVNode(prevElm), nextVNode).elm;
+        assert.strictEqual(elm, prevElm);
+        assert.equal(elm.tagName, 'DIV');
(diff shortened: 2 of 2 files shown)


caridy:
Ok. Here is the explanation:

* container is in the DOM, and has child elements attached to it (whatever is in the DOM).
* toVNode(container) returns a vnode collection that not only contains the vnode of container, but also children vnodes from childNodes collection.
* the patching mechanism patches the vnode first, then it patches the collection of children.
* when patching container's vnode, and setting innerHTML prop (which I strongly recommend not to do) it removes all childNodes from the DOM but still they are referenced by the children vnode array.
* when attempting to patch each individual child from container's vnode's children, it will attempt to remove the element from the DOM because the new vnode does not have children, and at that point, it fails because they were already removed in a previous call via innerHTML.

That explains that why! now we need to figure out if we want to support innerHTML prop, and how?

tokichie:
Thank you for the explanation of processing flow. 
I understand why parent is null in this case.
Then, I wonder whether this patch makes some bad effects on snabbdom's vdom system or not. (from the explanation, I can't figure out the case that this code destroys snabbdom)

caridy:
I don't think it destroys it, snabbdom patching mechanism is very resilient. My main concern is that by allowing innerHTML as a prop against a vnode that has children, the diffing algo might have other unpredictable consequences that we are not seeing here. This is a new problem introduced by `toVNode()`, which is fairly new. I will like to get more input here from the other maintainers, just to be sure.

caridy:
It turns out that I ran into this exact problem when trying to implement a recovery mechanism (similar to react's error boundaries), where a child vnode throws an error during the life cycle, and the repatching with empty children list fails due to the fact that the offender old vnode has an element that was never inserted into the DOM. So, I'm supportive of this.

cornerman:
I am also in need for this change, because it allows mixing `innerHTML` with a snabbdom `VNode`. We use `innerHTML` to render markdown as html. We use this patch now for a while and did not have any problems. Any more concerns about this PR?

staltz:
Good to merge in my opinion

mightyiam:
@tokichie would you like to rebase?

paldepind:
This has been merged through #1083.

PR #355, opened 2017-12-13
Also create a dist build for the dataset module

jcbrand:
Currently the dataset module doesn't get a distribution build. This pull request adds the necessary gulp step and the built file.


caridy:
I believe we are also missing the `hero` module as well.

Issue #359, opened 2018-01-04
[Proposal] Snabbdom V2


caridy:
# Snabbdom V2

This issue is a placeholder for some discussions about snabbdom v2.

## Goals

1. 20% perf improvement on the most common cases.
1. 30% size reduction of the core library.
1. Provide extension point for custom comparison of vnodes.
1. Provide a way to normalize children collection in modules.
1. Clear separation between helpers and core lib.
1. Formalize hooks and modules workflow for the different type of vnodes.
1. Preserve module's API for backward compatibility.
1. Make shadowDOM and Fragments first class citizens to support web components.
1. Eliminate the ambiguious double exports.

## Other changes

* No more dist/ folder, users of this low-level library, users of it can bundle it at their will.
* Perf infrastructure to measure incremental changes.
* Improve test infrastructure.
* Improve the overall contributing experience.

### 1. Perf Improvements

* Reduce string manipulation and string comparison during the critical path.
* Reduce the amount of branches during the critical path.
* Off-load part of the heavy lifting (vnode normalization) to the vnode creation.
* Reduce type cohertion.

### 2. Size Reduction

* Core lib does not need to know about the DOM or Elements at all.
* Patch API reduced to only deal with normalized vnodes.
* Eliminate unnecessary abstractions.

### 3. Custom Comparison

Using `key` and `sel` as a way to determine whether or not two vnodes are alike can still cover a lot of ground, but in some cases, users might want to apply a more restrictive mechanism to avoid recycling old vnodes. We should provide a way to define such comparison as an extension point.

### 4. Children collection could be normalized by modules

As today, the children reference is stored before invoking `update` or `create` hook, this makes it difficult for modules to do any kind of slotting or normalization of children without mutating the original `children` array, which is not ideal and goes against the principle of snabbdom of never mutating objects that it doesn't own.

A possible solution is to only extract the `children` value of the vnode after all modules are done with the `update` or `create` process.

### 5. Core vs Helpers vs Modules

This project contains the "core lib" whos job is to diff and patch, it also contains a default set of modules that covers most common cases for a DOM Element patching, and a set of helpers to pave the way for users to adopt snabbdom for the Web. This separation is not clear today, and should be very explicit in the next version. The proposed separation is:

* Helpers: `attachto`, `h`, `thunk`, `tovnode`
* Modules: same as today
* Core: `snabbdom` plus the types defined in `hooks` and `vnode`

[...]

PR #362, opened 2018-01-10
[related to issue #359] Reference Implementation for Snabbdom V2

caridy:
[WIP] 

* ~~Provide extension point for custom comparison of vnodes.~~
* Provide a way to normalize children collection in modules.
* Clear separation between helpers and core lib.
* Make shadowDOM and Fragments first class citizens to support web components.
* Eliminate the ambiguous double exports.

## Open Questions

* `src/types.ts` is about types, maybe we should just use the default typescript typing files.
* hook.create vs module's create, who goes first? what about insert?


paldepind:
I haven't had the time to closely look at this yet (I have exams next week). But, I'm interested in why you went with Jest? I have two gripes with Jest. First, it lacks browser testing which I think is sub-optimal for a library that runs in the browser. Secondly, I've had disappointing performance experiences with it. I evaluated it in a project of mine and found that it was 3 times as slow as Mocha (see [here](https://github.com/kulshekhar/ts-jest/issues/280)).

On the other hand, I've had a great experience with [karma-typescript](https://github.com/monounity/karma-typescript). It's easy to set up with TypeScript (like Jest), supports Mocha tests and can run tests both in NodeJS and browsers. It allows for setting up CI with Sauce Labs that runs tests in all major browsers which is really nice for ensuring browser support.

I'm not at all opposed to using Jest, though. But maybe karma-typescript is a better fit?

mightyiam:
+1 for AVA testing framework. Has TypeScript types. I dunno about running in the browser, though. It runs the test modules in parallel!

diervo:
@paldepind The reason we went with Jest is mostly because we been using it extensively and works really well for Unit testing: Is fast, has coverage built in, you can transform typescript/babel if you so desire, moreover we have found very little issues when dealing with the JSDom implementation.

However, as you point out this is not enough. For real functional and perf testing we have a separated runner and infrastructure to guarantee no regressions and functional correctness. Thats the part we haven't yet integrated. We are on it

pmdartus:
Even if I really like the way jest works, I like the idea of using Karma as @paldepind mentioned above. What about using a mix of:
* Jest for fast and reliable unit tests
* and Karma for integration test against the supported browsers

caridy:
Oh! don't worry much about that part of this PR, I was in Cuba without internet when working on this branch, and the mocha setup does not work offline (loads mocha lib over the wire), so I copied the jest set up from other project to run the exact same tests and get some code coverage report.

src/helpers/thunk.ts
```diff
@@ -1,5 +1,5 @@
-import {VNode, VNodeData} from './vnode';
-import {h} from './h';
+import {VNode, VNodeData} from '../types';
+import h from './h';
```
staltz: What's the reasoning behind default exported `h`? I've been hearing complaints about them and a trend to use only named exports.

src/snabbdom.ts
```diff
@@ -229,20 +249,23 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
           }
           newStartVnode = newCh[++newStartIdx];
         }
       }
     }
     if (oldStartIdx <= oldEndIdx || newStartIdx <= newEndIdx) {
       if (oldStartIdx > oldEndIdx) {
-        before = newCh[newEndIdx+1] == null ? null : newCh[newEndIdx+1].elm;
+        const n = newCh[newEndIdx+1];
+        before = isVNode(n) ? n.elm : null;
```
staltz: 👍 to this lookup optimization

src/helpers/thunk.ts
caridy: I'm fine either way, we just need to pick one instead of exporting both!

As one of the champions of ES6 modules, the idea was to always favor default exports when possible, but one thing is what TC39 believes, and a very different thing is what folks actually do :)

src/helpers/h.ts
```diff
@@ -0,0 +1,153 @@
+import {VNode, VNodes, VNodeData, VElement, VText, VComment, VFragment} from '../types';
+export type VNodesSparse = VNode | VNodes;
+
+function isVElement(vnode: VNode): vnode is VElement {
+  return vnode.nt === ELEMENT_NODE;
+}
+
+function addNS(vnode: VElement) {
+  const { data, children, sel } = vnode;
+  // TODO: review why `sel` equal `foreignObject` should get this `ns`
```
caridy: I could not understand the "why" here, maybe someone can explain it!

src/helpers/htmldomapi.ts
```diff
@@ -0,0 +1,61 @@
+import {DOMAPI} from '../types';
```
caridy: @staltz maybe you can provide some feedback here. The git diff is not that nice since it didn't detect the move operation, but in short, the changes in this file are around the fact that `toVNode()` only operates in DOM structures, and thats why I have removed few of the old APIs that were only used by `toVNode()` from this file.

Do you foresee `helpers/tovnode` operating on non-DOM trees? IMO if you are using a non-DOM tree, you can implement your own tovnode routine to build such tree.

src/helpers/tovnode.ts
```diff
@@ -0,0 +1,92 @@
...
+
+function isText(node: Node): node is Text {
+  return node.nodeType === 3;
+}
+
+function isComment(node: Node): node is Comment {
+  return node.nodeType === 8;
+}
+
+function toVNode(node: Node): VNode {
```
caridy: this refactor does not require the domapi anymore as a second argument since it only operates on the html DOM, and it knows what to do as described in my previous comment.

src/helpers/tovnode.ts
```diff
@@ -0,0 +1,92 @@
...
+  } else if (isText(node)) {
+    text = getTextContent(node) as string;
+    vnode = vnodeText(text == null ? '' : text);
+    vnode.elm = node;
+  } else if (isComment(node)) {
+    text = getTextContent(node) as string;
+    vnode = vnodeComment(text);
+    vnode.elm = node;
+  } else if (isFragment(node)) {
+    vnode = vnodeFragment({}, []);
```
caridy: fragment was introduced to be able to support a shadow root as the Node.

src/helpers/tovnode.ts
```diff
@@ -0,0 +1,92 @@
...
+    vnode.elm = node;
+    extractChildren(vnode);
+  } else {
+    throw new TypeError();
```
caridy: this is also new in v2, there are few places where a type error or reference error will be issue instead of trying to generalizing the operation.

src/hooks.ts
```diff
@@ -1,25 +0,0 @@
-import {VNode} from './vnode';
```
caridy: moved to types

src/htmldomapi.ts
```diff
@@ -1,97 +0,0 @@
-export interface DOMAPI {
```
caridy: moved to helpers/htmldomapi

src/is.ts
```diff
@@ -1,4 +0,0 @@
-export const array = Array.isArray;
-export function primitive(s: any): s is (string | number) {
```
caridy: only used by `h`, so, moved there.

src/modules/module.ts
```diff
@@ -1,10 +1 @@
-import {PreHook, CreateHook, UpdateHook, DestroyHook, RemoveHook, PostHook} from '../hooks';
```
caridy: moved to types

src/snabbdom.bundle.ts
```diff
@@ -1,16 +0,0 @@
-import {init} from './snabbdom';
```
caridy: dropped entirely. in v2 we only give you the low-level stuff, something like this can be a different package with some opinionated diffing configuration.

src/tovnode.ts
```diff
@@ -1,39 +0,0 @@
-import vnode, {VNode} from './vnode';
```
caridy: moved to helpers/tovnode

src/types.ts
```diff
@@ -0,0 +1,111 @@
...
+import {Props} from './modules/props'
+import {Dataset} from './modules/dataset'
+import {Hero} from './modules/hero'
+
+export type Key = string | number;
+
+export type VNodes = Array<VNode | null>;
+
+export interface VNode {
+  nt: number; // matches DOM Spec for Node's nodeType
```
caridy: this is one of the big changes. `nt` stands for `nodeType`, which is a numeric value matching the DOM spec to speed up the process of diffing by not having to compare the `sel` string values anymore.

additionally, developers can specify their own comparison mechanism if they have specific constrains around the identity of ever vnode.

src/types.ts
```diff
@@ -0,0 +1,111 @@
...
+  sel: string | undefined;
+  data: VNodeData;
+  children: VNodes | undefined;
+  elm: Node | undefined;
+  text: string | undefined;
+  key: Key | undefined;
+}
+
+export interface VElement extends VNode {
```
caridy: this is just for convenience, and helps a lot to hint at the type system what kind of vnode is the code dealing with.

src/vnode.ts
```diff
@@ -1,49 +0,0 @@
...
-  attachData?: AttachData;
-  hook?: Hooks;
-  key?: Key;
-  ns?: string; // for SVGs
-  fn?: () => VNode; // for thunks
-  args?: Array<any>; // for thunks
-  [key: string]: any; // for any other 3rd party module
-}
-
-export function vnode(sel: string | undefined,
```
caridy: this functionality is dropped entirely, instead, you either use `h`, or `toVNode` or you build your own vnode object manually, not need for such simple abstraction.

src/vnode.ts
```diff
@@ -1,49 +0,0 @@
...
-import {AttachData} from './helpers/attachto'
-import {VNodeStyle} from './modules/style'
-import {On} from './modules/eventlisteners'
-import {Attrs} from './modules/attributes'
-import {Classes} from './modules/class'
-import {Props} from './modules/props'
-import {Dataset} from './modules/dataset'
-import {Hero} from './modules/hero'
-
-export type Key = string | number;
```
caridy: all these types were moved to `types`.

src/snabbdom.ts
```diff
@@ -1,55 +1,90 @@
...
+  (oldVnode: VNode, vnode: VNode): VNode;
+  children(parentElm: Node, oldCh: VNodes, newCh: VNodes): VNodes;
+}
+
+function defaultCompareFn(vnode1: VNode, vnode2: VNode): boolean {
+  return vnode1.nt === vnode2.nt && vnode1.key === vnode2.key && vnode1.sel === vnode2.sel;
+}
+
+function isVNode(vnode: any): vnode is VNode {
+  return vnode != null && vnode !== undefined;
```
blikblum: If vnode is undefined, the second check (`vnode !== undefined`) will not happen making it redundant. The loose equality in first check will do implicit coercion between null and undefined

src/snabbdom.ts
```diff
@@ -1,55 +1,90 @@
...
-}
+function isUndef(s: any): s is undefined { return s === undefined; }
+function isDef(s: any): boolean { return s !== undefined; }
 
-function isVnode(vnode: any): vnode is VNode {
-  return vnode.sel !== undefined;
-}
+type VNodeQueue = VNode[];
+
+const emptyNode: VNode = {
```
caridy: empty node is now special since it uses `nt` 0, so folks can identify it. I was also thinking that maybe we should freeze it to avoid creating a side channel between different kind of vnodes.

src/snabbdom.ts
```diff
@@ -1,55 +1,90 @@
...
   [K in keyof T]: (T[K])[];
 }
 
-type ModuleHooks = ArraysOf<Module>;
+export interface PatchFunction {
+  (oldVnode: VNode, vnode: VNode): VNode;
+  children(parentElm: Node, oldCh: VNodes, newCh: VNodes): VNodes;
+}
+
+function defaultCompareFn(vnode1: VNode, vnode2: VNode): boolean {
```
caridy: this is the default behavior that determines whether or not two nodes are alike. a new configuration option for `init()` allows developers to specify their own comparing function. In theory, if you don't use `sel` at all, you can drop the string comparison from this function, and get away with a more performant comparison mechanism.

src/snabbdom.ts
```diff
@@ -1,55 +1,90 @@
...
+}
 
-export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
-  let i: number, j: number, cbs = ({} as ModuleHooks);
+type ModuleHooks = ArraysOf<LifeCycleHooks>;
 
-  const api: DOMAPI = domApi !== undefined ? domApi : htmlDomApi;
+const hooks: (keyof LifeCycleHooks)[] = ['create', 'update', 'remove', 'destroy', 'pre', 'post'];
 
+export function init(modules: Array<Partial<Module>>, api: DOMAPI,
```
caridy: the signature of the `init` has changed a little bit, it now takes 2 required arguments and one optional:

* modules (same as before)
* api (same as before, but it is now required)
* compareFn (optional argument that fallbacks to `defaultCompareFn` from above)

the reasoning behind the required second argument is that if you're bundling the snabbdom core library, it should work as a generic diffing library with no affiliations to the DOM at all, and the interaction with the real tree should be proxied by the api provided as a second argument as an indirection.

src/snabbdom.ts
```diff
@@ -1,55 +1,90 @@
...
+};
 
 type KeyToIndexMap = {[key: string]: number};
 
 type ArraysOf<T> = {
   [K in keyof T]: (T[K])[];
 }
 
-type ModuleHooks = ArraysOf<Module>;
+export interface PatchFunction {
```
caridy: the patch function returned by `init` is now more restrictive, it only accepts vnodes since snabbdom doesn't really know about the DOM anymore. By default, you can rely on `toVNode()` as the first argument. This removed some weird behavior that result in not spidering the whole structure of the provided node, which eventually let old DOM elements hanging there after the patch.

additionally, a `children` member is introduced to only patch the children collection without touching the parent node at all. This is mostly to support shadow root, and fragments, but I'm ok with removing this if you feel this is too much.

src/snabbdom.ts
```diff
@@ -60,125 +95,111 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
+      if (isDef(i = data.hook) && isDef(i.insert)) insertedVnodeQueue.push(vnode);
+    } else if (isTextVNode(vnode)) {
+      vnode.elm = api.createTextNode(vnode.text);
+    } else if (isCommentVNode(vnode)) {
+      vnode.elm = api.createComment(vnode.text);
+    } else if (isFragmentVNode(vnode)) {
+      vnode.elm = api.createFragment();
     } else {
-      vnode.elm = api.createTextNode(vnode.text as string);
+      throw new TypeError();
```
caridy: as I mentioned above, no more fallback mechanism, it is a error if you provide a weird constructed vnode.

src/snabbdom.ts
```diff
@@ -60,125 +95,111 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
+                        vnodes: VNodes,
                         startIdx: number,
                         endIdx: number): void {
     for (; startIdx <= endIdx; ++startIdx) {
       let i: any, listeners: number, rm: () => void, ch = vnodes[startIdx];
-      if (ch != null) {
-        if (isDef(ch.sel)) {
+      // text nodes do not have logic associated to them
+      if (isVNode(ch)) {
+        if (!isTextVNode(ch)) {
```
caridy: this still requires more thoughts in my opinion, why are text nodes special? what about comments, fragments, etc? should they also be skipped the destroy routine?

src/snabbdom.ts
```diff
@@ -252,45 +275,45 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
-    let oldCh = oldVnode.children;
-    let ch = vnode.children;
     if (oldVnode === vnode) return;
     if (vnode.data !== undefined) {
       for (i = 0; i < cbs.update.length; ++i) cbs.update[i](oldVnode, vnode);
-      i = vnode.data.hook;
-      if (isDef(i) && isDef(i = i.update)) i(oldVnode, vnode);
+      if (isDef(hook) && isDef(i = hook.update)) i(oldVnode, vnode);
     }
+    let oldCh = oldVnode.children;
```
caridy: these is another important change. `children` for old and new vnodes are extracted from the vnode after `update` patch (same for create), which means that a module can attempt to normalize the children collection. This is needed to emulate slotting of vnodes, etc. to match the shadow DOM slotting process.

It was doable before but requires developers to mutate the original children array, which wasn't very elegant, and an anti-pattern IMO.

src/snabbdom.ts
```diff
@@ -60,125 +95,111 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
-      const tag = hashIdx !== -1 || dotIdx !== -1 ? sel.slice(0, Math.min(hash, dot)) : sel;
-      const elm = vnode.elm = isDef(data) && isDef(i = (data as VNodeData).ns) ? api.createElementNS(i, tag)
-                                                                               : api.createElement(tag);
-      if (hash < dot) elm.setAttribute('id', sel.slice(hash + 1, dot));
-      if (dotIdx > 0) elm.setAttribute('class', sel.slice(dot + 1).replace(/\./g, ' '));
+    if (isElementVNode(vnode)) {
+      const {data, tag} = vnode;
+      const elm = vnode.elm = isDef(i = data.ns) ? api.createElementNS(i, tag)
+                                                 : api.createElement(tag);
+      if (isDef(i = data.hook) && isDef(i.create)) i.create(emptyNode, vnode);
```
caridy: this is another important change that requires more debate. In the previous iteration, the `data.hook.create` happens after the children is processed, while the `module.create` happens before. In this refactor, they are unified, and are happening at the same time. I was confused on the "why", could not really explain it.

Additionally, the more I understand the code, the more I think that `data.hook.*` are just redundant, and that you could create a module that emulates those hooks instead of having snabbdom to take care of it. Maybe that's a better option, to implement a optional module called hooks, that provides the exact same semantics that we have today, and you can use it if you really want to explore that avenue. What do you think?

src/snabbdom.ts
caridy: ah! good catch! will fix it.

src/snabbdom.ts
```diff
@@ -1,55 +1,90 @@
...
+function isCommentVNode(vnode: VNode): vnode is VComment {
+  return vnode.nt === COMMENT_NODE;
+}
+
+function createKeyToOldIdx(children: VNodes, beginIdx: number, endIdx: number): KeyToIndexMap {
   let i: number, map: KeyToIndexMap = {}, key: Key | undefined, ch;
   for (i = beginIdx; i <= endIdx; ++i) {
     ch = children[i];
-    if (ch != null) {
+    if (isVNode(ch)) {
```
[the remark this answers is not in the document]
caridy: few reasons:

* the intend of the check is clear (we want to know if it is a vnode or not)
* simple functions like that will be inlined anyway by the JIT, so perf will not be impacted.
* the minified version of the condition is still 4 characters, same as before.

src/helpers/htmldomapi.ts
staltz: That's a good point, when I wrote `toVNode` I introduced the `domApi` argument just because `toVNode` happened to do similar kinds of lookups that DOMAPI objects had, but I didn't argue for some use case where the domApi is something else than the htmldomapi.

So I'm okay with these changes.

On a bigger picture perspective, `DOMAPI` serves as an abstraction on how to **write** the DOM. `toVNode` needed an abstraction on how to **read** the DOM. In the typical use case for snabbdom, reading the DOM tree is not necessary since the VDOM is used for reading instead. So the VDOM is the source of truth. We could ask ourselves overall what are those cases where we need to read specifically from the DOM and not from the VDOM. If there are enough use cases, then it could justify an abstraction layer for reading the DOM, and perhaps reintroducing that back into `toVNode`.

src/snabbdom.ts
blikblum: You may want also to swap the check order to take advantage of short circuit evaluation, given that undefined values will be more frequent than null

src/snabbdom.ts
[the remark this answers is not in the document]
blikblum: > the intend of the check is clear (we want to know if it is a vnode or not)

In fact is checking if the value is defined or not. IMO `isDefined` or `isDef` would match closer the function role

> simple functions like that will be inlined anyway by the JIT, so perf will not be impacted.

This assumes the code will be compiled by JIT that is not guaranteed although very probably in this case.

Also, there's a tradeoff of inline functions that makes the code bigger (not the JS code, the processor instructions) and can affect the ability to fit the code in the CPU cache

Anyway, any performance gain / loose can be only checked by measuring it, and i would bet that in this case the difference is none. Is more a stylistic thing

src/snabbdom.ts
blikblum: > This is needed to emulate slotting of vnodes, etc. to match the shadow DOM slotting process.

I implemented something like that ( prototype at https://codepen.io/blikblum/pen/BJrqyM ,  package coming soon)

I  implemented as a high order function (withSlot)

Could not figure how to implement as a module. Curious if/how these changes will allow a different implementation

src/helpers/htmldomapi.ts
caridy: excellent! thanks.

src/snabbdom.ts
caridy: Hmm, I don't think that's the case. In the places where we call `isVNode()`, usually we encounter a real vnode, or null, undefined is probably less common since `h` normalizes `undefined` to null in children collection: https://github.com/snabbdom/snabbdom/pull/362/files#diff-8fa4f0f1ff07b4217fd7b7626c8cf01eR117

src/snabbdom.ts
caridy: oh, in the past, we have done this by doing this in a module:

```js
   const slottedChildren = slotMyChildren(vnode.children, vnode.data.withSlot);
   vnode.children.length = 0;
   vnode.children.push.apply(vnode.children, slottedChildren);
```

which is nasty because you're mutating the original `children` provided in the vnode, which could have undesired side effects if memoized at any level in your app.

With the new code in this branch, you could improve that by doing:

```js
   vnode.children = slotMyChildren(vnode.children, vnode.data.withSlot);
```

sill you have the memoization process, but only if you cache the result of calling `h`.

blikblum: > oh, in the past, we have done this by doing this in a module

What hook did you used? `create`?

According to docs `create` is called after DOM element creation so it could lead to issues. That's why i did not tried it.

But reading your comment in this PR, i realized that is called before element creation so is possible to use `create` for such task

> In the previous iteration, the data.hook.create happens after the children is processed, while the > module.create happens before. In this refactor, they are unified, and are happening at the same > time. I was confused on the "why", could not really explain it.

Maybe to allow mutating vnode before element creation, as above?

caridy: oh, no, this operation is not tied to `create` only, it must be done in `update` as well because the children collection might change for an already created vnode/elm, in which case you will have apply the slotting mechanism again.

src/helpers/thunk.ts
paldepind: I agree that only exporting one is best. Personally, I prefer to avoid default exports but it's not a big deal to me.

src/snabbdom.ts
```diff
@@ -1,55 +1,90 @@
...
-function isDef(s: any): boolean { return s !== undefined; }
+/* global Node */
+import {Hooks, InsertHook, VNode, VNodes, Key, VText, VComment,
+        Module, VElement, VFragment, DOMAPI, PreHook, CreateHook, UpdateHook, DestroyHook, RemoveHook, PostHook} from './types';
 
-type VNodeQueue = Array<VNode>;
+const { isArray } = Array;
 
-const emptyNode = vnode('', {}, [], undefined, undefined);
+const ELEMENT_NODE = 1, TEXT_NODE = 3, COMMENT_NODE = 8, DOCUMENT_FRAGMENT_NODE = 11;
```
paldepind: I think introducing the `nt` property on nodes is a great idea. But instead of these constants we could use a const enum:

```ts
const enum VNodeType { Empty, Element, Text, Comment, Fragment };
```

Const enums add no run-time overhead (they're compiled to number literals) but they provide extra type-safety and look nicer.

If using the same numbers as the DOM is important then that can also be done with const enums.

```ts
const enum VNodeType {
  Empty = 0,
  Element = 1,
  Text = 3,
  ...
}
```

caridy: will do!

src/helpers/h.ts
```diff
@@ -0,0 +1,153 @@
...
+  } else if (children != null) {
+    children = [children];
+  }
+  // normalize data
+  data = data || {};
+  if (sel === '') {
+    return vnodeFragment(data, children, text);
+  }
+  // parse selector
+  const hashIdx = sel.indexOf('#');
```
paldepind: I really like how you've moved the selector parsing into `h` while still making it possible to use id and static classes in the selector to guide the DOM diffing.

paldepind:
I'd like to understand why you need a custom function to compare virtual nodes? Depending on what you'll be using it for it may or may not introduce a problem.

As you may know, there are two parts to the children reconciliation algorithm. The [first part](https://github.com/caridy/snabbdom/blob/6f2a84ed9eb3466d6fd024a78016bbddacb9efdb/src/snabbdom.ts#L216-L241) consists of heuristics that handle a lot of common cases in a very fast way. The [second part](https://github.com/caridy/snabbdom/blob/6f2a84ed9eb3466d6fd024a78016bbddacb9efdb/src/snabbdom.ts#L243-L262) is a "fallback", more general algorithm that can handle all cases at the expense of having to build a map. Only the first part uses the `sameVnode` function. In the second part, new and old vnodes are connected by inserting the old vnodes into a map from their key to their index. This means that only the `key` is used to figure out if two vnodes are the same.

That may a problem? It means that if a user supplies a custom comparison function that comparison function will only be used in the cases that can be handled by the simpler algorithm.

src/snabbdom.ts
```diff
@@ -60,125 +95,111 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
-        }
+    if (isDef(i = data.hook) && isDef(i = i.destroy)) i(vnode);
+    for (i = 0; i < cbs.destroy.length; ++i) cbs.destroy[i](vnode);
+    const { children } = vnode;
+    if (isUndef(children)) {
+      return;
+    }
+    for (j = 0; j < children.length; ++j) {
+      i = children[j];
+      if (!isTextVNode(i)) {
```
caridy: @davidturissini found an issue here when trying to test this branch, here the children might be `null`, and `isTextVNode()` expects a valid vnode. this issue might be present in other places where we reuse local variables with type `any`, which in this case is the `i`. I will fix this.

paldepind:
@caridy Did you get a chance to consider my concerns with regards to the custom comparison function as I mentioned above? I think I'd be awesome to get this PR merged but I'm still not sure if the custom comparison is a good idea.

caridy:
Yes! we have discussed (@diervo and I), I will get back to you this week.

caridy:
@paldepind we gather more perf numbers today about the `sameVnode()` when comparing `nt`, `key` and `sel`, and we encounter that in systems where the key is always present (or almost always present), like our case, the extra comparison of the `sel` doesn't really add a significant overhead, literally we could not pinch it, so, I'm withdrawing that part of the proposal.

Now, about the reason, maybe to keep this as historical information, we were after something similar to Ember's Glimmer, trying to get the compiler to add a key to every element in a template tag, as a way to always preserve the state of the elements, effectively reduce the recycling of elements implemented by snabbdom for certain scenarios. If that was the case, then comparing the string `sel` after matching the `key`, was always going to be `true`, that's where we were looking for some extra optimization. It turns out, that engines are really smart, and where string comparisons of the exact same string seems to be wired. (@diervo will confirm this with implementers).

caridy:
@paldepind I think I have addressed all the comments. I do have all the tests (except one) running locally but I'm not committing them to avoid the noise in the diffing for the review. The failure is on:

> snabbdom hooks element hooks calls `create` listener before inserted into parent but after children

which is expected due to the open question above. we were debating about specifying hooks at the vnode level (`vnode.data.hook`), and we came to the conclusion that this feature could probably be removed entirely from snabbdom and just have a standard module to implement it for those folks who really want that. This removes the ambiguity described above (which one runs first, the global or the local?), and also the ambiguity of when are those hooks suppose to be invoked.

in some of my projects I do use the hooks, but I believe no many folks do, so I'm very supportive of moving that out into a module. What is your take on this?

JSteunou:
Any news or plans for the v2 ? I'm very curious about it.

caridy:
@JSteunou yes, a little bit stale at the moment, and I'm about to go on vacation for a month, but before I go, I will submit the tests and everything else that I have locally. At the moment, I'm waiting for @paldepind to discuss my comment (https://github.com/snabbdom/snabbdom/pull/362#issuecomment-368689171) about moving out the code for vnode's hook into its own module, for maximum flexibility. If we settle on that, I can try to push the update this week, and we can start planning for the release later this month or next month.

P.S.: We have been using this version (this branch) in a bucket for a while without major problems.

JSteunou:
Thanks for the update

> P.S.: We have been using this version (this branch) in a bucket for a while without major problems.

This is good to know :)

JSteunou:
@caridy did you try `npm run compile` ? There is at least one error

src/snabbdom.ts
```diff
@@ -252,45 +275,45 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
-  return function patch(oldVnode: VNode | Element, vnode: VNode): VNode {
-    let i: number, elm: Node, parent: Node;
-    const insertedVnodeQueue: VNodeQueue = [];
-    for (i = 0; i < cbs.pre.length; ++i) cbs.pre[i]();
-
-    if (!isVnode(oldVnode)) {
-      oldVnode = emptyNodeAt(oldVnode);
+  const patch = function patch(oldVnode, vnode) {
+    if (!isVNode(oldVnode) || !isVNode(vnode)) {
+      throw new TypeError();
```
JSteunou: I have 2 issues: 

* `isVnode` returns true against a DOM Element (`document.createElement('div')`) which make sense when you look at the test inside the function...
* Why remove the support of `Element` ? I think it wouldnt cost that much to include the call to `toVnode` inside this lib if the oldVnode is an element instead of introduce a breaking change a make everyone duplicate & include this piece of code

src/snabbdom.ts
```diff
@@ -1,55 +1,88 @@
...
-function isVnode(vnode: any): vnode is VNode {
-  return vnode.sel !== undefined;
+function isVNode(vnode: any): vnode is VNode {
+  return vnode != null;
 }
 
-type KeyToIndexMap = {[key: string]: number};
+function isElementVNode(vnode: VNode): vnode is VElement {
+  return vnode.nt === VNodeType.Element;
+}
```
JSteunou: This provoke a regression when tested against current code base because my initial vnode dont have `nt` property. Always `undefined`. When this should be set ?

caridy: Yes, this is a breaking change if you are building your own vnodes (you will need to change the shape of those vnodes to include this new property), but if you use our `h()` or `toVNode()`, we will take care of it.

src/snabbdom.ts
caridy: yeah, that part was discussed on the initial proposal, snabbdom doesn't know about elements anymore, it knows about vnodes, this can be done in userland code before calling to patch, and we offer a utility function for it.

src/snabbdom.ts
JSteunou: I'm using snabbdom-jsx so that's why :)

caridy: we will need to have a new version for that one, once this is ready. feel free to branch out snabbdom-jsx, to adjust the produced shaped.

kuraga:
About testing: What about [tman](https://github.com/thunks/tman)?

kuraga:
@caridy , thanks for your work!

Could you please review this PR, **please**, in case there is an intention to finish this case by us?

caridy:
@kuraga

> Could you please review this PR, _please_, in case there is an intention to finish this case by us?

This is a very old PR, nobody else is working on it during this time. I will close it.

kuraga:
"Nice" tradition to close after somebody who wants to continue asked...

caridy:
@kuraga this is labeled as a "reference implementation". It is perfectly fine to pull from this branch, and continue working on such until it can be considered production-level, and get more eye balls on it. Go for it!

jvanbruegge:
Yes, also this is how open source development works. The original author closed this PR because they have no interest in reviving it. The code however is still here. You can still open new PRs for stuff based on this code. Just a note, please do not open one giant PR, make smaller ones, one per feature/bug fix you want to add.

At the moment my time is very limited, but in May I will have time to look over contributions

kuraga:
Main word was "to close". Just ideas from closed PRs have a minimal chance to be found...

kuraga:
@caridy , just one question, please.

`Provide a way to normalize children collection in modules.` Show me some lines of this change in the PR, please? Just want to get bearings...

caridy:
@kuraga 

> Provide a way to normalize children collection in modules

This idea was something that we discussed back in 2018, basically, in a module, you compare two vnodes (old and new), but the `children` of the new vnode is changed (normalized) by a module, the diffing of the children -- which is going to happen right after -- is using the old references. (https://github.com/snabbdom/snabbdom/blob/dca0b0c9af5ff514b26ca684c2c75b11733c2418/src/init.ts#L383-L384). This forces modules to instead of just replacing vnode.children collection with a new collection after normalization, to actually alter the array that was provided, which is problematic because snabbdom doesn't own that array.

A good use-case here is flatting children. Some frameworks can produce arrays of arrays for children, which might need to be flattened. Their only choice is to flatten the array because calling snabbdom, which is fine, but it will be nice if snabbdom can take care of that via a module that normalize the `.children` property of any new vnode the first time it sees it.

Issue #361, opened 2018-01-09
Type Error in h()?


gfmio:
Hi!

I believe there is a typing error / the typing could be improved for `h()`. From `h.d.ts`:

```ts
import { VNode, VNodeData } from './vnode';
export declare type VNodes = Array<VNode>;
export declare type VNodesSparse = VNode | Array<VNode | undefined | null>;
export declare function h(sel: string): VNode;
export declare function h(sel: string, data: VNodeData): VNode;
export declare function h(sel: string, text: string): VNode;
export declare function h(sel: string, children: VNodesSparse): VNode;
export declare function h(sel: string, data: VNodeData, text: string): VNode;
export declare function h(sel: string, data: VNodeData, children: VNodesSparse): VNode;
export default h;
```

The final parameter `children` can not just be `VNode` or `Array<VNode | undefined | null>`, but also `string` or `number` and each of the array elements can also be `string` or `number`.

I ran into this as I was writing a wrapper around snabbdom's `h` for my own render function in the framework I'm developing. I would suggest to update the typing to:

```ts
export declare type VNodes = Array<VNode>;
export declare type VNodeChildElement = VNode | string | number | undefined | null;
export declare type ArrayOrElement<T> = T | T[];
export declare type VNodeChildren = ArrayOrElement<VNodeChildElement>

export declare function h(sel: string): VNode;
export declare function h(sel: string, data: VNodeData): VNode;
export declare function h(sel: string, children: VNodeChildren): VNode;
export declare function h(sel: string, data: VNodeData, children: VNodeChildren): VNode;
export default h;
```
Thoughts? Shall I create a pull request?

paldepind:
I think that sounds like a good idea 👍  The code you proposed both reflects the API better and is a nice refactor that simplifies things. I think a PR would be appreciated.

Do note though that the types should _not_ be changed in `h.d.ts`. They have their origin in `src/h.ts`.

greydrizzle:
Hello. I ran into this problem also. Could this be closed and included in a new release? I'm using typescript, so to work around it I have been wrapping text children (mixed with other children) in a span.

I'm having a good time using snabbdom, thanks for your work on this!

mightyiam:
Closed in #364.

PR #364, opened 2018-02-04
Fix for issue 361

gfmio:
Sorry for the delay in creating this small fix. Here's the PR for #361.


paldepind:
Thank you @gfmio. This is a good generalization of the types 👍

Issue #367, opened 2018-03-04
style module: `transitionend` doesn't fire when no transition is needed


whitecolor:
Whle developing in chrome on Windows I've noticed and debugged some heisenbug with style module removal: when there are multiple extensive transitions taking place `transitionend` sometimes doesn't fire, and the element is not removed. 

The case with concrete prop: say opacity should change (from 1 t 0). If bug happens element stays transparent (actual opacity is 0), but not removed from DOM.

The reason for this bug is that `opacity` in computed style `compStyle[opacity]` is actually already equals `0` (target transition value), and the browser doesn't perform transtion for this prop.

So it seems quite  easy to workaround/fix by adding check [here](https://github.com/snabbdom/snabbdom/blob/master/src/modules/style.ts#L78):
```ts
 if(applied.indexOf(props[i]) !== -1 && elm.style[props[i]] !== compStyle[props[i]]) amount++;
```
and `if (amount === 0)` call `rm()` without waiting for `transitionend`

caridy:
I will suggest to send an actual PR so it can be reviewed and evaluated.

whitecolor:
I've added a PR. https://github.com/snabbdom/snabbdom/pull/368

Another solution would probably be putting listener for `transitionend` before computing style, though I don't think that it would the better option.

jvanbruegge:
I also stumbled across this when debugging https://github.com/cyclejs/cyclejs/issues/747. I am not sure why the `transitionend` event does not fire in Chrome. Another solution would be to not use the event, but a timeout instead:
```js
setTimeout(() => {
        --amount;
        if (amount === 0)
            rm();
    }, parseFloat(compStyle['transition-duration'].slice(0, -1)));
```

whitecolor:
@jvanbruegge 

`setTimeout` is not appropriate in many edge cases, I think this is quite approved in particular with cycle)

Did you try my PR for your case? Will it work for you? It works well for me. And all it does just is fixes a little bit logic. But in my case it does not fire because there is no actual change of "transitioned" property value.

I suggest it just merged, to fix the bug for this edge case.

cc @caridy

jvanbruegge:
I tried, but your PR does not fix the issue.

jvanbruegge:
I finally found a solution for my bug, but I have no idea *why* it fixes the bug:
```diff
function applyRemoveStyle(vnode, rm) {
    var s = vnode.data.style;
    if (!s || !s.remove) {
        rm();
        return;
[...]

PR #368, opened 2018-03-05
Style module: handle case if no transition is needed

whitecolor:
Fixes https://github.com/snabbdom/snabbdom/issues/367
Checks if actual (computed) css prop value already equals target remove value.


whitecolor:
Another solution would probably be putting listener for transitionend before computing style, though I don't think that it would the better option.

PR #369, opened 2018-03-13
Refactor to remove function casting.

tobli:
Casting the setAttribute and setAttributeNS functions at each call site make the code harder to read. Instead extract the new overloads for Element into a global declare statement.

This does not change the generated javascript code,  just makes the typescript easier to read.


staltz:
Looks good to merge

paldepind:
Thank you @tobli :+1:

PR #371, opened 2018-03-18
Selectors under symbols

lukaszmakuch:
# What
This is a change of the VDom interface.
Instead of this:
```ts
interface VNode {
  sel: string | undefined;
  //... 
}
``` 
I propose this:
```ts
export interface VNode {
  [SELECTOR_KEY]: string | undefined;
  //...
}
```
Where _SELECTOS_KEY_ is:
```ts
const SELECTOR_KEY:unique symbol = Symbol.for('snabbdom.selector');
```
# Why
To reduce the risk of an XSS attack. Currently a valid VDOM node may come from some JSON payload, HTML form or MongoDB.

Let's assume we have a VDOM node like this:
```ts
h('div', {}, username)
```
It's safe as long as _username_ is a string. But as soon as the attacker manages to pass the following JSON object as _username_, we have a problem:
```json
{"sel":"script","data":{},"text":"alert(1)"}
```
# Advantages
- one less way to perform an XSS attack
# Disadvantages
This change requires _Symbol.for_ support. For applications targeting older browsers, it will be necessary to include a polyfill (which may not provide safety) to prevent the app from crashing. 

# Things I've done
- [x] I made the tests pass under Chrome and Safari. 
- [x] I rejected the idea of using Infinity to prevent JSON-driven XSS, because Infinity may be stored in MongoDB

# Things I haven't done
- [ ] Testing it with a Symbol polyfill under IE
- [ ] Although I feel more like using a polyfill for older browsers, because the developer may then decide whether to support old browsers, I haven't fully rejected the idea of using some type of runtime fallback mechanism in case of missing _Symbol_ support. Users of old browsers won't be safe anyway, it will probably make stuff slightly slower, but it won't crash even without a polyfill. As far as I understand, React does it this way.

# Ideas
- If we decide this change makes sense and we want to merge it, we should mention in the README that a Symbol polyfill may be required.


paldepind:
Thank you for creating this PR @lukaszmakuch. This is definitely something that we should find a solution to.

I'm curious if this is a problem you discovered in actual usage or if you realized the problem by looking at the way VNodes are currently represented? In case you ran into the problem in actual usage it would be interesting to hear about that.

If I understand correctly, the problem it is that if you get data into your app (for instance through HTTP requests), call `JSON.parse` on the data, and then want to show the data by putting it into the virtual DOM you may be vulnerable. If an adversary manages to get some JSON into the data that happens to be a valid VNode he can execute code.

I'm wondering if we could solve this in alternative ways. Instead of changing the key for `key` could we alternatively add a new property containing something that can never occur when parsing JSON? For instance something like:

```js
export interface VNode {
  safetyTag: Function;
  //...
}
```

Then, to ensure that a VNode wasn't injected you can check that `safetyTag` is a function. I'm thinking that might work since JSON parse cannot result in a function. Such a trick would have the benefit that it doesn't rely on symbols. Alternatively, could `VNode` be a class with some method?

I also have to stress that for people who create `VNode`s without using `h` this will be a **breaking change**. For those that do use `h`, any changes to the `VNode` interface can be made there without users noticing.

lukaszmakuch:
I discovered this problem when I was preparing to write a blog post about the cause of injection attacks. I just got an idea to check how my favorite VDOM library deals with it. That's how I discovered an opportunity to contribute to snabbdom. :) 

So far I've written just two applications using snabbdom. One is a demo for a talk, and the other is the Rosmaro Editor. Fortunately right now the VDOM tree used there is not vulnerable, as most of the user input lands in attributes. Unfortunately, it's enough to add something like a "b" tag with user input to the tree in order to make the app vulnerable.

The Rosmaro Editor allows the user to load a JSON file, but it is not the only way how a vulnerability may be exploited. It could simply consume some API. In the era of schema-less databases (I love MongoDB, but this is the risky part...) it is quite possible that some API will consume a PUT request with an object instead of a string, then save it to the database, and then return to the user. Another scenario is reading directly from the local storage or a cookie. It poses a threat to people using shared computers, like in a library.

I really like your proposition to make VDOM a class. Instead of using .sym we could use .sym() and it should do the job!

You're also right that changing the interface of VNode will be a breaking change for people who don't use its factory function. I must admit I don't know how to mitigate this.

lukaszmakuch:
I've changed the VNode to a class. Now it works without symbols. 
```ts
export class VNode {
  constructor(
    private _sel?: string,
    public data?: VNodeData,
    public children?: VNodeData,
    public elm?: Node,
    public text?: string,
    public key?: Key,
  ) {}
  sel(): string | undefined {
    return this._sel;
  }
}
```

Unfortunately this doesn't help users users who build VNodes manually.

staltz:
Very interesting PR. I think either with class or symbol, someone who build VNodes manually will have to anyway import something (either import the class or import the symbol) in order to construct them.

@aronallen : you've been building passing serialized vnodes back and forth from Web Workers for the purpose of security, I think this thread and PR may interest you.

aronallen:
I think this is better handled with a snabbdom module on the create hook. I have a module with a whitelist of allowed tagNames. Generally I try to avoid using the selector syntax "div.cell", and I only provide the tag name.
The VNode is not dangerous until a real node is created, thus i think any security should be in the patching layer, and not in the definition layer.

lukaszmakuch:
I'm afraid that whitelisting allowed tags may not be enough, as it still gives the attacker the possibility of injecting arbitrary HTML with things like fake login forms 🔐 or simply a gigantic pink unicorn 🦄 spanning across the whole page.

aronallen:
@lukaszmakuch we might have different use cases.

I am trying to transfer serialised VDOM trees over a thread wall, so this would just add unneeded I/O to my web workers. I "trust" the worker, insofar that it won't serve me up rotten VTrees.

This is a concern most applications out there don't have, but I would be cautious to add this to snabbdom, as it is based on the assumption that you only would want to render VNodes that have been created in the main thread, and not on a foreign thread or machine.

lukaszmakuch:
@aronallen I see two different solutions to this problem. 

1. Mark every node created by the programmer with an instance of RegExp (it survives structured cloning)
2. Pass data-only javascript objects from thread to thread and after receiving them from a trusted source attach necessary methods (so they meet the interface)

What do you think?

lukaszmakuch:
I played a bit with RegExps. The result:
- protects against injections coming from HTTP requests
- should work in IE
- works with structured cloning (postMessage, Workers)

The disadvantages I can see are all about reading rendered VNodes from databases. Because this approach works with structural cloning, it is possible to read VNodes from IndexedDB. Also MongoDB will store a VNode like that.

aronallen:
@lukaszmakuch allocating a regex for each node seems a bit excessive.
They could just all have a reference to NaN or Infinity. These values cannot be serialized by JSON.
I currently don't have a use case for rendering nodes from a database, but someone else might.

lukaszmakuch:
@aronallen I replaced the RegExp with Infinity.

aronallen:
Cool.

I think this makes a lot of sense from a security standpoint, and this change won't break my apps but it may break others.

Perhaps there is  a space for a utility function called `confide`, `secure` or `trust` that lets you pass an untrusted VTree and returns a trusted VTree. Then anyone relying on serialization of VNodes can easily upgrade without changing  too much of their code-base.

Alternitavely we could introduce a new function called spatch for safe-patch, and let patch remain unsafe and issue a deprecation warning or something like that.

lukaszmakuch:
I like the `trust` name. It's easy to guess that it gets something what may not be trusted and turns it into a trusted thing.

I implemented it and it looks like that:
```javascript
var trust = require('snabbdom/trust').default; 
var h = require('snabbdom/h').default;

var originalVNode = h('div', {}, ['hello', h('span#world')]);
// It's untrusted, because the `safetyTag` attribute doesn't survive JSON serialization.
var untrustedVNode = JSON.parse(JSON.stringify(originalVNode));
// All three nodes are trusted.
var trustedVNode = trust(untrustedVNode);
```

aronallen:
@lukaszmakuch cool

I guess trust should be recursive.

lukaszmakuch:
@aronallen as far as I understand, it actually is. My reasoning behind this is that every VNode object has an optional list of children, what makes it a tree. In the example above, originalVNode is a root node with two children. Passing it to the `trust` function gives us a new tree with all three nodes trusted. 

Is this the expected behavior? If not, could you please give me a hint how can I improve it?

aronallen:
@lukaszmakuch, great i just couldn’t see the implementation when you wrote the comment before.

I am happy with these changes, it is for the betterment of snabbdom. But I am not her custodian, I am just a humble user.

paldepind:
> allocating a regex for each node seems a bit excessive.

I like `Infinity` but one could just create a single regex and then reuse it for all created vnodes.

I think a `trust` function is a really good idea.

@lukaszmakuch Didn't you mention that MongoDB could transfer `Infinity` values? Are we ok with that?

lukaszmakuch:
@paldepind it turns out MongoDB and IndexedDB will also store a RegExp. 

The only solution I know about would be to use a Symbol, but this comes at the cost of dropping support for older browsers and (what's even more important) makes it impossible to transfer nodes using postMessage or Workers.

We could try to mix two approaches. Instead of reading the `sel` property directly, we can call a method to read it. It's exactly what you suggested. Then, every time we receive a VNode from a trusted source (like a verified postMessage), we can use the `trust` helper to turn it onto a trusted one. Because the `trust` helper may be used quite often in such a scenario, I think it's worth considering making it impure. That is instead of building a new tree, it could simply mutate the one we pass to it.

That way it would be hard to accidentally read a valid VNode from a database. In the same time it shouldn't be 

There's just once scenario I'm a bit worried about:
1. Window A gets a fake VNode instead of a string 
2. Window A builds a VNode with the fake VNode as its child
3. Window A sends this VNode to window B
4. Window B trust the received VNode
5. Security problem

It's possible, because currently VNodes are verified only when they are used to patch something, and not when they are built. I suggest that we also make sure it's impossible to build a VNode with untrusted children. That way we could turn the scenario mentioned above into this one:
1. Window A gets a fake VNode instead of a string 
2. Window A builds a VNode with the fake VNode as its child
3. Error

What do you think guys?

caridy:
This solution does not belong into snabbdom! Remember, snabbdom is a very low level API that allows libraries and frameworks to tap into it to render a UI. If you need to sign your vnodes, to prevent injection of any kind, we are giving you the hooks for you to do so:

a) You can sanitize vnodes creation with your own copy of `h()`, instead of the one built-in version. If you're creating your own vnodes, then it is your responsibility to provide valid values.
b) if you want to sanitize vnodes globally, you have the hooks for it.

Snabbdom does not define a global variable that can be exploit, so it is definitely a responsibility of the app to protect against those things.

Alternative, I think we will be fine with a standard module that helps with this things, but definitely not inside the engine itself.

lukaszmakuch:
Thank you for a detailed answer. I totally understand your point of view. :) The simplicity of snabbdom is one of the reasons why did I decide to use it. It was what I needed and nothing more.

I think that it's doable to introduce a security mechanism as a module (to make sure we're patching only with signed VNodes) + a separated VNode helper (to make sure we're not building a node with unsigned children). 

The reason why did I suggest introducing changes in the core in the first place is because personally I find it challenging to really understand every security detail of every tool I use. As a newcomer in the world of snabbdom, I incorrectly assumed that some things I had learned about React also applied to snabbdom. Specifically, this characteristic of React:

> It is safe to embed user input in JSX:
```
const title = response.potentiallyMaliciousInput;
// This is safe:
const element = <h1>{title}</h1>;
```
> By default, React DOM escapes any values embedded in JSX before rendering them. Thus it ensures that you can never inject anything that’s not explicitly written in your application. 

So far I've used snabbdom only in two applications. In one of them I used the built-in `h` helper and in the other I used `snabbdom-jsx`. Much to my surprise both of them turned out to be vulnerable, as neither `<b>{text}</b>` nor `h('b', text)` provides protection against injection attacks.

My mistake was to assume that things I didn't actually know about snabbdom were like I wanted them to be. 

Anyway, I'm willing to create a module and an alternative `h` helper for that if this is the way to go. :)

lukaszmakuch:
I've created a tiny package called [snabbdom-signature](https://github.com/lukaszmakuch/snabbdom-signature). 

It consists of 2 main parts:
- `h` decorator called `signingH`
- `denyUnlessSigned` module defining update/create hooks

There are also 2 helper functions:
- `sign` used when a trusted tree is coming from some serialized data and it doesn't contain the required Symbols
- `removeSignature` used when we want to serialize a tree and we need to get rid of Symbols because the serialization algorithm doesn't handle them

Long story short, it should make the following fragment XSS-free:
```javascript
const text = userInput.potentiallyMaliciousInput;
const vnode = h('p', text);
```

Please let me know If you have any idea how could I improve it. :)

caridy:
@lukaszmakuch this is wonderful! thanks for putting this together. My only recommendation is to revisit those names (exported methods) before the package becomes popular, maybe you can open an issue in that repo to bikeshed on the names.

As for us, we should probably add a section in the readme, or create a new `.md` file to list out modules that can work with snabbdom with a brief description and the link to them, so we can start with your module.

caridy:
Btw, I'm closing this in favor of your package.

lukaszmakuch:
@caridy I've just opened [an issue regarding the names](https://github.com/lukaszmakuch/snabbdom-signature/issues/1). I'd be really happy if you could suggest there some better names. :)

PR #378, opened 2018-05-15
Updated documentation for eventlisteners and fixed missing event/vnode for arrays of handlers

kay999:


caridy:
I'm probably OK with this change, but tests are missing.

paldepind:
Thank you for the bugfix and added documentation @kay999 :+1:  This is now part of the most recent release.

PR #384, opened 2018-05-28
Remove redundant is.array check in h

blikblum:
In h function, the children variable can be assigned four times (lines 28, 30, 32 and 34) always to an array, making the is.array check in line 37 redundant.

### DIFF
--- a/src/h.ts
+++ b/src/h.ts
@@ -34,7 +34,7 @@ export function h(sel: any, b?: any, c?: any): VNode {
     else if (b && b.sel) { children = [b]; }
     else { data = b; }
   }
-  if (is.array(children)) {
+  if (children !== undefined) {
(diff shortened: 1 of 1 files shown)


staltz:
What if the function is called with `h('div')`? I think, then, we would need the is.array check

blikblum:
The is.array check is always done in every argument combination for the second or third argument.

This PR does not change that, only the subsequent is.array call is removed

paldepind:
> In h function, the children variable can be assigned four times (lines 28, 30, 32 and 34) always to an array, making the is.array check in line 37 redundant.

That is a very good catch 👍 We only ever assign an array value to `children`. So unless it is `undefined` it will necessarily be an array.

Thank you for the PR @blikblum. It's definitely nice to get rid of small unnecessary oddities like that.

PR #387, opened 2018-08-06
Add toJSON to vnodes

hath995:
I wanted to be able to send vnodes to the server and needed to convert them to JSON. The elm properties creates circular references that break JSON.stringify. This pr adds the needed method.

### DIFF
--- a/src/vnode.ts
+++ b/src/vnode.ts
@@ -9,6 +9,7 @@ import {Dataset} from './modules/dataset'
 import {Hero} from './modules/hero'
 
 export type Key = string | number;
+export type JsonVNode = Pick<VNode, Exclude<keyof VNode, "elm" | "toJSON">>;
 
 export interface VNode {
   sel: string | undefined;
@@ -17,7 +18,7 @@ export interface VNode {
   elm: Node | undefined;
   text: string | undefined;
   key: Key | undefined;
-  toJSON(): Pick<VNode, Exclude<keyof VNode, "elm" | "toJSON">>;
+  toJSON(): JsonVNode;
 }
 
 export interface VNodeData {
@@ -37,7 +38,7 @@ export interface VNodeData {
   [key: string]: any; // for any other 3rd party module
 }
 
-function toJSON(this: VNode): Pick<VNode, Exclude<keyof VNode, "elm" | "toJSON">> {
+function toJSON(this: VNode): JsonVNode {
   return {
     sel: this.sel,
     data: this.data,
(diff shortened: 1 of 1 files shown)


staltz:
Hi @hath995, I believe you could implement this function as a `this`less function, taking the vnode as argument and returning the jsonvnode. Additionally, it doesn't need to be in snabbdom proper, you could make it be its own library.

hath995:
Could you elaborate more, I was under the impression that toJSON is a magic method which is used by JSON.stringify on objects and is therefor mandatory. I also tried to make this update initially by directly patching vnode externally but that didn't work, probably because other snabbdom components had a reference to the original function which did not have the toJSON method.

staltz:
> I was under the impression that toJSON is a magic method which is used by JSON.stringify on objects and is therefor mandatory

Oh, I didn't know about that.

I'll let other people comment on this pull request, then.

hath995:
An alternative that might have better performance. 
```typescript
export type Key = string | number;
export type JsonVNode = Pick<VNode, Exclude<keyof VNode, "elm" | "toJSON">>;

export interface VNode {
  sel: string | undefined;
  data: VNodeData | undefined;
  children: Array<VNode | string> | undefined;
  elm: Node | undefined;
  text: string | undefined;
  key: Key | undefined;
}

export interface VNodeData {
  props?: Props;
  attrs?: Attrs;
  class?: Classes;
  style?: VNodeStyle;
  dataset?: Dataset;
  on?: On;
  hero?: Hero;
  attachData?: AttachData;
  hook?: Hooks;
  key?: Key;
  ns?: string; // for SVGs
  fn?: () => VNode; // for thunks
  args?: Array<any>; // for thunks
  [key: string]: any; // for any other 3rd party module
}

class VNodeC implements VNode {

  constructor(public sel: string | undefined,
              public data: any | undefined,
              public children: Array<VNode | string> | undefined,
              public text: string | undefined,
              public elm: Element | Text | undefined,
              public key: Key) {}
  
  toJSON(this: VNode): JsonVNode {
    return {
      sel: this.sel,
      data: this.data,
      children: this.children,
      text: this.text,
      key: this.key,
    };
  }
}

export function vnode(sel: string | undefined,
                      data: any | undefined,
                      children: Array<VNode | string> | undefined,
                      text: string | undefined,
                      elm: Element | Text | undefined): VNode {
  let key = data === undefined ? undefined : data.key;
  return new VNodeC(sel, data, children, text, elm, key);
}

export default vnode;
```

hath995:
Although this was handy I found another way to do this with json.stringify resolvers.

PR #390, opened 2018-08-24
Force evaluation of the element stylesheet

jvanbruegge:
This is the fix for the issue https://github.com/cyclejs/cyclejs/issues/747
I assume that modern Chrome (the only browser that failed AFAICT), uses some sort of optimization which causes the transform not to be applied as transition, but as a normal style property and thus not emitting the `transitionend` event.
Apparently using `getComputedStyle().transition` forces the browser to evaluate that

See https://github.com/snabbdom/snabbdom/issues/367#issuecomment-415849750

### DIFF
--- a/src/modules/style.ts
+++ b/src/modules/style.ts
@@ -8,6 +8,7 @@ export type VNodeStyle = Record<string, string> & {
 
 var raf = (typeof window !== 'undefined' && window.requestAnimationFrame) || setTimeout;
 var nextFrame = function(fn: any) { raf(function() { raf(fn); }); };
+var reflowForced = false;
 
 function setNextFrame(obj: any, prop: string, val: any): void {
   nextFrame(function() { obj[prop] = val; });
@@ -66,6 +67,10 @@ function applyRemoveStyle(vnode: VNode, rm: () => void): void {
     rm();
     return;
   }
+  if(!reflowForced) {
+    getComputedStyle(document.body).transform;
+    reflowForced = true;
+  }
   var name: string, elm = vnode.elm, i = 0, compStyle: CSSStyleDeclaration,
       style = s.remove, amount = 0, applied: Array<string> = [];
   for (name in style) {
@@ -83,7 +88,12 @@ function applyRemoveStyle(vnode: VNode, rm: () => void): void {
   });
 }
 
+function forceReflow() {
+  reflowForced = false;
+}
+
 export const styleModule = {
+  pre: forceReflow,
   create: updateStyle,
   update: updateStyle,
   destroy: applyDestroyStyle,
--- a/test/style.js
+++ b/test/style.js
@@ -116,6 +116,22 @@ describe('style', function() {
     fakeRaf.step();
     assert.equal(elm.style.fontSize, '20px');
   });
+  it('applies tranform as transition on remove', function(done) {
+    var btn = h('button', { style: {
+        transition: 'transform 0.5s',
+        remove: { transform: 'translateY(100%)' }
+    }}, ['A button']);
+    var vnode1 = h('div.parent', {}, [btn]);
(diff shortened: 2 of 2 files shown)


jvanbruegge:
cc @staltz

jvanbruegge:
I want to test the fix first in a few more browsers

jvanbruegge:
Ok, I tested this fix with Browserstack and this is the result:
| Browser | Without fix | With fix |
|-----------|--------------|-----------|
| Chrome 68 | **fail** | _pass_ |
| Firefox 61 | **fail** | **fail** |
| Safari 11.1.2 | **fail** | _pass__ |
| Chrome Mobile 63 | **fail** | _pass_ |
| Chrome 49 | **fail** | _pass_ |
| Firefox 52 | **fail** | **fail** |
| Safari 9.1.3 | **fail** | _pass_ |
| Safari Mobile 10 | **fail** | _pass_ |
| Edge 17 | **fail** | **fail** |
| IE 11 | _pass_ | _pass_ |
| IE 10 | _pass_ | _pass_ |

paldepind:
Would you consider adding a test to the tests as well? :smile:

Having to use `getComputedStyle` is a not optimal as `getComputedStyle` [forces the browser to do a layout reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a) which is quite bad for performance. If could be triggered several times in a single patch which will result in many layout reflows.

It would be really nice if we could find a solution that also works in Edge and Firefox and which isn't so expensive with regards to performance. I don't have any ideas unfortunately.

jvanbruegge:
Thanks to the link be @paldepind I was able to find a fix that works for all browsers and is cheap from a performance standpoint.
I've also added a test that fails without the fix, but passes with it

staltz:
Smart to do it in a `pre` hook 👍

jvanbruegge:
@paldepind Can this be merged?

paldepind:
Now `getComputedStyle` will only be called once on every `patch` 👍  But it will be called even if no elements are removed. I think the solution would be even better if the `getComputedStyle` call was only executed when an element is removed.

Maybe we could do this: The `style` module sets a flag `reflowForced = false` in the `pre` hook. Then `forceReflow` is called in the `remove` hook only if `reflowForced === false`. At the same time, `reflowForced` is set to `true` such that no additional reflows will be forced if other elements are removed. With that addition, `getComputedStyle` would only be called a single time and if no elements are removed in a patch it would be called zero times.

Does the above make sense? What do you think about it?

src/modules/style.ts
```diff
@@ -83,7 +83,12 @@ function applyRemoveStyle(vnode: VNode, rm: () => void): void {
   });
 }
 
+function forceReflow() {
+    getComputedStyle(document.body).transform;
```
paldepind: Indent with 2 spaces 😉

jvanbruegge:
done

paldepind:
Looks good 👍

Thanks a lot @jvanbruegge for tackling this issue and contributing a great PR with a fix. Especially the nasty difference between `.transition` and `.transform` is pretty hairy. I highly appreciate that you went through the effort of testing it with Browserstack.

jvanbruegge:
@paldepind if you want I could set up Travis with Browserstack for snabbdom too, i can mirror the cycle setup

You would have to write them an email to get an open source account first.

paldepind:
That would be awesome! I've contacted Browserstack through their website. I'll let you know when I hear from them :smile:

jvanbruegge:
@paldepind Also could you release a new version with this fix included? So I can close the issue in cycle.js too

paldepind:
Released as 0.7.2 👍

paldepind:
@jvanbruegge Browserstack has replied and I've created an account. I've not used Browserstack before. Do you know what the next step is?

PR #392, opened 2018-08-28
make it obvious how to unmount

trusktr:


paldepind:
That's a nice little trick to document. Thanks a lot, @trusktr.

PR #393, opened 2018-08-29
Replace broken SVG classes example in readme with link to classList polyfill

jscissr:
This snippet was added in #287, but it clearly was never tested because it doesn't work (the `underline` class is simply overwritten).


paldepind:
Thank you for the pull request @jscissr. I agree with your reasoning. It's better and simpler to just use a polyfill instead of an odd workaround.

PR #397, opened 2018-09-03
Explain insertBefore in README

### LINKED ISSUE #395: Error: insertBefore when reusing vnodes
Reproducable example:
```js
var sharedNode = h('div', {}, 'Selected');
var vnode1 = h('div', [
  h('div', {}, ['One']),
  h('div', {}, ['Two']),
  h('div', {}, [sharedNode]),
]);
var vnode2 = h('div', [
  h('div', {}, ['One']),
  h('div', {}, [sharedNode]),
  h('div', {}, ['Three']),
]);
patch(vnode0, vnode1);
patch(vnode1, vnode2);
```
This example crashes on the second `patch`.
The reason is that when the actual DOM node for `sharedNode` on position two is created, it assigns it as `vnode.elm`. The `vnode` is still that same `sharedNode`, that the third node still has a reference to (as it is not yet processed). When the third node gets created, it tries to do `insertBefore` with `vnode.elm` which is **not** the actual elm from the last patch (on position 3), but the new dom node of `sharedNode` created in **this** patch (position 2). Thus the `insertBefore` crashes as the element is not a child of the parent of the third node.

Note that this issue only happens in this direction (first insert, then remove) if you swap the two patches this does not happen, as first the node gets removed (and thus the reference) and then a new one gets inserted

This is the cause for https://github.com/cyclejs/cyclejs/issues/713

I am not sure what is the best way to solve this.

jvanbruegge:
fixes #395

### DIFF
--- a/README.md
+++ b/README.md
@@ -708,3 +708,43 @@ Here are some approaches to building applications with Snabbdom.
 
 Be sure to share it if you're building an application in another way
 using Snabbdom.
+
+## Common errors
+
+```
+Uncaught NotFoundError: Failed to execute 'insertBefore' on 'Node':
+    The node before which the new node is to be inserted is not a child of this node.
+```
+The reason for this error is reusing of vnodes between patches (see code example), snabbdom stores actual dom nodes inside the virtual dom nodes passed to it as performance improvement, so reusing nodes between patches is not supported.
+```js
+var sharedNode = h('div', {}, 'Selected');
+var vnode1 = h('div', [
+  h('div', {}, ['One']),
+  h('div', {}, ['Two']),
+  h('div', {}, [sharedNode]),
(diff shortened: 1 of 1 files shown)


paldepind:
Looks good.

Personally I prefer to wrap elements like that in a function:

```
var sharedNode = () => h('div', {}, 'Selected');
```

And then I call them in my views. This also ensures a fresh vnode on every patch.

What do you think about suggesting that as an alternative as well?

jvanbruegge:
Added it

jethrolarson:
Thanks for putting this in, I found it useful today :)

staltz:
Seems like it was approved by enough people, and it's just docs, so I'll merge. :)

PR #398, opened 2018-09-09
fix calling multiple event listeners (fixes #316)

### LINKED ISSUE #316: Multiple eventlistners can't get the event.
I want to add multiple event listener to an vnode, So I write code like this.
```javascript

function fn1 (event) {
    console.log(event);
    // ...
}

function fn2 (event) {
    // ...
}

h('button', {
    on: {
        click: [
            [fn1],
            [fn2]
    }
});
```

But I get the event is undefined, so i checked the source code and find in `src/modules/eventlisteners.ts` 29 line

```javascript
// call multiple handlers
for (var i = 0; i < handler.length; i++) {
    invokeHandler(handler[i]);
}
```
Should not it be the following?

```javascript
// call multiple handlers
for (var i = 0; i < handler.length; i++) {
    invokeHandler(handler[i], vnode, event);
}
```
Is this a bug or how can i use multiple eventlistener?

cornerman:


kw7722:
Same as https://github.com/snabbdom/snabbdom/pull/378

Would really be nice to merge this finally because the functionality is in snabbdom for some time now but broken in if you need access to event or vnode.

paldepind:
Thank you for the fix @cornerman. And thank you for bringing #378 to our attention @kw7722.

Neither this PR nor #378 had any tests. But #378 had documentation so I decided to merge that one and add a test myself.

This bug is now fixed on the latest version 0.7.3 :smile:

PR #399, opened 2018-09-19
Move to karma, allow to test on browserstack

jvanbruegge:
ISSUES FIXED: #396 
Before this can be merged, @paldepind has to activate travis for this repository. I can't finish the travis config until this is done.
Now this allows for easy porting of the tests to typescript
Also this now generates coverage, so there is the possibility of using codecov.io or similar

PR #400, opened 2018-09-26
Finish CI configuration

jvanbruegge:
#399 was not finished yet, not this PR finishes the configuration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,10 @@ script:
   - npm test
 addons:
   browserstack:
-    username:
-      secure: <TODO>
+    username: "snabbdombot1"
     access_key:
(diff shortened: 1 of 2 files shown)


jvanbruegge:
I just learned that travis does not provide encrypted environment variables to pull requests from forks
That's why they are not encrypted in the config

jvanbruegge:
There are a few tests that fail in certain Browsers
@paldepind has to decide which tests to fix or which browsers to run, all browsers are configured in the `browserstack-karma.js` file

paldepind:
Sorry for the slow response to this.

Initially, I think we should run the tests in the browsers for which they pass. That would allow us to catch any regressions.

PR #407, opened 2018-10-24
Fix for patching node with text over node with children

mariusmuja:
When patching a text vnode over a vnode with children, the children vnodes must be removed first (so that their hooks are called). The `api.setTextContent` call will end up removing the children elements from the DOM.


paldepind:
Thank you for the Pr @mariusmuja. This looks correct to me!

Can you please add a test to the tests that would fail before this fix? Thank you.

mariusmuja:
@paldepind Thanks, I added a test.

paldepind:
Really great! Thanks a lot :smile:

PR #408, opened 2018-11-14
Run karma on npm test script

blikblum:
The test script was running testem which was recently replaced by karma


paldepind:
Thank you for the PR @blikblum. This was indeed wrong. Another PR has fixed this so I'll close this one.

Issue #360, opened 2018-01-08
Microsoft Edge error with style module delayed prop


whitecolor:
If `removed` property is used in `style` in MS Edge there is an error raised:

```
stack: "TypeError: Invalid calling object at nextFrame (eval code:4:33) at setNextFrame (eval code:6:5) at updateStyle (eval code:33:21) at createElm (eval code:86:17) at createElm (eval code:91:25) at createElm (eval code:91:25) at updateChildren (eval code:212:21) at patchVnode (eval code:257:21) at patch (eval code:287:13) at f (eval code:544:33)"
```
![image](https://user-images.githubusercontent.com/736697/34696172-96cfc556-f4ef-11e7-809b-29cb18a10ac9.png)

It seems due to this line:
https://github.com/snabbdom/snabbdom/blob/master/src/modules/style.ts#L9

Seems edge doesn't like assigning of native methods to some variables and using them.

Can be fixed if to replace:
```ts
var raf = (typeof window !== 'undefined' && window.requestAnimationFrame && function (cb) {window.requestAnimationFrame(cb)}) || setTimeout;
```

nlarche:
Hello, I've got the same issue on Edge and IE11 with snabbdom v0.7.0,  do you know if a correction is planed ? 
I can make a PR if needed.
thx

whitecolor:
You can make a PR, and hope that it will be merged someday. =)

Snabbdom maintainers seem to be not very attentive to some issues about some bugs.

@caridy @paldepind

caridy:
@whitecolor sorry guys, I was on a long vacation, trying to get back on track. If you can send a PR (with tests please), I can take a look at it for sure.

whitecolor:
```ts
var raf = (typeof window !== 'undefined' && (window.requestAnimationFrame).bind(window)) || setTimeout;
```

I've made PR https://github.com/snabbdom/snabbdom/pull/409

Please merge.

mightyiam:
Seems like this was fixed in #409.

PR #409, opened 2018-11-18
Fix for #360 error in IE/Edge

whitecolor:
Binding raf to window resolves the error.

### DIFF
--- a/src/modules/style.ts
+++ b/src/modules/style.ts
@@ -6,7 +6,7 @@ export type VNodeStyle = Record<string, string> & {
   remove?: Record<string, string>
 }
 
-var raf = (typeof window !== 'undefined' && window.requestAnimationFrame) || setTimeout;
+var raf = (typeof window !== 'undefined' && (window.requestAnimationFrame).bind(window)) || setTimeout;
 var nextFrame = function(fn: any) { raf(function() { raf(fn); }); };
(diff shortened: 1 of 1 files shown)


paldepind:
Hi @whitecolor. Thanks a lot for the fix to the error :+1: 

Would you mind adding a test that fails without the patch? That would be really nice. We prefer having tests for all bugfixes.

whitecolor:
Look at the changes, there is actually nothing special to test.

As a see tests are not running currently at all because of some error not related to this PR. If the test for `requestAnimFrame` pass then it will be ok, if such test exists...

paldepind:
What I mean is, the PR fixes a bug in IE/Edge. It would be nice to have a test that triggers that bug in IE/Edge.

whitecolor:
in IE you just can not do
```js
const raf = window.requestAnimation
raf(() => ...) // will trigger the error, so have to bind first
```

Don't know how tests here work in this repo yet.

But I think probably this test should trigger:
https://github.com/snabbdom/snabbdom/blob/master/test/style.js#L102

Is it passing in IE? Are tests tunned to do IE  testing in CI?

whitecolor:
`Edge` will soon anyway move to `webkit`, though not tomorrow of course.

PR #410, opened 2018-11-25
README: add pypyjs-todomvc

amirouche:


paldepind:
This doesn't seem to use Snabbdom.

PR #413, opened 2019-01-10
Copy over namespace when using thunk

thomaswelter:
This is to resolve a bug when using thunks.
Also remove a duplicate line


iambumblehead:
@jvanbruegge this PR could be closed; changes defined here are found in master branch sources already

paldepind:
Thanks for the PR @thomasvelter. This change has already been made in another PR that happened to be merged instead for no particular reason.

Thanks for the help @iambumblehead.

PR #414, opened 2019-01-16
shouldn't compare non-primitive type

clmystes:
As vnode is non-primitive type,  i think it's intuitive that mutating old node and patching it.


paldepind:
This is an optimization such that if Snabbdom sees two vnodes that are pointer equal it skips the diffing and patching since it assumes you don't mutate you vnodes.

PR #416, opened 2019-01-26
Remove the attribute the proper way

batiste:


kay999:
Checking via ! is risky, because this would also fire for '' or 0.

Better use` !== undefined` or at least` != null`.

batiste:
@kay999 you are right... But I didn't write this code. I can also try to fit that... Why not.

PR #418, opened 2019-02-07
Fixed #417

### LINKED ISSUE #417: Logical error in props module?
In the props module there seem to be handling the special case for 'value' props:

`if (old !== cur && (key !== 'value' || (elm as any)[key] !== cur)) ...`

I suppose this should handle the case that a changed value property isn't updated correctly if the user typed it in after snabbdom was changing it. 

But it doesn't work, because if old === cur (both vdom nodes) the check for 'value' doesn't even happen.

So maybe replace it with something like this:

`if (old !== cur || (key === 'value' && (elm as any)[key] !== cur)) ...`

This way changes to elm.value would always be in sync with the dom. Which is probably the intended behavior anyway.

kay999:
This shows and fixes https://github.com/snabbdom/snabbdom/issues/417#issue-407698836

This shows that the current implementation fails if the DOM is changed directly (for example by user input). Because snabbdom doesn't see any changes in the vdom, it doesn't patches it.

### DIFF
--- a/test/core.js
+++ b/test/core.js
@@ -264,6 +264,14 @@ describe('snabbdom', function() {
       patch(vnode1, vnode2);
       assert.equal(elm.src, undefined);
     });
+    it('changes an elements value-prop it was changed in the dom', function() {
+      var vnode1 = h('input', {props: {value: 'original'}});
+      patch(vnode0, vnode1);
+      vnode1.elm.value = 'changed';
+      var vnode2 = h('input', {props: {value: 'original'}});
+      elm = patch(vnode1, vnode2).elm;
+      assert.equal(elm.value, 'original');
+    });
     describe('using toVNode()', function () {
       it('can remove previous children of the root element', function () {
         var h2 = document.createElement('h2');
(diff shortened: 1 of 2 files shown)


kay999:
The CI server seems to have problems running testem in the moment, so the fix is marked as failed.

test/core.js
```diff
@@ -264,6 +264,14 @@ describe('snabbdom', function() {
       patch(vnode1, vnode2);
       assert.equal(elm.src, undefined);
     });
+    it('changes an elements value-prop it was changed in the dom', function() {
```
mightyiam:
```suggestion
    it('sets the `value` prop if it was changed in the DOM', function() {
```

kay999: The name change is ok.

I hope it won't take too long to get into a release, because it's a quite annoying problem if you want any kind of input-field completion/formatting.

One possible extension would be enabling similar behaviour for the 'selected' and 'checke' props. But those aren't as often overwritten as value so it's not as important.

But we may need a documentation update for this.

cornerman: I would be in favor of doing this for `selected` and `checked` as well, because overwritting these does happen (at least in my application). Thus, it would be really convenient to solve these properties, too.

PR #427, opened 2019-03-25
fix: npm test script `karma start`

### LINKED ISSUE #426: test script fails with missing `testem` command
And that is after running `npm install`.

Screenshot from 2019-03-25 11-21-04

I imagine that the command `testem` is supposed to originate from https://www.npmjs.com/package/testem.

Yet, `testem` is not in the `devDependencies`.

After an `npm install --save-dev testem`, and running `npm test`, I get this:
Screenshot from 2019-03-25 11-25-20

Pressing `q` doesn't really quit.

The `before_tests` in which the issue seems to be triggered seems to be this:
https://github.com/snabbdom/snabbdom/blob/81da0c124257d393e29796a13e92f6f3016ac20d/testem.json#L10

So, the first thing that came to my mind is trying to run that command independently:
Screenshot from 2019-03-25 11-28-11

And sure enough, there is a problem there. Seemingly a missing file, `test/index.js`. According to git bisect, it seems that the file existed and was removed by 2e73a926cdcbec30e712bed24973d13e5e18c089.

So, are we using testem or karma? It seems like `npx karma start` successfully runs the test suite.

So, this might be a simple fix of changing what the npm test script does.

mightyiam:
Fixes #426

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   },
   "scripts": {
     "pretest": "npm run compile",
-    "test": "testem",
+    "test": "karma start",
     "compile": "npm run compile-es && npm run compile-commonjs",
     "compile-es": "tsc --outDir es --module es6 --moduleResolution node",
     "compile-commonjs": "tsc --outDir ./",


mightyiam:
Since CI is not fully set up, the failures in CI should not block the merging of this PR, I feel.

paldepind:
I've merged #400 which fixes this as well as a couple of other things.

PR #436, opened 2019-06-16
Added Pureact

irony:
Suggested adding Pureact which only has one dependency - Snabbdom.


paldepind:
Looks cool! Thank you for sharing the project :smile:

PR #443, opened 2019-09-04
Remove travis browserstack addon to fix HTTP error

jvanbruegge:
I saw that in master the setup is broken, we had the same issue in the cycle repo lately. This is the fix we used

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -2,8 +2,7 @@ const ci = !!process.env.CI;
 const watch = !!process.env.WATCH;
 const live = !!process.env.LIVE;
 
-const identifier = process.env.BROWSERSTACK_LOCAL_IDENTIFIER;
-const ip = process.env.IP_ADDR;
+const ip = 'bs-local.com';
 
 const browserstack = require('./browserstack-karma.js');
 
@@ -35,9 +34,7 @@ module.exports = function(config) {
     },
     browserStack: {
       name: 'Snabbdom',
-      startTunnel: false,
       retryLimit: 3,
-      tunnelIdentifier: identifier,
     },
     browserNoActivityTimeout: 1000000,
     customLaunchers: browserstack,
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,7 @@ language: node_js
 node_js:
   - '6.10.1'
 script:
-  - export IP_ADDR=$(ip addr | grep eth -A 4 | grep 'inet ' | awk '{ print $2 }' | sed 's/\/..//')
   - npm test
-addons:
-  browserstack:
-    username: "snabbdombot1"
-    access_key:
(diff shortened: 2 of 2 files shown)


jvanbruegge:
@paldepind As you can see now the CI just does not pass because some tests fail is some browsers

paldepind:
Thanks a lot. I really appreciate it :+1: I'll uncomment the browsers that fail.

I have a quick question which you may know the answer to. I'm trying to create a BrowserStack badge that we can add to the readme. Then BrowserStack will upgrade us to their open source plan. They write this in their instructions:

> To add status badges to a project, you need to give it a project name. If you haven’t specified a project name for your builds / sessions, add the capability project: <project_name> in your test scripts.

When I look at the current projects we only have one named "Untitled Project". Do you know how we can change the configuration such that we get a project named "Snabbdom"?

karma.conf.js
```diff
@@ -2,8 +2,7 @@ const ci = !!process.env.CI;
 const watch = !!process.env.WATCH;
 const live = !!process.env.LIVE;
 
-const identifier = process.env.BROWSERSTACK_LOCAL_IDENTIFIER;
-const ip = process.env.IP_ADDR;
+const ip = 'bs-local.com';
```
paldepind: Why that ip?

jvanbruegge: That is a domain that browserstack owns and knows how to forward to the device. `localhost` or the old way do not work any more

jvanbruegge:
> When I look at the current projects we only have one named "Untitled Project". Do you know how we can change the configuration such that we get a project named "Snabbdom"?

Oh, yeah, forgot that. Just add `name: 'Snabbdom'` to the `browserStack` key in `karma.conf.js`. For the disclaimer this is what we wrote: https://github.com/cyclejs/cyclejs#thanks

jvanbruegge:
Oh, I saw that the name is already set, let me look for the other differences

paldepind:
In case it helps. To see the projects I used the following command:

```
> curl -u "snabbdombot1:Wx2FUSehQLBf3haqHphd" https://api.browserstack.com/automate/projects.json
[{"id":817691,"name":"Untitled Project","group_id":967240,"user_id":3319678,"created_at":"2019-09-04T09:35:51.000Z","updated_at":"2019-09-04T10:20:44.000Z"}]
```

jvanbruegge:
Maybe in the browserstack UI? I can not find differences to the cycle config

paldepind:
That's odd. It looks good to me as well.

For the time being I've added just a Travis badge and the same disclaimer as used in the CycleJS repo (I hope that is okay).

I've emailed BrowserStack to give us the full plan.

Thank you so much for helping set this up and for continuing to fix things.

jvanbruegge:
No problem, most of it was copy paste from cyclejs

PR #449, opened 2019-09-10
Revert "Disable failing browsers in CI configuration"

mightyiam:
This reverts commit 7e09d27f5680de198b7b363f574f531f4f1aa180.


paldepind:
Nice. Great to get these back in.

paldepind:
I haven't looked deeply into it, but some of these browsers seem to still fail?

image

#460 passes but #449 doesn't.

PR #451, opened 2019-09-12
snabbdom can be used with jsx/tsx via jsxFactory: jsx

### LINKED ISSUE #450: Make the h function compatible with jsx/tsx.
I know #6 was closed a long time ago. 

The snabbdom h function is pretty close to being jsx compatible. Except for handling null in attrs (snabbdom handles undefined) and passing children as rest arguments.

I want to send a small PR that makes the h function of snabbdom jsx compatible so we don't need to resort to something like this.

```js
        <div attrs={{}}>
          {new Array(100).fill(0).map(() => (
            <div on={{click: this.helpers.handleClick}}>{`${state.count} `.repeat(30)}</div>
          ))}
        </div>
```

Nothing against snabbdom-jsx. Great module, but one shouldn't require an extra module which does a number of other things than just use the built ins that are already 99% there.

I hope the PR won't be rejected or left in limbo state forever. I see a number of open PRs for years now and can't assess how active the repo owners are.

nojvek:
Fixes #450 

The h (hyperscript) function is 99% close to being jsx compatible. I acknowledge there is a snabbdom-jsx module, but I think it makes sense to have the built in h function support tsx mode with jsxFactory and babel @jsx pragma. The snabbdom-jsx module supports a lot of bells and whistles which the base h function doesn't really need.

This would make the base h function very usable in typescript projects without adding a lot of fat.

- [x] Ability to accept null props and children passed as rest args. 
- [x] Ability to flatten children and filter out falsey values.
- [x] Tests in core.js and a h.tsx test that tests the h function with jsxFactory.

Really hope this isn't rejected outright and we can have some sort of a conversation. I'm a big fan of snabbdom and it's simplicity. This makes snabbdom much more powerful without needing extra modules while still being relatively simple.

**Example**

```tsx
import * as assert from 'assert';
import {h} from '../h';

describe('snabbdom', function() {
  describe('hyperscript', function() {
    it(`can be used as a jsxFactory method`, function() {
      const Part = ({part}: {part: string}) => <span>{part}</span>
      const vnode = (
        <div>
          <a attrs={{href: 'https://github.com/snabbdom/snabbdom'}}>Snabbdom</a>
          and tsx
          {['work', 'like', 'a', 'charm!'].map(part => <Part part={part}></Part>)}
          {null && '🤕'}
          {'💃🕺🎉'}
        </div>
      )

      assert.equal(JSON.stringify(vnode), JSON.stringify({
        sel: 'div',
        data: {},
        children: [
          {sel: 'a', data: {attrs: {href: 'https://github.com/snabbdom/snabbdom'}}, text: 'Snabbdom'},
          {text: 'and tsx'},
          {sel: 'span', data: {}, text: 'work'},
          {sel: 'span', data: {}, text: 'like'},
          {sel: 'span', data: {}, text: 'a'},
          {sel: 'span', data: {}, text: 'charm!'},
          {text: '💃🕺🎉'},
        ]
      }));
    })
  });
});
```

### DIFF
--- a/src/h.ts
+++ b/src/h.ts
@@ -1,6 +1,6 @@
 import {vnode, VNode, VNodeData} from './vnode';
 export type VNodes = VNode[];
-export type VNodeChildElement = VNode | string | number | boolean | undefined | null;
+export type VNodeChildElement = VNode | string | number | undefined | null;
 export type ArrayOrElement<T> = T | T[];
 export type VNodeChildren = ArrayOrElement<VNodeChildElement>
 import * as is from './is';
--- a/src/h.ts
+++ b/src/h.ts
@@ -1,6 +1,6 @@
 import {vnode, VNode, VNodeData} from './vnode';
 export type VNodes = Array<VNode>;
-export type VNodeChildElement = VNode | string | number | undefined | null;
+export type VNodeChildElement = VNode | string | number | boolean | undefined | null;
 export type ArrayOrElement<T> = T | T[];
 export type VNodeChildren = ArrayOrElement<VNodeChildElement>
 import * as is from './is';
@@ -18,17 +18,17 @@ function addNS(data: any, children: VNodes | undefined, sel: string | undefined)
 }
 
 export function h(sel: string): VNode;
-export function h(sel: string, data: VNodeData): VNode;
+export function h(sel: string, data: VNodeData | null): VNode;
 export function h(sel: string, children: VNodeChildren): VNode;
-export function h(sel: string, data: VNodeData, children: VNodeChildren): VNode;
+export function h(sel: string, data: VNodeData | null, children: VNodeChildren): VNode;
 export function h(sel: any, b?: any, c?: any): VNode {
   var data: VNodeData = {}, children: any, text: any, i: number;
   if (c !== undefined) {
-    data = b;
+    if (b !== null) data = b;
     if (is.array(c)) { children = c; }
     else if (is.primitive(c)) { text = c; }
     else if (c && c.sel) { children = [c]; }
-  } else if (b !== undefined) {
+  } else if (b !== undefined && b !== null) {
     if (is.array(b)) { children = b; }
     else if (is.primitive(b)) { text = b; }
     else if (b && b.sel) { children = [b]; }
--- a/src/h.ts
+++ b/src/h.ts
@@ -24,7 +24,7 @@ export function h(sel: string, data: VNodeData | null, children: VNodeChildren):
 export function h(sel: any, b?: any, c?: any): VNode {
   var data: VNodeData = {}, children: any, text: any, i: number;
   if (c !== undefined) {
-    if (b !== null) data = b;
+    if (b !== null) { data = b; }
     if (is.array(c)) { children = c; }
     else if (is.primitive(c)) { text = c; }
     else if (c && c.sel) { children = [c]; }
--- a/src/jsx-global.d.ts
+++ b/src/jsx-global.d.ts
@@ -1,0 +1,14 @@
+import {VNode, VNodeData} from './vnode';
+
+declare global {
+  /**
+   * opt-in jsx intrinsic global interfaces
+   * see: https://www.typescriptlang.org/docs/handbook/jsx.html#type-checking
+   */
+  namespace JSX {
+    type Element = VNode;
+    interface IntrinsicElements {
+      [elemName: string]: VNodeData;
+    }
+  }
+}
--- a/src/jsx.ts
+++ b/src/jsx.ts
@@ -1,0 +1,39 @@
+import {vnode, VNode, VNodeData} from './vnode';
+import {h, VNodeChildren} from './h';
+
+export type FunctionComponent = (props: {[prop: string]: any} | null, children?: VNode[]) => VNode;
+
+function flattenAndFilter(children: VNodeChildren[], flattened: VNode[]): VNode[] {
+  for (const child of children) {
+    // filter out falsey children, except 0 since zero can be a valid value e.g inside a chart
+    if (child !== undefined && child !== null && child !== false &&  child !== '') {
+      if (Array.isArray(child)) {
+        flattenAndFilter(child, flattened);
+      } else if (typeof child === 'string' || typeof child === 'number' || typeof child === 'boolean') {
+        flattened.push(vnode(undefined, undefined, undefined, String(child), undefined));
+      } else{
+        flattened.push(child);
+      }
+    }
+  }
+  return flattened;
+}
+
+/**
+ * jsx/tsx compatible factory function
+ * see: https://www.typescriptlang.org/docs/handbook/jsx.html#factory-functions
+ */
+export function jsx(tag: string | FunctionComponent, data: VNodeData | null, ...children: VNodeChildren[]): VNode {
+  const flatChildren = flattenAndFilter(children, []);
+  if (typeof tag === 'function') {
+    // tag is a function component
+    return tag(data, flatChildren);
+  } else {
+    if (flatChildren.length == 1 && !flatChildren[0].sel && flatChildren[0].text) {
+      // only child is a simple text node, pass as text for a simpler vtree
+      return h(tag, data, flatChildren[0].text);
+    } else {
+      return h(tag, data, flatChildren);
+    }
+  }
+}
--- a/src/jsx.ts
+++ b/src/jsx.ts
@@ -1,9 +1,13 @@
 import {vnode, VNode, VNodeData} from './vnode';
-import {h, VNodeChildren} from './h';
+import {h, ArrayOrElement} from './h';
+
+// for conditional rendering we support boolean child element for falsey filtering
+export type JsxVNodeChild = VNode | string | number | boolean | undefined | null;
+export type JsxVNodeChildren = ArrayOrElement<JsxVNodeChild>
 
 export type FunctionComponent = (props: {[prop: string]: any} | null, children?: VNode[]) => VNode;
 
-function flattenAndFilter(children: VNodeChildren[], flattened: VNode[]): VNode[] {
+function flattenAndFilter(children: JsxVNodeChildren[], flattened: VNode[]): VNode[] {
   for (const child of children) {
     // filter out falsey children, except 0 since zero can be a valid value e.g inside a chart
     if (child !== undefined && child !== null && child !== false &&  child !== '') {
@@ -23,7 +27,7 @@ function flattenAndFilter(children: VNodeChildren[], flattened: VNode[]): VNode[
  * jsx/tsx compatible factory function
  * see: https://www.typescriptlang.org/docs/handbook/jsx.html#factory-functions
  */
-export function jsx(tag: string | FunctionComponent, data: VNodeData | null, ...children: VNodeChildren[]): VNode {
+export function jsx(tag: string | FunctionComponent, data: VNodeData | null, ...children: JsxVNodeChildren[]): VNode {
   const flatChildren = flattenAndFilter(children, []);
   if (typeof tag === 'function') {
     // tag is a function component
--- a/src/jsx.ts
+++ b/src/jsx.ts
@@ -1,7 +1,7 @@
 import {vnode, VNode, VNodeData} from './vnode';
 import {h, ArrayOrElement} from './h';
 
-// for conditional rendering we support boolean child element for falsey filtering
+// for conditional rendering we support boolean child element e.g cond && <tag />
 export type JsxVNodeChild = VNode | string | number | boolean | undefined | null;
 export type JsxVNodeChildren = ArrayOrElement<JsxVNodeChild>
 
--- a/test/core.js
+++ b/test/core.js
@@ -67,6 +67,12 @@ describe('snabbdom', function() {
       var vnode = h('a', {}, 'I am a string');
       assert.equal(vnode.text, 'I am a string');
     });
+    it('can create vnode with null props', function() {
+      var vnode = h('a', null);
+      assert.deepStrictEqual(vnode.data, {});
+      vnode = h('a', null, ['I am a string']);
+      assert.equal(vnode.children[0].text, 'I am a string');
+    });
     it('can create vnode for comment', function() {
       var vnode = h('!', 'test');
       assert.equal(vnode.sel, '!');
--- a/test/jsx.tsx
+++ b/test/jsx.tsx
@@ -1,0 +1,118 @@
+import * as assert from 'assert';
+import {jsx} from '../src/jsx';
+
+describe('snabbdom', function() {
+  describe('jsx', function() {
+    it('can be used as a jsxFactory method', function() {
+      const vnode = <div title="Hello World">Hello World</div>;
+
+      assert.equal(JSON.stringify(vnode), JSON.stringify({
+        sel: 'div',
+        data: {title: 'Hello World'},
+        text: 'Hello World',
+      }));
+    });
+
+    it('creates text property for text only child', function() {
+      const vnode = <div>foo bar</div>;
+
+      assert.equal(JSON.stringify(vnode), JSON.stringify({
+        sel: 'div',
+        data: {},
+        text: 'foo bar',
+      }));
+    });
+
+    it('creates an array of children for multiple children', function() {
+      const vnode = <div>{'foo'}{'bar'}</div>;
+
+      assert.equal(JSON.stringify(vnode), JSON.stringify({
+        sel: 'div',
+        data: {},
+        children: [
+          {text: 'foo'},
+          {text: 'bar'},
+        ]
+      }));
+    });
+
+    it('flattens children', function() {
+      const vnode = (
+        <section>
+          <h1>A Heading</h1>
+          some description
+          {['part1', 'part2'].map(part => <span>{part}</span>)}
+        </section>
+      );
+
+      assert.equal(JSON.stringify(vnode), JSON.stringify({
+        sel: 'section',
+        data: {},
+        children: [
+          {sel: 'h1', data: {}, text: 'A Heading'},
+          {text: 'some description'},
+          {sel: 'span', data: {}, text: 'part1'},
+          {sel: 'span', data: {}, text: 'part2'},
+        ],
+      }));
+    });
+
+    it('removes falsey children', function() {
+      const showLogin = false;
+      const showCaptcha = false;
+      const loginAttempts = 0;
+      const userName = ``;
+      const profilePic = undefined;
+      const isLoggedIn = true;
+      const vnode = (
+        <div>
+          Login Form
+          {showLogin && <login-form />}
+          {showCaptcha ? <captcha-form /> : null}
+          {userName}
+          {profilePic}
+          Login Attempts: {loginAttempts}
+          Logged In: {isLoggedIn}
+        </div>
+      );
+
+      assert.equal(JSON.stringify(vnode), JSON.stringify({
+        sel: 'div',
+        data: {},
+        children: [
+          {text: 'Login Form'},
+          {text: 'Login Attempts: '},
+          {text: '0'},
+          {text: 'Logged In: '},
+          {text: 'true'},
+        ],
+      }));
+    });
+
+    it('works with a function component', function() {
+      const Part = ({part}: {part: string}) => <span>{part}</span>
+      const vnode = (
+        <div>
+          <a attrs={{href: 'https://github.com/snabbdom/snabbdom'}}>Snabbdom</a>
+          and tsx
+          {['work', 'like', 'a', 'charm!'].map(part => <Part part={part}></Part>)}
+          {'💃🕺🎉'}
+        </div>
+      );
+
+      assert.equal(JSON.stringify(vnode), JSON.stringify({
+        sel: 'div',
+        data: {},
+        children: [
+          {sel: 'a', data: {attrs: {href: 'https://github.com/snabbdom/snabbdom'}}, text: 'Snabbdom'},
+          {text: 'and tsx'},
+          {sel: 'span', data: {}, text: 'work'},
+          {sel: 'span', data: {}, text: 'like'},
+          {sel: 'span', data: {}, text: 'a'},
+          {sel: 'span', data: {}, text: 'charm!'},
+          {text: '💃🕺🎉'},
+        ],
+      }));
+    })
+  });
+});
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,9 @@ node_modules
 /es/is.d.ts
 /es/is.js
 /es/is.js.map
+/es/jsx.d.ts
+/es/jsx.js
+/es/jsx.js.map
 /es/snabbdom.bundle.d.ts
 /es/snabbdom.bundle.js
 /es/snabbdom.bundle.js.map
@@ -101,6 +104,10 @@ node_modules
 /is.d.ts
 /is.js
 /is.js.map
+/jsx.d.ts
+/jsx-global.d.ts
+/jsx.js
+/jsx.js.map
 /snabbdom.bundle.d.ts
 /snabbdom.bundle.js
 /snabbdom.bundle.js.map
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -30,11 +30,14 @@ module.exports = function(config) {
     hostname: ci ? ip : 'localhost',
     preprocessors: {
       'src/**/*.ts': ['karma-typescript'],
-      'test/**/*.js': ['karma-typescript'],
+      'test/**/*.{js,ts,tsx}': ['karma-typescript'],
     },
     browserStack: {
       name: 'Snabbdom',
       retryLimit: 3,
+    },
+    client: {
+      captureConsole: true,
     },
     browserNoActivityTimeout: 1000000,
     customLaunchers: browserstack,
--- a/package-lock.json
+++ b/package-lock.json
@@ -98,6 +98,12 @@
         "through2": "^2.0.3"
       }
     },
+    "@types/assert": {
+      "version": "1.4.3",
+      "resolved": "https://registry.npmjs.org/@types/assert/-/assert-1.4.3.tgz",
+      "integrity": "sha512-491hfOvNr0+BGOHT2m36xJ+LK68IuOshvxV0VIrKOnzBDL11WlDa3PwO+drTYkwCdfzJRN9REcDPZVVcrx1ucw==",
+      "dev": true
+    },
     "@types/eslint-visitor-keys": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/@types/eslint-visitor-keys/-/eslint-visitor-keys-1.0.0.tgz",
@@ -108,6 +114,12 @@
       "version": "7.0.3",
       "resolved": "https://registry.npmjs.org/@types/json-schema/-/json-schema-7.0.3.tgz",
       "integrity": "sha512-Il2DtDVRGDcqjDtE+rF8iqg1CArehSK84HZJCT7AMITlyXRBpuPhqGLDQMowraqqu1coEaimg4ZOqggt6L6L+A==",
+      "dev": true
+    },
+    "@types/mocha": {
+      "version": "5.2.7",
+      "resolved": "https://registry.npmjs.org/@types/mocha/-/mocha-5.2.7.tgz",
+      "integrity": "sha512-NYrtPht0wGzhwe9+/idPaBB+TqkY9AhTvOLMkThm0IoEfLaiVQZwBwyJ5puCkO3AUCWrmcoePjp2mbFocKy4SQ==",
       "dev": true
     },
     "@types/normalize-package-data": {
@@ -3283,8 +3295,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -3305,14 +3316,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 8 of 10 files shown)


nojvek:
Kind ping again @paldepind. Not sure who else is active maintainer around here

nojvek:
Kind ping for review again.

Trying some more members this time @paldepind / @mightyiam / @kl2134 / @staltz

nojvek:
### Typescript
https://www.typescriptlang.org/docs/handbook/jsx.html#factory-functions

Also implemented in repo. https://github.com/snabbdom/snabbdom/pull/451/files#diff-e5e546dd2eb0351f813d63d1b39dbc48R10-R11

Basically two config options.

1. `jsx: 'react'` tells typescript to start transpiling jsx/tsx like `<div>Hello World</div> to `React.creatElement('div', null 'Hello World')`.

2. `jsxFactory: 'h'` tells typescript to use `h` function instead of default `React.createElement` so output is `h('div', null, 'Hello World')`

3. Alternatively one can also add a comment at top line of a file such as below and it has same effect as jsxFactory compiler setting.

```
import {h} from 'snabbdom';
/* @jsx preact.h */

const foo = <div>foo</div>;
```

### Babel 
It's very similar in babel world too.

https://babeljs.io/docs/en/babel-plugin-transform-react-jsx/

```
{
  "plugins": [
    ["@babel/plugin-transform-react-jsx", {
      "pragma": "h", // default pragma is React.createElement
    }]
  ]
}
```

or 

```
/** @jsx h */
import {h} from 'snabbdom'';

const foo = <div>foo</div>;
```

nojvek:
@mightyiam if you're okay with the meat of the PR, I can add commit for README.md and document how to use it with jsx + babel or tsx + typescript.

I've also added a tsx test in the repo to guard against future breaks.

mightyiam:
@nojvek thank you for your work. I'd like to start by seeing a spec on the API that the function is supposed to provide. How did you know that API? Am I missing something obvious?

nojvek:
API is well known in jsx/tsx/react community. Basically snabbdom already has this.

```
export function h(sel: string, data: VNodeData, children: VNodeChildren): VNode;
```

The jsx/tsx version is a strict superset of that.

```
// jsx/tsx compatible h signature
export function h(sel: string | Function, data: VNodeData | null, ...children: VNodeChildren[]): VNode;
```

The change is to handle null in data (currently only undefined is handled), and accept children as rest arguments too.

see react.d.ts https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L246-L249

and preact.d.ts https://github.com/matthewmueller/tstypes/blob/master/preact.d.ts#L116-L125

In terms of typescript, the behaviour is defined in https://www.typescriptlang.org/docs/handbook/jsx.html

The core is defined by JSX.IntrinsicElements interface https://basarat.gitbooks.io/typescript/docs/jsx/others.html

We have implemented that in the PR as https://github.com/snabbdom/snabbdom/pull/451/files#diff-70083611527b9d7d728e1e14dd8f977fR1-R9

Which will allow all tags, including custom elements. 

Snabbdom was already 99% jsx/tsx compatible. This is just the last 1%.

nojvek:
@mightyiam please let me know what are the next steps you'd like me to do. I acknowledge that open source is always hard and requires voluntary time commitment. Would really love to see this move slowly towards a path where it could get merged rather than stay in limbo and eventually abandoned.

src/h.ts
```diff
@@ -17,29 +17,48 @@ function addNS(data: any, children: VNodes | undefined, sel: string | undefined)
...
   }
   if (children !== undefined) {
-    for (i = 0; i < children.length; ++i) {
-      if (is.primitive(children[i])) children[i] = vnode(undefined, undefined, undefined, children[i], undefined);
-    }
+    children = flattenChildren(children, []);
   }
-  if (
+  if (typeof sel === `function`) {
+    return sel(data, children);
```
mightyiam: When is this the case, please?

src/h.ts
```diff
@@ -17,29 +17,48 @@ function addNS(data: any, children: VNodes | undefined, sel: string | undefined)
...
     if (is.array(b)) { children = b; }
     else if (is.primitive(b)) { text = b; }
     else if (b && b.sel) { children = [b]; }
     else { data = b; }
   }
   if (children !== undefined) {
-    for (i = 0; i < children.length; ++i) {
-      if (is.primitive(children[i])) children[i] = vnode(undefined, undefined, undefined, children[i], undefined);
-    }
+    children = flattenChildren(children, []);
```
mightyiam: Potential performance regression?

src/h.ts
```diff
@@ -1,6 +1,6 @@
 import {vnode, VNode, VNodeData} from './vnode';
 export type VNodes = Array<VNode>;
-export type VNodeChildElement = VNode | string | number | undefined | null;
+export type VNodeChildElement = VNode | string | number | boolean | undefined | null;
```
mightyiam: Why is this required, please?

src/jsx.d.ts
```diff
@@ -0,0 +1,9 @@
+import {VNode, VNodeData} from './vnode';
+
+declare global {
+  namespace JSX {
+    interface IntrinsicElements {
+      [elemName: string]: VNodeData;
```
mightyiam: On first look, I'm surprised that this is `VNodeData` and not `VNode`.

test/core.js
```diff
@@ -67,6 +67,48 @@ describe('snabbdom', function() {
...
+      assert.deepStrictEqual(vnode.data, {});
+      vnode = h('a', null, ['I am a string']);
+      assert.equal(vnode.children[0].text, 'I am a string');
+    });
+    it('can create vnode with children in vargs ', function() {
+      var vnode = h('a', null,
+        'some content',
+        h('span'),
+        ['part 1', 'part 2'],
+      );
```
mightyiam: This is how JSX passes children to the factory, right?

test/core.js
```diff
@@ -67,6 +67,48 @@ describe('snabbdom', function() {
...
+      }));
+    });
+    it('can create vnode with falsey children removed', function() {
+      var vnode = h('a', null,
+        false,
+        [null, undefined],
+        'hello',
+        true,
+        [0, 1],
+      );
```
mightyiam: Removal of falsey children is not an API I particularly like. Because it relaxes type checking in a way that instead of getting a proper type error when you've made a mistake, you get a missing element where you expect one.

In what case does JSX pass falsy values as children, please?

nojvek:
> @nojvek thank you for your contribution and your patience. What is the reason you chose to contribute JSX support to snabbdom instead of settling with using snabbdom-pragma, please?

Great Q, so I evaluated both the snabbdom-pragma and snabbdom-jsx. Nice modules. The reasoning for why I think it would be nice to have this support in core is following.

1) h function is already 99% there. Getting it jsx compatible was just adding null handing for data, rest arguments for children and flattening. It would keep the core simple and elegant, yet not need extra modules. Snabbdom would just work out of the box like preact, react e.t.c 

2) both module support two ways of passing data e.g `on-click={handler}` attr or `on={{click: handler}}`. While this may seem nice, string parsing 'on-click' to click not only has perf cost but now has two ways of doing same thing. And typing them needs redundancy. There's also the mental cost of what is default, props or attrs. Special parsing for `key` e.t.c  e.g you'd need to have

```
interface InputData extends VNodeData {
   on: {
       click: (ev: MouseEvent) => void;
   }
   on-click: (ev: MouseEvent) => void
}

declare global {
  namespace JSX {
    interface IntrinsicElements {
      input: InputData;
    }
  }
}
```

and do this for attrs, dataset and other snabb modules. It gets yucky pretty fast. In our open source library that stitches custom elements + snabbdom i.e https://github.com/mixpanel/panel, we really want to have a good typing experience when creating/using custom elements.

In the current PR, the typing is more opinionated but simpler. You always pass in data, like you pass into h calls. e.g `div attrs={{title: `hello`}} on={{click: handler}}`. It's fast, and simple. One way of doing things. NOTE: This doesn't break snabbdom-jsx or snabbdom-prgama. Typescript merges the global defintions, and the h function still preserves 100% backward compatible api.

The other alternative I was considering was just to make another module for this, but for ~10-15 lines of code in h function, I was hesitant to create even more pollution in npm modules. It's extra noise for someone to figure how to make snabbdom and jsx work. As a library user, having something simple with batteries included. We had already used the h function in jsx/tsx with some success.

e.g

```
/** @jsx h */
import {h} from 'snabbdom';

const template = ({$app}) => (
  <div attrs={{}}>
    {$app.showChild && <mp-blah-child />}
  </div>
)
```

Note since, snabbdom doesn't like null as data, we have to pass an empty `attrs={{}}`. Something this PR fixes

3) It's 100% backwards compatible api and doesn't break snabbdom-pragma, and snabbdom-jsx. If someone does want those extra features as `on-click` to `on={{click}}` and other fancy features, those libraries will still work, no changes necessary on their part.

src/h.ts
nojvek: See the falsey comment below.

src/h.ts
nojvek: Potential, I haven't benchmarked it. Does snabbdom have any benchmarking tests/utils for this. I can do a benchmark. I don't think it should be a major regression though compared to tree diffing which is a much bigger cost. Agreed though, best to benchmark and have conclusive results.

src/h.ts
nojvek: This is when a FunctionComponent is used. I've added a better type for that e.g 

```ts
import {h, init} from 'snabbdom';
const patch = init([..]);

function App() {
  const greeting = 'Hello Function Component!';
  return <Headline value={greeting} />;
}

function Headline({value}) {
  return <h1>{value}</h1>;
}

patch(document.getElementById(`app`), <App />)
```

jsx treats any functions begining with capital letter e.g `<Headline>` as components and small letters e.g `<h1>` as intrinsic elements. A function component is basically a function that takes some props and returns vnodes. This allows composition of a large h call tree into smaller re-usable functions. Something very encouraged when building large views. In h world they get transpiled to `h(Headline, {value: 'Hello Function Component'})`. So we're essentially just calling the function to retrieve the children equivalent to `Headline({value: 'Hello Function Component'})`

src/jsx.d.ts
nojvek: yeah so the type of Element is VNode as described on line 5. IntrinsicElement defines what props an element accepts. so we say any dom element (including custom elements) can be created and it accepts the structure of VNodeData.

e.g ```
<input
   key='somekey' 
   attrs={{title: `some input`, type: `text`}}
   on={{input: this.handleInput, click: this.handleClick}}
   props={{value: 'someValue'}}
   class={{fancy: true}}
/>
```

File: src/jsx.d.ts [hunk 7]
```diff
@@ -0,0 +1,9 @@
...
+    }
+  }
+}
```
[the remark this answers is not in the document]
nojvek: I've left it a bit free because I don't want snabbdom to be tied to html spec and what elements are accetable / not acceptable. Especially in world of custom elements where a user can create any name, we don't want things too strict. The user can make them strict if they wish.

e.g

```
<my-tag attrs={{super: true}} on={{superEvent: this.handleSuperEvent}} />
```

can be typed in the user library as.

```
interface MyTagData extends VNodeData {
  attrs: {
    super: boolean;
  }
  on: {
   superEvent: EventListener
  }
}

declare global {
  namespace JSX {
    interface IntrinsicElements {
      'my-tag': MyTagData;
    }
  }
}
```

since TS merges interfaces, this will support both my-tag and other html elements. For my-tag the compiler will be extra strict and validate that only certain attributes and events are allowed. It will also provide really nice code completion when coding jsx because it can infer the structure from MyTagData.

test/core.js
nojvek: that's right, the children will be passed as variable number of ending args.

so 

```
<div>
  <span>Hello</span>
  <span>World</span>
</div>
```

is transpiled as 

```
h("div", null,
    h("span", null, "Hello"),
    h("span", null, "World"));
```

e.g: https://typescript-play.js.org/?jsx=2#code/PQKhAIAECsGcA9wAtwmAKADwBMCWA3APnXHE1gAcBDAO0IAkBTAG2YHtNhLbjTzq6AdTYAnZtk7c6WYHiJA

And that's why we need the flattening because user can do

```ts
<div>
  <span>Hello</span>
  <span>World</span>
  {["one", "two", "three"].map(str => <span>{str}</span>)}
</div>
```

which transpiles to

```
h("div", null,
    h("span", null, "Hello"),
    h("span", null, "World"),
    ["one", "two", "three"].map(str => h("span", null, str)));
```

test/core.js
nojvek: The reason for removing falsey children is because this is a common usage

```ts
function render(showLogin: boolean, showCaptcha: boolean) {
  <div>
    {showLogin && <login-form />}
    {showCaptcha && <captcha-form />}
  </div>
}
```

which transpiles to

```
function render(showLogin, showCaptcha) {
    h("div", null,
        showLogin && h("login-form", null),
        showCaptcha && h("captcha-modal", null));
}
```

suppose showLogin and showCapcha are false, the call becomes `h('div', null, false, false)`. After flattening it becomes `h('div', null, [false, false])`. The expectation is the && if guards don't render `login-form` and `captcha-modal`. That's why I added boolean as primitive and was filtering it out.

My assumption is that it makes sense for the base h case as well. Generally most hyperscript/jsx based frameworks do filter out falsey (null, undefined, false) values due to the `xx && h(...)` kind of usage. 0 is a special case we don't want to filter because showing 0 in charts is common. Undefined will also be more common when `??` and `?.` (optional chaining) operators make it into spec. It's already stage3.

If you are against handling false, that is cool. We can only filter out null/undefined. That works with me.

in which case the usage will be

```
  <div>
    {showLogin ? <login-form /> : null}
    {showCaptcha  ? <captcha-form /> : null}
  </div>
```

src/h.ts
mightyiam: There seems to be a benchmark [here](https://github.com/snabbdom/snabbdom/tree/master/perf). Not sure whether it covers this change. If it doesn't, would you mind adding that coverage, please?

src/h.ts
```diff
@@ -1,8 +1,9 @@
...
 export type ArrayOrElement<T> = T | T[];
 export type VNodeChildren = ArrayOrElement<VNodeChildElement>
+export type FunctionComponent = (props: any, children?: VNode[]) => VNode;
```
mightyiam: Can there be some type checking on `props`?

src/jsx.d.ts
```diff
@@ -2,6 +2,7 @@ import {VNode, VNodeData} from './vnode';
 
 declare global {
   namespace JSX {
+    type Element = VNode;
```
mightyiam: What's this for, please?

nojvek: As the intrinsic interface it says what a jsx element type is. It’s a Typescript built in. 

Essentially any jsx node has its inferred type as JSX.Element. With a type alias. 

`const node = <div />` will have it’s inferred type as `VNode`

src/h.ts
nojvek: fixed

src/h.ts
nojvek: Thinking through this I had an idea. What if we don't bundle everything in h. Expose a different function `jsx` and have h only handle the null bit. 

This not only alleviates the perf concerns, it moves towards a design of composition

nojvek:
@mightyiam pulled out the jsx bits out of h.ts and made a separate jsx.ts which should give good separation of concerns.

src/jsx.ts
```diff
@@ -0,0 +1,36 @@
...
+  } else {
+    if (flatChildren.length == 1 && !flatChildren[0].sel && flatChildren[0].text) {
+      return h(tag, data, flatChildren[0].text);
+    } else {
+      return h(tag, data, flatChildren);
+    }
+  }
+}
+
+export default jsx;
```
mightyiam: Could this be a named export please?

nojvek: It’s also named. But I can remove the default if you want. Kept it consistent with how vnode.ts was doing things.

mightyiam:
@nojvek I value your contribution and I feel it's a fantastic addition to snabbdom.

At this point it is likely that the only person who could possibly press the green button is @paldepind.

I hope that happens soon.

src/jsx-global.d.ts
```diff
@@ -0,0 +1,10 @@
+import {VNode, VNodeData} from './vnode';
+
+declare global {
+  namespace JSX {
+    type Element = VNode;
+    interface IntrinsicElements {
+      [elemName: string]: VNodeData;
+    }
+  }
+}
```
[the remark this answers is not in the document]
nojvek: Good question. I don’t think i’ve encountered two jsx providers in one project since the jsxFactory option is set per Typescript project. 

I’m just following the Typescript and react/preact conventions.

nojvek:
Thanks for your kind words and thorough review @mightyjam. 

I’ll keep on weekly pinging @paldepind a couple of more times. He seems to be unresponsive. May be vacationing ? Not sure.

nojvek:
@paldepind kind ping ^

nojvek:
Kind ping again @paldepind ^ 

I can understand open source being a time sink, but the inactivity in the repo/PR doesn't inspire much confidence. 

I'd advice getting more active contributors and trusting them to help out so you're not a bottleneck.

nojvek:
Last ping @paldepind

paldepind:
I'm not sure if I'm a fan of having both a `jsx` function and a `h` function with a very similar purpose.

It seems to me like the changes made to `h` in order to make it JSX compatible are ok to have in the default `h` function?

We should remember that `h`i not a core part of Snabbdom. It is a helper function for creating vnodes. Putting some extra niceties into a helper function that is already supposed to be nice sounds ok to me. Whereas having two such helper functions seems a bit more complex.

What do you guys think? Was the original approach better?

mightyiam:
@paldepind copy-paste from https://github.com/snabbdom/snabbdom/pull/451#discussion_r329893655

> Removal of falsey children is not an API I particularly like. Because it relaxes type checking in a way that instead of getting a proper type error when you've made a mistake, you get a missing element where you expect one.

Having our `h` support JSX necessitates relaxing of type checking in this way, where falsey children are allowed. I feel that separating it into a JSX would be the right thing to do for that reason. To use the same term, the type checking is one of the niceties for me.

And think of it this way — the `h` function is for human readable code, while the `JSX` function is transpilation target for another human readable code (JSX). Them being quite similar should not mean that we are to sacrifice some niceties from `h` to support JSX. Let them be two different functions and everyone's happy, no?

mightyiam:
@paldepind that sounds reasonable for the humans but I could imagine it backfiring at some point if TypeScript ever type checks the JSX function or that something else uses it and there are type errors.

But that's kind of a nit pick. The `h` function is easy to re-implement if anyone likes to have a more strict alternative. I think Cycle.js does that.

I'm :+1: either way.

nojvek:
I’m not sure where we landed on this discussion. 

I can move the null data support in core h function. That would be nice. 

Moving Boolean as children in and filtering them out. I’m not sure if it would break some other consumer. 

@mightyjam had a point that bundling flatChildren into h may have a tiny perf cost for clients that directly use h. 

The current implementation makes jsx helper opt-in for clients that want to use jsx as their template language. 

I kinda like the separation of concerns. It allows us to future add some nicecities such as Jsx fragments if we wanted. 

This PR mostly leaves h, as-is allowing this to be a minor patch since there’s no breaking changes to existing core.

nojvek:
okay, I did some of the suggested tweaks. Lemme know what you think.

src/jsx.ts
```diff
@@ -1,12 +1,12 @@
...
 
-export type VNodeChildren = ArrayOrElement<VNodeChildElement | boolean>;
-export type FunctionComponent = (props: {[prop: string]: any}, children?: VNode[]) => VNode;
+export type FunctionComponent = (props: {[prop: string]: any} | null, children?: VNode[]) => VNode;
 
 function flattenAndFilter(children: VNodeChildren[], flattened: VNode[]): VNode[] {
   for (const child of children) {
-    if (child !== undefined && child !== null && child !== false) {
+    // filter out falsey children, except 0 since zero can be a valid value e.g inside a chart
+    if (child || child === 0) {
```
mightyiam: What about an empty string? Ignore or empty string node?

test/jsx.tsx
```diff
@@ -38,14 +38,16 @@ describe('snabbdom', function() {
       const showLogin = false;
       const showCaptcha = false;
       const loginAttempts = 0;
-      const userName = undefined;
+      const userName = ``;
```
mightyiam: I see below that an empty string is ignored instead of becoming an empty DOM text node. Is this the right thing to do?

RyanHirsch: In JSX an empty string is ignored. https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored

mightyiam: @RyanHirsch thank you for the input.

Seems right.

nojvek: let me change it back to `if (child !== undefined && child !== null && child !== false) {`

!child may mean too many things in JS, NaNs are also falsey and we may want that.

nojvek:
@mightyiam lemme know if you need anything else

src/jsx.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export function jsx(tag: string | FunctionComponent, data: VNodeData | null, ...children: VNodeChildren[]): VNode {
+  const flatChildren = flattenAndFilter(children, []);
+  if (typeof tag === 'function') {
+    // tag is a function component
+    return tag(data, flatChildren);
+  } else {
+    if (flatChildren.length == 1 && !flatChildren[0].sel && flatChildren[0].text) {
+      // only child is a simple text node, pass as text for a simpler vtree
+      return h(tag, data, flatChildren[0].text);
```
mightyiam: Is there a test covering this, please?

src/jsx.ts
```diff
@@ -0,0 +1,39 @@
...
+    } else {
+      return h(tag, data, flatChildren);
+    }
```
mightyiam: Would this mean the same if it were the following (same but without the `else`)?
```ts
return h(tag, data, flatChildren);
```

nojvek: Not sure if I understand, Since we have rest args `...children: VNodeChildren`, passing in a text node would always be encapsulated in an array[].

In terms of what snabbdom does, `h(tag, data, flatChildren)` would still work as h converts primitives to vnodes with text. The if just makes the vtree simpler.

src/jsx.ts
nojvek: yeah that's already covered by the first test.

https://github.com/snabbdom/snabbdom/blob/8624fe5d0b574a7dc842e7be7d718548234a54b1/test/jsx.tsx#L6-L15

and the else part is covered by the flattens test.

https://github.com/snabbdom/snabbdom/blob/8624fe5d0b574a7dc842e7be7d718548234a54b1/test/jsx.tsx#L16-L35

nojvek:
seems like force pushing to same as origin/master automatically closed the PR. 

@mightyiam the PR is squashed to a single commit now.

src/jsx.ts
mightyiam: Ideally, the tests should convey their coverage to the reader. If a future author edits this test he might alter it in a way that reduces coverage.

src/jsx.ts
mightyiam: If all previous branches `return` then there's no need for this `return` to be in an `else`. It could be after the if-else structure. But that's subject to code style preferences. Something we could possibly lint for.

.eslintignore
```diff
@@ -14,6 +14,10 @@ htmldomapi.js.map
 is.d.ts
 is.js
 is.js.map
+jsx.d.ts
+jsx-global.d.ts
+jsx.js
+jsx.js.map
```
mightyiam: Nit pick: not sorted. But this file is likely going away in #480 so don't bother.

mightyiam:
I'm not sure what you mean by "same as origin/master", @nojvek. It would be nice to understand why the force-push resulted in a close.

nojvek:
> If all previous branches return then there's no need for this return to be in an else. It could be after the if-else structure. But that's subject to code style preferences. Something we could possibly lint for.

Typescript compilation will fail if all branches don't return the same type. I am not sure what the proposed fix here is.

src/jsx.ts
nojvek: Typescript compilation will fail if all branches don't return the same type. I am not sure what the proposed fix here is.

nojvek: I think I understand what you mean. Will change to keep you happy.

nojvek: tried the fix with return outermost, kinda reads funny to be honest. May be personal preference but I like all if branches to be accounted for and this explicitly accounts for them.

nojvek:
> I'm not sure what you mean by "same as origin/master", @nojvek. It would be nice to understand why the force-push resulted in a close.

You wanted me to squash into single commit. So I soft reset to origin/master and force pushed. This resulted in this PR having exactly the same code as origin/master and the github PR auto-closed (since empty PR).

I had forgotten to commit the soft reset, once I did that, we had a commit and I could re-open

mightyiam:
> You wanted me to squash into single commit. So I soft reset to origin/master and force pushed. This resulted in this PR having exactly the same code as origin/master and the github PR auto-closed (since empty PR).
>
> I had forgotten to commit the soft reset, once I did that, we had a commit and I could re-open

I understand. Thank you for the explanation.

nojvek:
`Fab! Just squash and I'm 👍 on this PR.` I'm not sure why this is a requirement to be don't from the PR submitter side.

Github has a "Squash and merge" option right ? Isn't that better so master has squashed commit but the referenced PR has the full commit history ?

nojvek:
Awesome. Thanks @mightyiam. 

Will keep on weekly pinging @paldepind as a reminder. 

In the meantime I can work on a documentation PR on how to use snabbdom with jsx/tsx.

nojvek:
I feel we shouldn’t get into a scope creep and follow separation of concerns principle. This PR has been open for more than 3 months now. This PR will need a new release which I believe only @paldepind can do. 

Documentation shouldn’t need a new release. We can independently bikeshed about it in another PR.

src/h.ts
```diff
@@ -18,17 +18,17 @@ function addNS(data: any, children: VNodes | undefined, sel: string | undefined)
...
-export function h(sel: string, data: VNodeData): VNode;
+export function h(sel: string, data: VNodeData | null): VNode;
 export function h(sel: string, children: VNodeChildren): VNode;
-export function h(sel: string, data: VNodeData, children: VNodeChildren): VNode;
+export function h(sel: string, data: VNodeData | null, children: VNodeChildren): VNode;
 export function h(sel: any, b?: any, c?: any): VNode {
   var data: VNodeData = {}, children: any, text: any, i: number;
   if (c !== undefined) {
-    data = b;
+    if (b !== null) data = b;
```
paldepind: Please wrap the branch in braces here for consistency.

src/h.ts
paldepind: Is this addition of `boolean` a leftover from when JSX support was added  directly to`h`?

paldepind:
I still kinda like the simplicity of the original PR of making the default `h` function JSX compatible. But, I also see the argument for keeping them separate. So I agree that we should do that.

I've left a few comments, but besides those I'm 👍 for merging.

src/h.ts
nojvek: Would you be okay if we converted all primitives to string when making vnodes text properties. This would simplify a bunch of things and not have special children type for jsx vs h.

I think @mightyiam's decision behind jsx being separate is that he wanted flatChildren not to impact h calls since that may be a surface area change for h  and possibly a small perf hit.

Adding `boolean`  was going back to the original comment you made.

```
I definitely see your point. On the other hand, the childen type is already fairly liberal and the only extra thing that gets added is boolean.

Could another option be to not add boolean to the VNodeChildElement type but still support it in the code? I don't think the TypeScript JSX mechanism would care about that so it would give us the best of both worlds. h would still have the stricter type for the benefit of TypeScript developers and it will work with JSX.
```

nojvek:
@paldepind made requested changes. 

@mightyiam regarding squashing.

You can choose "squash and merge". Perhaps we should set this as the default PR merging strategy. I believe github has a setting for that.

image

mightyiam:
Thank you for the tip, @nojvek.

nojvek:
Happy Monday, Kind ping @paldepind, for review -> approve -> merge -> npm publish please.

paldepind:
It whould be nice to have documentation before we release such that the release notes can point towards the documentation.

nojvek:
Awesome. Thanks @paldepind. I’m on it.

strass:
Hey, did any documentation on how to use snabbdom with typescript jsx ever get written? I'm about to look into snabbdom-jsx but h seems so close to compatible...

RyanHirsch:
> Hey, did any documentation on how to use snabbdom with typescript jsx ever get written?

I think this is the documentation
https://github.com/snabbdom/snabbdom/pull/494
But it is stalled like this one.

nojvek:
@paldepind hasn't published what has landed into master. He's been AWOL since last year and snabbdom is somewhat in a state of limbo. He's the only one that can publish or merge major PRs.

The best course of action is to make another npm module because it's unlikely it will land in core snabbdom anytime soon. I'm at the end of my patience.  Will update this thread when npm module is ready.

strass:
huh, that's really quite a shame. its literally a 1 line change to get it working for me.

please let me know if you end up forking the module.

nojvek:
This is where I’m working on it. https://github.com/nojvek/snabbdom-tsx

Snabbdom-tsx v1 should be out by next week with docs, tests and an example.

nojvek:
@strass I've moved the relevant jsx bits to a new repo https://github.com/nojvek/snabbdom-jsx-lite

It's published on npm. It supports `JSX`, `JSX Fragments`, `sel` attribute for css selector shorthand. 

Use that instead. Even though this PR is showing merged, there's about 5 months of pending changes that haven't been published to npm.

strass:
🙌 thanks! perfect timing I'll give it a shot this weekend.

strass:
Looks good but I'm having trouble with SVGs. This creates the nodes and the polygon even has styles on it, but the attributes don't show up (nor do they when I use props):

```
    <svg
      attrs={{
        height: "250",
        width: "450",
      }}
    >
      <polygon
        attrs={{
          points: "225,10 100,210 350,210",
        }}
        style={{
          fill: "white",
          stroke: "black",
        }}
      />
</svg>
```

nojvek:
Yeah let’s take the discussion there. Did you init the snabbdom modules?

There’s an svg example you can follow https://github.com/nojvek/snabbdom-jsx-lite/blob/master/example/app.tsx

PR #452, opened 2019-09-15
chore: update nodejs versions

mightyiam:


mightyiam:
@paldepind this should be a no-brainer. Could possibly review?

mightyiam:
Currently, tests are running only in browsers (via karma). We should probably run tests in Node.js, as well, because snabbdom is meant to run in Node.js, as well, right? For server side rendering. But until we actually do that, we might as well upgrade the Node.js version to something modern, to allow modern tooling, like ESLint, which [dropped support for Node.js v6](https://github.com/eslint/eslint/commit/6e7da57dddc41830df4aee77e31c4320c1557350).

paldepind:
Looks good!

> Currently, tests are running only in browsers (via karma). We should probably run tests in Node.js, as well, because snabbdom is meant to run in Node.js, as well, right? For server side rendering.

I'm not sure about that. Most of Snabbdom depends on the DOM and doesn't really work without it. Additionally, the whole "diff and patch" process isn't necessary when doing server side rendering because you only render a single view once and then send it off. It seems to me that for SSR one would want to use [snabbdom-to-html](https://github.com/snabbdom/snabbdom-to-html) instead.

mightyiam:
Thank you for the correction and the review, @paldepind.

PR #454, opened 2019-09-15
chore: soft introduce linting

mightyiam:

PR #455, opened 2019-09-15
chore: editorconfig + checker

mightyiam:


mightyiam:
Based on #454.

tsconfig.configs.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "noEmit": true
+  },
+  "files": [
+    "./karma.conf.js",
+    "./gulpfile.js"
+  ]
+}
```
paldepind: What is the purpose of this tsconfig file?

mightyiam: Since this is from #454, I'll answer there.

mightyiam: Does not exist any more.

PR #456, opened 2019-09-18
style: fix package.json husky field indent

mightyiam:


mightyiam:
Trivial, so merging in spite of #457.

PR #474, opened 2019-11-13
ci: disable failing browsers

### LINKED ISSUE #428: Development workflow seems broken
The expected workflow:
1. `git clone`
1. Select Node.js version that the project expects
1. `npm install --global npm`
1. `npm install`
1. `npm test`

Fails.

I would say that this issue depends on

- [x] #422
- [x] #424
- [x] #426
- [x] #322
- [x] #460
- [ ] #473

### LINKED ISSUE #473: CI fails on some browsers
- #472 
- #471 
- #470 
- #469 
- #468

mightyiam:
- Fixes #473
- Fixes #428

### DIFF
--- a/browserstack-karma.js
+++ b/browserstack-karma.js
@@ -31,6 +31,7 @@ module.exports = {
   },
 
   // Older mainstream
+  /* https://github.com/snabbdom/snabbdom/issues/468
   BS_Chrome_49: {
     base: 'BrowserStack',
     browser: 'chrome',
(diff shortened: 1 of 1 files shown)


browserstack-karma.js
```diff
@@ -31,27 +31,31 @@ module.exports = {
   },
 
   // Older mainstream
+  /*
```
nojvek: perhaps a comment on why they are disabled.

browserstack-karma.js
```diff
@@ -31,27 +31,31 @@ module.exports = {
   },
 
   // Older mainstream
+  /* https://github.com/snabbdom/snabbdom/issues/468
```
mightyiam: @nojvek I agree. Here are comments linking to specific issues. Details on those failures could be added to those issues later (or just solved).

PR #479, opened 2019-11-14
fix: npm script prepublish -> prepublishOnly

### LINKED ISSUE #476: prepublish script is deprecated since npm@5
```
npm WARN prepublish-on-install As of npm@5, `prepublish` scripts are deprecated.
npm WARN prepublish-on-install Use `prepare` for build steps and `prepublishOnly` for upload-only.
```

mightyiam:
Fixes #476.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "compile": "npm run compile-es && npm run compile-commonjs",
     "compile-es": "tsc --outDir es --module es6 --moduleResolution node",
     "compile-commonjs": "tsc --outDir ./",
-    "prepublish": "npm run compile",
+    "prepublishOnly": "npm run compile",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
     "compile": "npm run compile-es && npm run compile-commonjs",
     "compile-es": "tsc --outDir es --module es6 --moduleResolution node",
     "compile-commonjs": "tsc --outDir ./",
-    "prepublish": "npm run compile",
+    "prepublishOnly": "npm run compile",
```
mightyiam: While compiling is not the most thorough set of checks we could run before publishing, this change just gets us up to date with modern npm.

https://docs.npmjs.com/misc/scripts#prepublish-and-prepare

Issue #481, opened 2019-11-14
How about not tracking `dist` in VCS?


mightyiam:
I feel against tracking build artifacts in VCS.

mreinstein:
`prepublishOnly` npm event is probably the right place to generate the dist directory

kuraga:
IMHO `type:bug` -> `type:enhancement` should be relabeled...

mreinstein:
depends on how you define an enhancement. This particular task is probably not much more work than would be required to adjust the labeling of it. :) Might be time better spent to just complete it.

PR #482, opened 2019-11-14
v0.7.5

mightyiam:


mightyiam:
The previous release PR, #477 did not include a UMD rebuild. And it kinda seems like some releases prior to that also neglected to do that. This PR fixes that. And in this context, #481.

nojvek:
where is the umd support being added ? I only see dist files

mightyiam:
@nojvek, the files in `dist` seem to be `UMD`. See here:

https://github.com/snabbdom/snabbdom/blob/4b19aebd5e82eb8b7fc93946798bad6b8df3ad99/gulpfile.js#L9-L13

See [the `--standalone` flag here](https://github.com/browserify/browserify#usage).

PR #485, opened 2019-11-19
style: @typescript-eslint/array-type

mightyiam:


.eslintrc.js
```diff
@@ -2,7 +2,6 @@ module.exports = {
   extends: 'standard-with-typescript',
   parserOptions: { project: [ './tsconfig.json', ] },
   rules: {
-    '@typescript-eslint/array-type': 'off',
```
mightyiam: - [Rules docs](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/array-type.md)
- [Our rule options](https://github.com/standard/eslint-config-standard-with-typescript/blob/4bfcb96fdd1820a6b3d952a38655ad9d681acf27/src/index.ts#L46)

paldepind:
Looks good to me. I prefer that style as well.

PR #488, opened 2019-11-22
docs: readme point on contributing

mightyiam:


paldepind:
I agree with this, but only partly. For instance, #451 is exactly new API surface and (in its current state where it no longer changes `h`) it could very well be a third party module as far as I can tell.

There have definitely been suggestions over the years where which I did not find appropriate to include. But I don't think it is a big enough problem that we need this disclaimer in the readme. It's better to get too many than too few suggestions, right? :smile:

nojvek:
I'd kinda agree with @paldepind here. From my understanding it's not strictly about api surface area but about separation of concerns.

I view snabbdom as a lightweight and fast virtual dom library. You can use it with vanillajs, web components, some other library, or good ol jquery. It takes vnodes and patches the real dom. 💥

But how it does that we should be open. e.g may be a wasm differ that's insanely fast, or very popular ways of having the dom representations (jsx and hyperscript). Or have snabbdom be agnostic to dom and can target native apps too. snabbdom allows custom modules so this is possible too.

like using snabbdom + some library to build cli apps. I gonna spend a decent amount of  time writing all sorts of good examples and tutorials to preach snabbdom.

PR #489, opened 2019-11-25
Update TypeScript, refactor

paldepind:
Updates dependencies to the most recent version of TypeScript.

Tweak/refactor a bunch of things in `snabbdom.ts`. Most of the changes are related to using new features in JavaScript/TypeScript.

I won't go through each change but please comment if there is anything that does not look like an improvement :smile:

### DIFF
--- a/src/snabbdom.ts
+++ b/src/snabbdom.ts
@@ -1,12 +1,13 @@
 /* global module, document, Node */
 import {Module} from './modules/module';
-import {Hooks} from './hooks';
-import vnode, {VNode, VNodeData, Key} from './vnode';
+import vnode, {VNode} from './vnode';
 import * as is from './is';
 import htmlDomApi, {DOMAPI} from './htmldomapi';
 
+type NonUndefined<T> = T extends undefined ? never : T;
+
 function isUndef(s: any): boolean { return s === undefined; }
-function isDef(s: any): boolean { return s !== undefined; }
+function isDef<A>(s: A): s is NonUndefined<A> { return s !== undefined; }
 
 type VNodeQueue = VNode[];
 
@@ -29,12 +30,11 @@ type ArraysOf<T> = {
 type ModuleHooks = ArraysOf<Module>;
 
 function createKeyToOldIdx(children: VNode[], beginIdx: number, endIdx: number): KeyToIndexMap {
-  let i: number, map: KeyToIndexMap = {}, key: Key | undefined, ch;
-  for (i = beginIdx; i <= endIdx; ++i) {
-    ch = children[i];
-    if (ch != null) {
-      key = ch.key;
-      if (key !== undefined) map[key] = i;
+  const map: KeyToIndexMap = {};
+  for (let i = beginIdx; i <= endIdx; ++i) {
+    const key = children[i]?.key;
+    if (key !== undefined) {
+      map[key] = i;
     }
   }
   return map;
@@ -78,8 +78,9 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
   function createElm(vnode: VNode, insertedVnodeQueue: VNodeQueue): Node {
     let i: any, data = vnode.data;
     if (data !== undefined) {
-      if (isDef(i = data.hook) && isDef(i = i.init)) {
-        i(vnode);
+      const init = data.hook?.init;
+      if (isDef(init)) {
+        init(vnode);
         data = vnode.data;
       }
     }
@@ -88,7 +89,7 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
       if (isUndef(vnode.text)) {
         vnode.text = '';
       }
-      vnode.elm = api.createComment(vnode.text as string);
+      vnode.elm = api.createComment(vnode.text!);
     } else if (sel !== undefined) {
       // Parse selector
       const hashIdx = sel.indexOf('#');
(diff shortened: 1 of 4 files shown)


src/snabbdom.ts
```diff
@@ -112,13 +113,15 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
       } else if (is.primitive(vnode.text)) {
         api.appendChild(elm, api.createTextNode(vnode.text));
       }
-      i = (vnode.data as VNodeData).hook; // Reuse variable
-      if (isDef(i)) {
-        if (i.create) i.create(emptyNode, vnode);
-        if (i.insert) insertedVnodeQueue.push(vnode);
+      const hook = vnode.data!.hook;
+      if (isDef(hook)) {
+        hook.create?.(emptyNode, vnode);
```
mightyiam: Look at you using the latest features right out of the oven :medal_sports:

src/snabbdom.ts
```diff
@@ -112,13 +113,15 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
       } else if (is.primitive(vnode.text)) {
         api.appendChild(elm, api.createTextNode(vnode.text));
       }
-      i = (vnode.data as VNodeData).hook; // Reuse variable
-      if (isDef(i)) {
-        if (i.create) i.create(emptyNode, vnode);
-        if (i.insert) insertedVnodeQueue.push(vnode);
+      const hook = vnode.data!.hook;
```
mightyiam: Yes. Frankly, I find the reuse of variables hard to follow. I'm happy you're removing it.

src/snabbdom.ts
```diff
@@ -140,15 +143,15 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
-          i = vnode.children[j];
-          if (i != null && typeof i !== "string") {
-            invokeDestroyHook(i);
+        for (let j = 0; j < vnode.children.length; ++j) {
+          const child = vnode.children[j];
+          if (child != null && typeof child !== "string") {
+            invokeDestroyHook(child);
           }
         }
       }
```
mightyiam: Remove variable reuse and use optional chaining. :+1:

src/snabbdom.ts
```diff
@@ -160,20 +163,21 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
           listeners = cbs.remove.length + 1;
-          rm = createRmCb(ch.elm as Node, listeners);
-          for (i = 0; i < cbs.remove.length; ++i) cbs.remove[i](ch, rm);
-          if (isDef(i = ch.data) && isDef(i = i.hook) && isDef(i = i.remove)) {
-            i(ch, rm);
+          rm = createRmCb(ch.elm!, listeners);
+          for (let i = 0; i < cbs.remove.length; ++i) cbs.remove[i](ch, rm);
+          const removeHook = ch?.data?.hook?.remove;
+          if (isDef(removeHook)) {
+            removeHook(ch, rm);
```
mightyiam: The current variable reuse pattern, is it for performance or for less code or what, please?

In any case, this :+1:

src/snabbdom.ts
```diff
@@ -254,39 +258,34 @@ export function init(modules: Array<Partial<Module>>, domApi?: DOMAPI) {
...
-      i(oldVnode, vnode);
-    }
-    const elm = vnode.elm = (oldVnode.elm as Node);
-    let oldCh = oldVnode.children;
-    let ch = vnode.children;
+    const hook = vnode.data?.hook;
+    hook?.prepatch?.(oldVnode, vnode);
+    const elm = vnode.elm = oldVnode.elm!;
+    let oldCh = oldVnode.children as VNode[];
+    let ch = vnode.children as VNode[];
```
mightyiam: Assert earlier is better.

src/snabbdom.ts
```diff
@@ -1,12 +1,13 @@
 /* global module, document, Node */
 import {Module} from './modules/module';
-import {Hooks} from './hooks';
```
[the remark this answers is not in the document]
mightyiam: So, due to replacing of type assertions with optional chaining, the type reference is not necessary. Cool.

src/snabbdom.ts
```diff
@@ -1,12 +1,13 @@
...
-import vnode, {VNode, VNodeData, Key} from './vnode';
```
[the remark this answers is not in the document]
mightyiam: Same reason as above, probably.

src/snabbdom.ts
```diff
@@ -1,12 +1,13 @@
...
+import vnode, {VNode} from './vnode';
 import * as is from './is';
 import htmlDomApi, {DOMAPI} from './htmldomapi';
 
+type NonUndefined<T> = T extends undefined ? never : T;
```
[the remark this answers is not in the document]
mightyiam: So, TypeScript has a built-in type, `NonNullable<T>`. Are you aware of it? If you are, why did you choose to write this (admittedly more specific) type, instead of using it, please?

src/snabbdom.ts
paldepind: 😄

src/snabbdom.ts
paldepind: Mostly for less code. In particular with the deep access of properties that might be undefined along the way, I found it to looks nice back when I wrote Snabbdom.

But we agree that it confusing and somewhat off-putting. We could probably cut back on it even more.

src/snabbdom.ts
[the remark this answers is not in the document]
paldepind: I did it this way because `NonNullable` removes both `undefined` and `null` but the code in `isUndef` where the type is used only checks for `undefined`. Maybe we could get away with using `NonNullable` but I thought it would be best to have the type reflect exactly what the code actually does.

paldepind:
Thanks for the comments @mightyiam. Btw, feel free to merge my PRs. It's safe to assume I'm okay with that :wink:

There are many places in Snabbdom where we access nested properties that may be undefined along the way. Now that those have been replaced by optional chaining (`?.`) we could consider removing the `isUndef` function altogether? Its benefits are less clear now.

PR #491, opened 2019-11-30
Broader type signature for event-handlers

### LINKED ISSUE #431: The typescript declaration for eventlisteners only supports function listeners
The TS def in eventlisteners.d.ts is

```
export type On = {
  [N in keyof HTMLElementEventMap]?: (ev: HTMLElementEventMap[N]) => void
} & {
  [event: string]: EventListener
};
```
which is much to limiting.

A temporary fix is to change it into 

```
export type On = { 
    [event: string]: any[] | ((...args:any[]) => void);
}
```

which may be a little bit to broad but at least it works.

_Originally posted by @kay999 in https://github.com/snabbdom/snabbdom/issues/391#issuecomment-478296040_

kay999:
Fixes #431

### DIFF
--- a/src/modules/eventlisteners.ts
+++ b/src/modules/eventlisteners.ts
@@ -1,10 +1,10 @@
 import {VNode, VNodeData} from '../vnode';
 import {Module} from './module';
 
-export type On = {
-  [N in keyof HTMLElementEventMap]?: (ev: HTMLElementEventMap[N]) => void
-} & {
-  [event: string]: EventListener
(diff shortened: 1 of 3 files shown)


mightyiam:
@kay999, tests were converted to TypeScript. This should make it possible to test the types — which we would like in this PR. Would you care to rebase this branch?

mightyiam:
Thank you again for the pull request, @kay999. I hope the issue is resolved to your satisfaction.

PR #494, opened 2019-12-04
jsx documentation + example

nojvek:

PR #502, opened 2019-12-17
feat(jsx): sel property

nojvek:


nojvek:
@mightyiam split the jsx-global.d.ts file move from #494

test/jsx.tsx
```diff
@@ -36,6 +36,16 @@ describe('snabbdom', function() {
...
+    it('supports sel property shorthand', function() {
+      const vnode = <div sel="#id.foo.bar"></div>;
+
+      assert.equal(JSON.stringify(vnode), JSON.stringify({
+        sel: 'div#id.foo.bar',
+        data: {},
+        children: [],
+      }));
+    });
+
```
mightyiam: I don't take it for granted that this should be supported. I'm thinking that since snabbdom JSX is so different than React JSX (there is no `<h href="foo">foo</h>`, you have to do `<h attrs={{ href: 'foo' }}/>foo</h>`), we have the freedom to let this `h` feature reach JSX-land. snabbdom JSX attributes are sharing HTML attribute namespace. I rather like that. It saves us a lot of trouble React code is dealing with due to this. OK. I'm OK with this.

jsx-global.d.ts
```diff
@@ -8,7 +9,7 @@ declare global {
   namespace JSX {
     type Element = VNode;
     interface IntrinsicElements {
-      [elemName: string]: VNodeData;
+      [elemName: string]: JsxVNodeProps;
     }
   }
 }
```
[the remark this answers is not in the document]
nojvek: so as a library user you can simply use it as `snabbdom/jsx-global` inside tsconfig rather than having to go into `snabbdom/src/jsx-global`. Just looks neater as a user since jsx pragma itself is imported as `snabbdom/jsx`

test/jsx.tsx
nojvek: I'm not sure what to make of this comment. The reasoning behind having sel was developer UX. 

It's much easier and readable to write

```jsx
<div sel=".profile">My Name</div>
```

and equivalent css has

```
.profile {
 // some css props
}
```

makes it easier to just grep for css selectors and see both style/dom tree come up in search.

rather than

```
<div class={{profile: true}}>My Name</div>
// or 
<div attrs={{class: `profile`}}>My Name</div>
```

Plus `h('tag#id.class')` is optimized because snabbdom only does id class attribute adding at tag create time and doesn't diff it after, so its a minor perf win.

Re: snabbdom namespacing, I really love it. It's better than react because react can only do props. With snabbdom you can have custom modules for animations, hooks, datasets, listen to custom events e.t.c

jsx-global.d.ts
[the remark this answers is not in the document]
mightyiam: If it is in `src/jsx-global.d.ts` and it is included in the `files` field in `tsconfig.json`, I expect it to be copied to the root by `tsc` in the CommonJS build. Isn't this so?

test/jsx.tsx
mightyiam: I apologize. I meant to write "snabbdom JSX attributes are *not* sharing HTML attribute namespace". I'm alright with this change. Your reasoning and examples are great.

jsx-global.d.ts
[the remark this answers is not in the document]
nojvek: Not quite. The jsx-global.d.ts here provides type info for input files. That doesn’t get copied. Only the auto generated .d.ts files from input files are outputted in root folder. Basically for outDir compiler option. Tsc makes a .d.ts and a .js for every .ts file.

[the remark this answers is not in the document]
mightyiam: [Right](https://stackoverflow.com/a/56440335/359072). How about rename it to `src/jsx-global.ts` as suggested in that answer?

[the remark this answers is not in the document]
nojvek: I’m not sure what you mean. Do you want the import to still be `snabbdom/src/jsx-global` ?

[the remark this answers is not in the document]
mightyiam: If it is renamed to a `.ts` file, `src/jsx-global.ts` then it will be compiled and end up in the two build outputs. Just make sure it's in the `tsconfig.json` `files` field. Do you like this idea?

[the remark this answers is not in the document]
nojvek: if we do that, then tsc will generate an empty jsx-global.js, and we'll get many copies of jsx-global.d.ts. One in root, one in /es and one in src. Feels uglier. Honestly I don't have strong opinions. If you feel it should really be `src/jsx-global.d.ts` I can revert those changes and move it back again.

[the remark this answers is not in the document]
mightyiam: I appreciate your willingness to change according toy preference. But let me explain why I prefer it that way.

It seems simpler if this file is considered source code and lives with all the other source code. There is no "Oh, and there's this other file outside of `src`. That feels like less complexity to me.

Regarding the multiple useless build artifacts I don't feel that's an issue at all. Just ignore them in git and include in the package only what should be. It's a whitelist anyway.

nojvek:
Happy New Decade @mightyiam , Take 2

nojvek:
kind ping @paldepind / @mightyiam ^

src/jsx.ts
```diff
@@ -27,17 +27,22 @@ function flattenAndFilter (children: JsxVNodeChildren[], flattened: VNode[]): VN
...
   const flatChildren = flattenAndFilter(children, []);
   if (typeof tag === 'function') {
     // tag is a function component
-    return tag(data, flatChildren);
+    return tag(props, flatChildren);
   } else {
+    // append sel css selector to tag to support equivalent of h('span.foo.bar')
+    if (props && props.sel) {
+      tag += props.sel;
+      delete props.sel;
```
mightyiam: This mutates the provided object. What if we didn't delete it?

nojvek:
@mightyiam merged with master and review feedback, ready for another pass/merge

mightyiam:
According to our policy, this requires approval from @paldepind, right?

nojvek:
its a super minor thing tho, policy says major things right?

anyways @paldepind , please ^

mightyiam:
@nojvek seems to me that it will be a commonly used part of the JXS feature.

And the way I interpret the policy is anything "more substantial" than

> Maintenance, documentation, configuration, linting, dependencies, bug fixes, etc.

Requires @paldepind.

src/test/jsx.tsx
```diff
@@ -36,6 +36,16 @@ describe('snabbdom', function() {
...
     });
 
+    it('supports sel property shorthand', function() {
+      const vnode = <div sel="#id.foo.bar"></div>;
+
+      assert.equal(JSON.stringify(vnode), JSON.stringify({
+        sel: 'div#id.foo.bar',
+        data: {},
+        children: [],
+      }));
```
mightyiam: There's `assert.deepStrictEqual`. Perhaps we should use that instead of JSON.

nojvek: I did try deepStrictEqual first, but deepStrictEqual looks at the undefined props as well and that fails. JSON.stringify is agnostic to undefined properties so our assert code can be smaller.

Also stringify gives a better error UX  because you see the full obj. deepStrictEquals only gives partial and wasn't as useful at first glance.

mightyiam: Thank you fro the explanation. What's different about this test that it fails with `undefined` values? I'm at this moment converting all `deepEqual` to `deepStrictEqual` and nothing fails.

nojvek: Oh cool. Didn’t know it was fixed. Will fix then.

kuraga: Sorry but which `deepStrictEqual` do we use? In NodeJS, obviously, from standard `assert`. But in browser?

mightyiam: @kuraga... good question. I don't know. It's a mystery.

kuraga: See https://github.com/snabbdom/snabbdom/pull/531#issuecomment-572177936, please.

src/jsx.ts
```diff
@@ -27,17 +27,22 @@ function flattenAndFilter (children: JsxVNodeChildren[], flattened: VNode[]): VN
...
   const flatChildren = flattenAndFilter(children, []);
   if (typeof tag === 'function') {
     // tag is a function component
-    return tag(data, flatChildren);
+    return tag(props, flatChildren);
   } else {
+    // append sel css selector to tag to support equivalent of h('span.foo.bar')
+    if (props && props.sel) {
+      tag += props.sel;
+      props.sel = undefined;
```
mightyiam: While this does not delete the property, it still mutates the provided object. Is there a problem with leaving it untouched?

nojvek: we could, but the test diff looks off because the vnode has {sel, props: {sel...}, children: ..}

I could make another object but that seemed like unnecessary perf hit.

usually for jsx, the props are generated by the transpiler which would end up as vnode objects anyway so it seemed okay modifying them before passing to underlying h which would generate the vnode.

Happy to change it if you feel really strongly, but I didn't consider mutating sel to be a no-no since it is a jsx specific key only.

mightyiam: I suppose leaving it as it is would be the least performance hit. And how the tests look doesn't seem to be a reason to change implementation, to me. I don't know if this is a hot path or not. I think it might be. So maybe we should see that as a reason to not mutate.

nojvek:
@mightyiam / @paldepind wanna have another look please ?

src/test/jsx.tsx
```diff
@@ -36,6 +36,18 @@ describe('snabbdom', function () {
...
+      const vnode = <div sel="#id.foo.bar"></div>;
+
+      assert.strictEqual(JSON.stringify(vnode), JSON.stringify({
+        sel: 'div#id.foo.bar',
+        data: {
+          sel: '#id.foo.bar',
+        },
+        children: [],
+      }));
+    });
```
mightyiam: `assert.deepStrictEqual` now that #531 is in.

nojvek:
@mightyiam 👀 please.

I was trying to commit from Nairobi, Kenya (internet is quite slow here). It seems we have a ton of git commit hooks (editorconfig was getting stuck, karma couldn't download firefox). Pretty terrible experience.

Had to commit with `git commit --no-verify".

Didn't quite realize how hostile working from Africa is.

mightyiam:
@paldepind :bell: :bulb: :zap: :fireworks: :pray:

jbwyme:
Would love to see this merged and a new snabbdom release cut :D

jbwyme:
Hey folks, any update here? 

Thanks for Snabbdom btw, love it.

nojvek:
current status is that feat jsx is blocked in @paldepind. 

Landing jsx has been more than a 3 month journey. I guess if @paldepind is still AWOL for another week or two, we can make a new npm package.

jbwyme:
any luck getting to this @paldepind ?

nojvek:
@paldepind kind ping. I hope you're keeping safe with the ongoing pandemic. We've been waiting on you for 4 months.  Would love to see this land before the virus gets to me.

nojvek:
Another kind ping @paldepind ^

nojvek:
Closing this PR: Moved all the relevant JSX bits to a new npm module

See https://github.com/nojvek/snabbdom-jsx-lite

Supports sel, jsx, jsx fragments and has a perf test to ensure it remains blazing fast.

PR #508, opened 2019-12-19
refactor(tests): convert to TypeScript

### LINKED ISSUE #293: Convert tests to TypeScript
Following up from https://github.com/mightyiam/snabbdom/commit/f3354732bf95103322fa1610e64fb4c71332f199#commitcomment-22137381.

### LINKED ISSUE #509: Karma coverage seems to include tests under Windows
https://github.com/snabbdom/snabbdom/blob/aac0233b1ff616ffbb47d06c39505fad0ec50476/karma.conf.js#L46

According to documentation, it seems that this would result in test files to be included in coverage under Windows:
https://github.com/monounity/karma-typescript/tree/master/packages/karma-typescript#advanced-configuration

> beware that Windows uses \ while UNIX-like systems use / as path separator.

mightyiam:
Closes #293.

Fixes #509.


src/test/attachto.ts
```diff
@@ -1,12 +1,13 @@
...
-var patch = snabbdom.init([]);
-var attachTo = require('../helpers/attachto').default;
-var h = require('../h').default;
+var patch = init([]);
+import attachTo from '../helpers/attachto'
+import h from '../h'
 
 describe('attachTo', function() {
-  var elm, vnode0;
+  var elm: any, vnode0: any;
```
mightyiam: This pattern of reusing these two variables is in all test modules. These definitely should be refactored, I feel. Probably removed and replaced with variables/constants in each test. But beyond the scope of this PR.

src/test/core.ts
```diff
@@ -311,7 +322,8 @@ describe('snabbdom', function() {
         prevElm.id = 'id';
         prevElm.className = 'class';
         var text = new Text('Foobar');
-        text.testProperty = function () {}; // ensures we dont recreate the Text Node
+        const reference = {};
+        (text as any).testProperty = reference; // ensures we dont recreate the Text Node
```
mightyiam: Is there a reason this is specifically a function and not any other reference? In TypeScript, our ESLint does complain about empty functions. So I changed this to an empty object. Cool?

src/test/core.ts
```diff
@@ -323,7 +335,7 @@ describe('snabbdom', function() {
         assert.strictEqual(elm.childNodes.length, 1);
         assert.strictEqual(elm.childNodes[0].nodeType, 3);
         assert.strictEqual(elm.childNodes[0].wholeText, 'Foobar');
-        assert.strictEqual(typeof elm.childNodes[0].testProperty, 'function');
+        assert.strictEqual(elm.childNodes[0].testProperty, reference);
```
mightyiam: And why is this a `typeof` check and not a reference check, please? Perhaps no reason?

src/test/core.ts
```diff
@@ -842,13 +856,13 @@ describe('snabbdom', function() {
...
+        function preCb() {
+          pre++
         }
-        function postCb(oldVnode, vnode) {
-          assert.equal(pre.length, post.length + 1);
-          post.push(post);
+        function postCb() {
+          assert.equal(pre, post + 1);
+          post++
         }
```
mightyiam: I found this one rather unintuitive. It seems to push an array reference into itself. Just for the purpose of counting. So I changed it to counting using numbers to simplify the types involved. Did I miss anything?

src/test/core.ts
```diff
@@ -968,19 +982,19 @@ describe('snabbdom', function() {
...
         elm = patch(vnode1, vnode2).elm;
         assert.equal(elm.children.length, 1);
-        rm1();
+        (rm1 as any)();
         assert.equal(elm.children.length, 1);
-        rm3();
+        (rm3 as any)();
         assert.equal(elm.children.length, 1);
-        rm2();
+        (rm2 as any)();
```
mightyiam: I gave up trying to understand this test because I don't have to for now. It most likely should be written in a way that doesn't upset TypeScript (used before defined).

src/test/core.ts
```diff
@@ -1105,9 +1119,9 @@ describe('snabbdom', function() {
     });
     it('does not update strictly equal children', function() {
       var result = [];
-      function cb(vnode) { result.push(vnode); }
+      function cb(vnode: VNode) { result.push(vnode); }
       var vnode1 = h('div', [
-        h('span', {hook: {patch: cb}}, 'Hello'),
+        h('span', {hook: {patch: cb} as any}, 'Hello'),
```
mightyiam: This seems to be an undocumented hook. `patch`. :man_shrugging: .

src/test/dataset.ts
```diff
@@ -1,15 +1,14 @@
-var assert = require('assert');
-var fakeRaf = require('fake-raf');
```
mightyiam: This import seems to be in this module for no reason.

src/test/eventlisteners.ts
```diff
@@ -38,25 +40,25 @@ describe('event listeners', function() {
...
+    function clicked(ev: Event) { result.push(ev); }
+    var vnode1 = h('div', {on: {click: [clicked, 1] as any}}, [
       h('a', 'Click my parent'),
     ]);
-    var vnode2 = h('div', {on: {click: [clicked, 2]}}, [
+    var vnode2 = h('div', {on: {click: [clicked, 2] as any}}, [
       h('a', 'Click my parent'),
     ]);
-    var vnode3 = h('div', {on: {click: [clicked, 3]}}, [
+    var vnode3 = h('div', {on: {click: [clicked, 3] as any}}, [
```
mightyiam: It seems that due to #431 type checking must be disabled here. And a few similar cases below.

src/test/style.ts
```diff
@@ -1,16 +1,15 @@
-var assert = require('assert');
-var fakeRaf = require('fake-raf');
```
mightyiam: See further below why this is removed.

src/test/style.ts
```diff
@@ -1,16 +1,15 @@
...
+import assert from 'assert'
 
-var snabbdom = require('../snabbdom');
-fakeRaf.use();
```
mightyiam: When using `require` calls, with whatever karma uses as a module system, this `.use()` call occurs before the style module is evaluated. But converting this to TypeScript and static imports, the only way to have the `.use()` call occur before the import, seems to be using dynamic `import()` call. If we had top-level `await` then it would have been reasonable. But [TypeScript doesn't support that yet](https://github.com/microsoft/TypeScript/issues/25988).

So I removed this sweet little tool by @paldepind and replaced it with actual calls to the real `requestAnimationFrame`. See the test below.

src/test/style.ts
```diff
@@ -67,9 +66,9 @@ describe('style', function() {
...
   });
   it('updates css variables', function() {
-    var vnode1 = h('div', {style: {'--myVar': 1}});
-    var vnode2 = h('div', {style: {'--myVar': 2}});
-    var vnode3 = h('div', {style: {'--myVar': 3}});
+    var vnode1 = h('div', {style: {'--myVar': 1 as any}});
+    var vnode2 = h('div', {style: {'--myVar': 2 as any}});
+    var vnode3 = h('div', {style: {'--myVar': 3 as any}});
     elm = patch(vnode0, vnode1).elm;
     assert.equal(elm.style.getPropertyValue('--myVar'), 1);
```
mightyiam: I wonder why these even use numbers instead of strings.

What I find interesting is that they come out as numbers on the other side, as well.

src/test/style.ts
```diff
@@ -89,44 +88,46 @@ describe('style', function() {
...
+        elm = patch(vnode1, vnode2).elm;
+        assert.equal(elm.style.fontSize, '18px');
+        requestAnimationFrame(() => {
+          requestAnimationFrame(() => {
+            assert.equal(elm.style.fontSize, '20px');
+            done()
+          })
+        })
+      })
+    })
```
mightyiam: I little bit more indented, but works. One thing I don't understand is why it takes two calls to read the change. Anyone?

tsconfig.json
```diff
@@ -9,6 +9,7 @@
         "noUnusedLocals": true,
         "jsx": "react",
         "jsxFactory": "jsx",
+        "esModuleInterop": true,
```
mightyiam: One of the imports required this. I don't even remember. Whatever. Should be on anyway.

tsconfig.json
```diff
@@ -25,6 +26,15 @@
...
         "src/modules/style.ts",
+        "src/test/attachto.ts",
+        "src/test/attributes.ts",
+        "src/test/core.ts",
+        "src/test/dataset.ts",
+        "src/test/eventlisteners.ts",
+        "src/test/htmldomapi.ts",
+        "src/test/jsx.tsx",
+        "src/test/style.ts",
+        "src/test/thunk.ts",
```
mightyiam: Hey, TypeScript, please compile these, as well.

mightyiam:
This should open the door to solving some type issues we have with confidence, because the code changes would be tested in the tests. Because the tests would be TypeScript and not just JavaScript. Forgive my English. It's 4:29 am.

src/test/attachto.ts
kuraga: Open a new issue?

kuraga:
Just as a side note: [https://github.com/thunks/tman](T-man): Super test manager for JavaScript. (Lightweight.)

PR #511, opened 2019-12-20
chore(tsc): simplify tsconfig.json using tsconfigs

mightyiam:


karma.conf.js
```diff
@@ -46,6 +46,7 @@ module.exports = function(config) {
         exclude: /test\//,
       },
       compilerOptions: {
+        module: 'commonjs',
```
mightyiam: In `tsconfigs/browser-module` the value for `module` is `ESNEXT`. I don't know what Karma's module system does or expects, but it seems that it does not expect ES modules. So CommonJS it is.

package.json
```diff
@@ -40,6 +40,7 @@
     "npm-run-all": "^4.1.5",
     "remark-cli": "^7.0.1",
     "remark-toc": "^6.0.0",
+    "tsconfigs": "^4.0.2",
```
mightyiam: https://www.npmjs.com/package/tsconfigs brought to you by some random developer. He probably has good intentions.

package.json
```diff
@@ -52,8 +53,8 @@
     "unit": "karma start",
     "test": "run-s lint compile unit",
     "compile": "npm run compile-es && npm run compile-commonjs",
-    "compile-es": "tsc --outDir es --module es6 --moduleResolution node",
```
mightyiam: `moduleResolution: 'node'` is set in `tsconfigs`.

package.json
```diff
@@ -52,8 +53,8 @@
...
-    "compile-commonjs": "tsc --outDir ./",
+    "compile-es": "tsc --outDir es --module es6",
+    "compile-commonjs": "tsc --outDir . --module commonjs",
```
mightyiam: [`tsconfigs` has `module: 'ESNEXT'`](https://github.com/mightyiam/tsconfigs/blob/1be54f2c1b2384828ced56327f77419e8a2a37a9/browser-module.json#L5).

tsconfig.json
```diff
@@ -1,19 +1,11 @@
 {
+    "extends": "tsconfigs/browser-module",
     "compilerOptions": {
+        "esModuleInterop": false,
```
mightyiam: If we let `tsconfigs` set this to `true`, we end up with a slightly different CommonJS output. [See this](https://stackoverflow.com/a/56348146/359072).

tsconfig.json
```diff
@@ -1,19 +1,11 @@
...
+        "strictFunctionTypes": false,
```
mightyiam: If we let `tsconfigs` set this to `true`, we fail to compile.

tsconfig.json
```diff
@@ -1,19 +1,11 @@
...
         "target": "ES5",
-        "noImplicitAny": true,
```
mightyiam: [tsconfigs has `strict: true`](https://github.com/mightyiam/tsconfigs/blob/1be54f2c1b2384828ced56327f77419e8a2a37a9/_private/common.json#L11).

tsconfig.json
```diff
@@ -1,19 +1,11 @@
...
-        "sourceMap": true,
```
mightyiam: [tsconfigs sets this](https://github.com/mightyiam/tsconfigs/blob/1be54f2c1b2384828ced56327f77419e8a2a37a9/_private/common.json#L10).

tsconfig.json
```diff
@@ -1,19 +1,11 @@
...
-        "strictNullChecks": true,
```
mightyiam: `strict: true`.

tsconfig.json
```diff
@@ -1,19 +1,11 @@
...
-        "declaration": true,
```
mightyiam: [tsconfigs sets this](https://github.com/mightyiam/tsconfigs/blob/1be54f2c1b2384828ced56327f77419e8a2a37a9/_private/module.json#L4).

tsconfig.json
```diff
@@ -1,19 +1,11 @@
...
-        "removeComments": false,
```
mightyiam: [TypeScript default is `false`](https://www.typescriptlang.org/docs/handbook/compiler-options.html) and tsconfigs doesn't set it.

tsconfig.json
```diff
@@ -1,19 +1,11 @@
...
-        "noUnusedLocals": true,
```
mightyiam: This is more suitable for linting. The TypeScript team feel this way. I won't go digging the comments now. We have linting now.

tsconfig.json
```diff
@@ -1,19 +1,11 @@
...
         "jsx": "react",
-        "jsxFactory": "jsx",
-        "lib": [
-            "dom",
-            "es5",
-            "es2015.core"
-        ]
```
mightyiam: [tsconfigs sets this to `["ESNext", "DOM"]`](https://github.com/mightyiam/tsconfigs/blob/1be54f2c1b2384828ced56327f77419e8a2a37a9/browser-module.json#L4). That means that TypeScript won't error if we accidentally attempt to use some environment feature that is newer. But our CI browser tests will. And besides, we should know. Although I've been caught once or twice using a totally new `Object.entries` on a project that targets older environments.

tsconfig.json
```diff
@@ -1,15 +1,12 @@
 {
+    "extends": "tsconfigs/browser-module",
     "compilerOptions": {
+        "strictFunctionTypes": false,
+        "strictBindCallApply": false,
+        "noImplicitThis": false,
```
mightyiam: If we let tsconfigs' `strict: true` set these implicitly to `true` then there are compilation errors.

mightyiam:
I also diffed between the build artifacts of the master branch and this branch. No change. Specifically, I diffed both `/snabbdom.js` and `/es/snabbdom.js`

PR #524, opened 2020-01-07
fix: include JSX related module in package

### LINKED ISSUE #503: JSX related files excluded from package
Should be added here:

https://github.com/snabbdom/snabbdom/blob/fa7b7d0796cfe2e8a0e61596beb4cbebeb53a146/package.json#L80-L235

mightyiam:
Fixes #503.


package.json
```diff
@@ -241,6 +247,10 @@
     "/is.d.ts",
     "/is.js",
     "/is.js.map",
+    "/jsx-global.d.ts",
+    "/jsx.d.ts",
+    "/jsx.js",
+    "/jsx.js.map",
```
mightyiam: Package the CommonJS version of the JSX related modules.

src/jsx-global.ts
```diff
@@ -5,6 +5,7 @@ declare global {
    * opt-in jsx intrinsic global interfaces
    * see: https://www.typescriptlang.org/docs/handbook/jsx.html#type-checking
    */
+  // eslint-disable-next-line @typescript-eslint/no-namespace
```
mightyiam: In a `.d.ts` file it doesn't complain. Perhaps we are not linting those at all? In any case, just ignoring this error seems to be the right thing to do.

tsconfig.json
```diff
@@ -36,7 +36,7 @@
         "src/htmldomapi.ts",
         "src/hooks.ts",
         "src/is.ts",
-        "src/jsx-global.d.ts",
+        "src/jsx-global.ts",
```
mightyiam: This source file is renamed.

src/jsx-global.ts
nojvek: I'm guessing namespaces might be allowed in .d.ts but not in .ts

nojvek:
Will fix #502 to just have sel shorthand once you merge this.

mightyiam:
Thank you, @nojvek.

PR #531, opened 2020-01-08
test: compare deep instead of JSON in assertions

mightyiam:


mightyiam:
Waiting for #533.

kuraga:
Note, please: seems like [we use](https://github.com/snabbdom/snabbdom/blob/2dc536a394980dc6e8e52ad8cfc4fd52fe5de15e/package.json#L21) Browserify v14.4.0 [which uses](https://github.com/browserify/browserify/blob/14.4.0/package.json#L24) [commonjs-assert](https://github.com/browserify/commonjs-assert) v1.4.0. But commonjs-assert [has been rewritten](https://github.com/browserify/commonjs-assert/pull/44) and released as v2.0.0. Including rewriting of `deep[Strict]Equals` function(s)...

nojvek:
not sure if you've seen the CI failed output. This is what I meant by undefined checks

```
Chrome 79.0.3945 (Windows 10.0.0) snabbdom jsx can be used as a jsxFactory method FAILED
	AssertionError: { sel: 'div',
	  data: { title: 'Hello World' },
	  children: undefined,
	  text: 'Hello World',
	  elm: undefined,
	  key: undefined deepStrictEqual { sel: 'div',
	  data: { title: 'Hello World' },
	  text: 'Hello World' }
	    at Context.<anonymous> (src/test/jsx.tsx:9:13 <- src/test/jsx.js:16:20)
Chrome 79.0.3945 (Windows 10.0.0) snabbdom jsx creates text property for text only child FAILED
	AssertionError: { sel: 'div',
	  data: {},
	  children: undefined,
	  text: 'foo bar',
	  elm: undefined,
	  key: undefined } deepStrictEqual { sel: 'div', data: {}, text: 'foo bar' }
	    at Context.<anonymous> (src/test/jsx.tsx:19:13 <- src/test/jsx.js:24:20)
Chrome 79.0.3945 (Windows 10.0.0) snabbdom jsx creates an array of children for multiple children FAILED
	AssertionError: { sel: 'div',
	  data: {},
	  children: 
	   [ { sel: undefined,
	       data: undefined,
	       children: undefined,
	       text: ' deepStrictEqual { sel: 'div',
	  data: {},
	  children: [ { text: 'foo' }, { text: 'bar' } ] }
	    at Context.<anonymous> (src/test/jsx.tsx:29:13 <- src/test/jsx.js:34:20)
```

mightyiam:
Yes, @nojvek, I saw it and figured that it is what you meant then.

mightyiam:
@nojvek and others, now that we have better assertions and I've added the undefined properties, could you please take a look and say whether this feels right? I personally feel that we should acknowledge the existence of these undefined properties and include them in the assertions. It could be argued that we could abstract them. Or some of them... but... abstractions of data in tests could confuse. Data in tests in general should be as is, even if it seems bloated. That's how I feel about it.

kuraga:
My two cents: I don't see any problem with this code's approach. Yes its code contains lot of `undefined`s. But that's true: `undefined`s are there.

mightyiam:
@nojvek I kind of lost you when I read "X's new state", because I don't see objects with state. I see stateless pure function tests.

```ts
    it('can be used as a jsxFactory method', function () {
      // This is transpiled into a call to a specific function that we're testing, with some arguments.
      const vnode = <div title="Hello World">Hello World</div>;

      // And this asserts on the output of that call.
      assert.deepStrictEqual(vnode, {
        sel: 'div',
        data: { title: 'Hello World' },
        children: undefined,
        elm: undefined,
        text: 'Hello World',
        key: undefined
      });
    });
```

Could you please go towards me and explain using terms of pure function return value testing?

nojvek:
I don't have strong opinions. Personally the undefines look like we're asserting internals, but I'm cool with what you have. 

Don't want to let perfect be the enemy of good. I value progress.

mightyiam:
I appreciate both of your view points, @nojvek. And most of all, the fact that you show up and present them.

Issue #461, opened 2019-10-30
Unmounting by `patch(vnode, null)` documented but not implemented


mreinstein:
My setup code:

```javascript
const snabbdom = require('snabbdom')
const h = require('snabbdom/h').default; // helper function for creating vnodes


const patch = snabbdom.init([ // Init patch function with chosen modules
  require('snabbdom/modules/class').default, // makes it easy to toggle classes
  require('snabbdom/modules/props').default, // for setting properties on DOM elements
  require('snabbdom/modules/style').default, // handles styling on elements with support for animations
  require('snabbdom/modules/eventlisteners').default, // attaches event listeners
]);


let _currentVnode, _counter = 0;


const update = (oldVnode) => {
    const newVnode = h('div#container.two.classes', { on: { click: someFn } }, [
        h('span', { }, `counter: ${_counter}`)
    ]);

    _currentVnode = patch(oldVnode, newVnode);
};


const destroy = () => {
    patch(_currentVnode, null);
};


const someFn = (e) => {
    _counter++;
    update(_currentVnode);
    destroy();
};


const container = document.querySelector('div');
update(container);

```

When I click the div, I see the counter increment, but then there's a console error, and I never see the dom get unmounted:

```
bundle.js:177 Uncaught TypeError: Cannot read property 'key' of null
```
which corresponds to this code from snabbdom:

```javascript
 function sameVnode(vnode1, vnode2) {
        return vnode1.key === vnode2.key && vnode1.sel === vnode2.sel;
    }
```

Any ideas what I'm doing wrong? Or is this a bug in the ~destroy~ patch code?

mreinstein:
I'm using snabbdom@0.7.3 (installed from npm)

mreinstein:
I've also tried the inline example listed in the readme, and this fails too, the same error: https://github.com/snabbdom/snabbdom#inline-example 

I'm becoming more convinced this is either a bug or a malformed example. Any help would be greatly appreciated please. :)

mightyiam:
Thank you for your interest in snabbdom, @mreinstein. I'm going to try to help you as much as I can.

I read your example code and I don't see what part the `destroy` function is meant to play in this example. Could you please try not calling `destroy`?

I'm going to close this but feel free to continue discussion.

mreinstein:
>  I'm going to try to help you as much as I can.

I appreciate the willingness to help! Can you please not close the issue before we've considered it? I think once we look into this you'll see that something is going on.

[...]

PR #536, opened 2020-01-10
fix: patch(vNode, null) creates empty node #461

mreinstein:


src/snabbdom.ts
```diff
@@ -297,6 +297,10 @@ export function init (modules: Array<Partial<Module>>, domApi?: DOMAPI) {
       oldVnode = emptyNodeAt(oldVnode);
     }
 
+    if (!(vnode && isVnode(vnode))) {
+      vnode = emptyNode;
+    }
```
mightyiam: I would check specifically for `vnode === null`.

mreinstein: @mightyiam doesn't it instead make sense to add the null check inside of `isVnode` ?

mightyiam: That doesn't feel right to me. It's not whether it's a VNode. It's whether it's specifically `null`.

PR #539, opened 2020-01-13
docs: rm unimplemented unmounting by 2nd arg null

### LINKED ISSUE #461: Unmounting by `patch(vnode, null)` documented but not implemented

mightyiam:
Fixes #461.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,9 +84,6 @@ var newVnode = h('div#container.two.classes', {on: {click: anotherEventHandler}}
 ]);
 // Second `patch` invocation
 patch(vnode, newVnode); // Snabbdom efficiently updates the old view to the new state
-
-// to unmount from the DOM and clean up, simply pass null
-patch(newVnode, null)
 ```
 
 ## More examples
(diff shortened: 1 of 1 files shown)


mightyiam:
@mreinstein how do you feel about this, please?

mreinstein:
Looks good to me! Thanks for the opportunity for input, and for the fix!

PR #542, opened 2020-01-15
chore(test): karma-webpack, test both CommonJS & ES modules

### LINKED ISSUE #543: Packaged artifacts not tested
The two sets of build artifacts that this project provides — CommonJS and ES modules — are not being tested currently. Instead, karma-typescript builds its own bundle (likely in-memory) and that what is being tested.

https://github.com/snabbdom/snabbdom/blob/d4a8640132ceb135c21eeda84e2c0b58828eff5a/karma.conf.js#L17-L64

mightyiam:
Fixes #543.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -17,20 +17,29 @@ const browsers = ci
 module.exports = function (config) {
   config.set({
     basePath: '.',
-    frameworks: ['mocha', 'karma-typescript'],
+    frameworks: ['mocha'],
     // list of files / patterns to load in the browser
-    files: [{ pattern: 'src/**/*' }],
+    files: [
+      { pattern: 'test/**/*.js' },
+      { pattern: 'es/test/**/*.js' }
+    ],
     plugins: [
       'karma-mocha',
       require('karma-mocha-reporter'),
       'karma-chrome-launcher',
       'karma-firefox-launcher',
       'karma-browserstack-launcher',
-      'karma-typescript',
+      require('karma-webpack')
     ],
     hostname: ci ? ip : 'localhost',
     preprocessors: {
-      'src/**/*.{ts,tsx}': ['karma-typescript']
+      '**/*.js': ['webpack']
+    },
+    webpack: {
+      mode: 'development'
+    },
+    webpackMiddleware: {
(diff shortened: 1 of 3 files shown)


karma.conf.js
```diff
@@ -17,19 +17,28 @@ const browsers = ci
 module.exports = function (config) {
   config.set({
     basePath: '.',
-    frameworks: ['mocha', 'karma-typescript'],
+    frameworks: ['mocha'],
     // list of files / patterns to load in the browser
-    files: [{ pattern: 'src/**/*' }],
+    files: [
+      { pattern: 'test/**/*.js' },
+      { pattern: 'es/test/**/*.js' }
```
mightyiam: Tests are run twice — once for CommonJS and once for ES modules.

karma.conf.js
```diff
@@ -40,20 +49,7 @@ module.exports = function (config) {
...
-      },
-      compilerOptions: {
-        module: 'commonjs'
-      },
-      tsconfig: './tsconfig.json',
-      include: {
-        mode: 'merge',
-        values: ['src/test/**/*'],
-      },
-    },
```
mightyiam: And it's simpler.

kuraga:
@mightyiam thanks very much!

1.
```
    "compile-es": "tsc --outDir es --module es6",
    "compile-commonjs": "tsc --outDir . --module commonjs",
```
move these artifacts in a subfolder, maybe?

2. We now use Webpack (for browser testing). Get rid of Browserify, maybe?

mightyiam:
@kuraga you're welcome :bouquet: 

Yes, I'd love to move at least the CommonJS build into a subfolder. Just to be clear — this suggestion is not related to this PR, right?

Browserify is still used to bundle some examples and the UMD bundles. Both of these are on my radar.

kuraga:
@mightyiam is not related. Just too much to open two issues for this :smile: And new code uses compilation.

> Browserify is still used to bundle some examples and the UMD bundles. Both of these are on my radar.

:+1:

mightyiam:
@kuraga what do you mean by "new code uses compilation", please? What new code and what compilation, please?

Issue #540, opened 2020-01-14
The value of "before" always seems to be null


zymfe:
In the updateChildren function, the value of the "before" variable always seems to be null, because when oldStartIdx > oldEndIdx,  the newly added nodes will be appended to the end of parentElm, what is the scenario with null before. Thanks.

mightyiam:
Thank you for the tip, @zymfe. Pull request welcome.

PR #544, opened 2020-01-15
fix: the value of before always seems to be null

zymfe:
The value of "before" always seems to be null (#540)


mightyiam:
@zymfe thank you for the pull request. It seems that some tests a failing. Perhaps `before` is sometimes not `null`?

PR #548, opened 2020-01-16
style: four eslint rules enabled

mightyiam:
- import/newline-after-import
- dot-notation
- import/first
- no-undef

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,6 +6,7 @@ module.exports = {
     '/dist/'
   ],
   rules: {
+    'import/newline-after-import': 'error',
     '@typescript-eslint/no-non-null-assertion': 'off',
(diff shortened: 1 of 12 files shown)


.eslintrc.js
```diff
@@ -16,10 +17,7 @@ module.exports = {
     '@typescript-eslint/restrict-plus-operands': 'off',
     '@typescript-eslint/strict-boolean-expressions': 'off',
     'comma-dangle': 'off',
-    'dot-notation': 'off',
-    'import/first': 'off',
```
paldepind: Why was this removed❓

mightyiam: @paldepind notice that most rules listed here are actually turning off rules that are turned on in the shareable config we extend from. So removing them is actually adding them.

mightyiam:
Thank you for reviewing, @paldepind .

PR #550, opened 2020-01-18
test: workaround IE11 defficiencies

### LINKED ISSUE #471: IE11 fails in CI
https://travis-ci.org/snabbdom/snabbdom/builds/611279225#L362-L414

mightyiam:
Fixes #471.

### DIFF
--- a/src/test/attributes.ts
+++ b/src/test/attributes.ts
@@ -38,10 +38,10 @@ describe('attributes', function () {
   it('are not omitted when falsy values are provided', function () {
     var vnode1 = h('div', { attrs: { href: null as any, minlength: 0, value: '', title: 'undefined' } });
     elm = patch(vnode0, vnode1).elm;
-    assert.strictEqual(elm.getAttribute('href'), 'null');
-    assert.strictEqual(elm.getAttribute('minlength'), '0');
-    assert.strictEqual(elm.getAttribute('value'), '');
-    assert.strictEqual(elm.getAttribute('title'), 'undefined');
+    assert.ok(elm.hasAttribute('href'));
+    assert.ok(elm.hasAttribute('minlength'));
+    assert.ok(elm.hasAttribute('value'));
+    assert.ok(elm.hasAttribute('title'));
   });
   it('are set correctly when namespaced', function () {
     var vnode1 = h('div', { attrs: { 'xlink:href': '#foo' } });
@@ -77,11 +77,11 @@ describe('attributes', function () {
       assert.strictEqual(elm.hasAttribute('required'), false);
       assert.strictEqual(elm.getAttribute('required'), null);
     });
-    it('is not omitted if the value is falsy but casted to string', function () {
+    it('is not omitted if the value is falsy', function () {
       var vnode1 = h('div', { attrs: { readonly: 0, noresize: null as any } });
       elm = patch(vnode0, vnode1).elm;
-      assert.strictEqual(elm.getAttribute('readonly'), '0');
-      assert.strictEqual(elm.getAttribute('noresize'), 'null');
+      assert.ok(elm.hasAttribute('readonly'));
+      assert.ok(elm.hasAttribute('noresize'));
     });
   });
   describe('Object.prototype property', function () {
--- a/src/test/core.ts
+++ b/src/test/core.ts
@@ -11,6 +11,8 @@ import vnode, { VNode } from '../vnode'
 import htmlDomApi from '../htmldomapi'
 import { CreateHook, InsertHook, PrePatchHook, RemoveHook, InitHook, DestroyHook, UpdateHook } from '../hooks'
 
+const hasSvgClassList = 'classList' in SVGElement.prototype
+
 var patch = init([
   classModule,
   propsModule,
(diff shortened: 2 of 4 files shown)


src/test/attributes.ts
```diff
@@ -38,10 +38,10 @@ describe('attributes', function () {
...
     var vnode1 = h('div', { attrs: { href: null as any, minlength: 0, value: '', title: 'undefined' } });
     elm = patch(vnode0, vnode1).elm;
-    assert.strictEqual(elm.getAttribute('href'), 'null');
-    assert.strictEqual(elm.getAttribute('minlength'), '0');
-    assert.strictEqual(elm.getAttribute('value'), '');
-    assert.strictEqual(elm.getAttribute('title'), 'undefined');
+    assert.ok(elm.hasAttribute('href'));
+    assert.ok(elm.hasAttribute('minlength'));
+    assert.ok(elm.hasAttribute('value'));
+    assert.ok(elm.hasAttribute('title'));
```
mightyiam: In IE11, the value for attributes set to `null` ends up being an actual `null`, instead of a `'null'` string. So I figured it would be best to just test for the existence of these attributes, instead of their values. "are not omitted" seems to be covered. I feel this is good enough.

src/test/attributes.ts
```diff
@@ -77,11 +77,11 @@ describe('attributes', function () {
...
       assert.strictEqual(elm.getAttribute('required'), null);
     });
-    it('is not omitted if the value is falsy but casted to string', function () {
+    it('is not omitted if the value is falsy', function () {
       var vnode1 = h('div', { attrs: { readonly: 0, noresize: null as any } });
       elm = patch(vnode0, vnode1).elm;
-      assert.strictEqual(elm.getAttribute('readonly'), '0');
-      assert.strictEqual(elm.getAttribute('noresize'), 'null');
+      assert.ok(elm.hasAttribute('readonly'));
+      assert.ok(elm.hasAttribute('noresize'));
```
mightyiam: Casting to a string is most likely performed by the platform. And since IE11 seems to keep the `null` as a real `null`, it seems enough for us to test for the presence of the boolean attribute. A boolean attribute with `null` as a value is still present (truthy), right?

src/test/core.ts
```diff
@@ -11,6 +11,8 @@ import vnode, { VNode } from '../vnode'
 import htmlDomApi from '../htmldomapi'
 import { CreateHook, InsertHook, PrePatchHook, RemoveHook, InitHook, DestroyHook, UpdateHook } from '../hooks'
 
+const hasSvgClassList = 'classList' in SVGElement.prototype
```
mightyiam: IE11 does not have a `classList` on `SVGElement`. So some tests would just be skipped in it. I'm OK with that.

src/test/core.ts
```diff
@@ -322,7 +326,7 @@ describe('snabbdom', function () {
         var prevElm = document.createElement('div');
         prevElm.id = 'id';
         prevElm.className = 'class';
-        var text = new Text('Foobar');
+        var text = document.createTextNode('Foobar');
```
mightyiam: IE11 does not have `window.Text`.

src/test/core.ts
```diff
@@ -358,9 +362,10 @@ describe('snabbdom', function () {
         assert.strictEqual(elm.childNodes[0].textContent, 'Hello');
       });
       it('can work with domApi', function () {
-        var domApi = Object.assign({}, htmlDomApi, {
+        var domApi = {
+          ...htmlDomApi,
           tagName: function (elm: HTMLElement) { return 'x-' + elm.tagName.toUpperCase(); }
-        });
+        };
```
mightyiam: IE11 does not have `Object.assign`.

src/test/style.ts
```diff
@@ -9,6 +9,10 @@ var patch = init([
   styleModule
 ]);
 
+const featureDiscoveryElm = document.createElement('div')
+featureDiscoveryElm.style.setProperty('--foo', 'foo')
+const hasCssVariables = featureDiscoveryElm.style.getPropertyValue('--foo') === 'foo'
```
mightyiam: IE11 does not have CSS variables. Simply skip those tests in it. I'm OK with that.

mightyiam:
Take a look at this line in CI where IE11 passes :fireworks: 
https://travis-ci.org/snabbdom/snabbdom/builds/638732294#L320

src/test/attributes.ts
kuraga: So `elm.getAttribute('href')` is `null` instead of `'null'` in IE11. What about three else lines?

src/test/attributes.ts
kuraga: So `elm.getAttribute('noresize')` is `null` instead of `'null'` in IE11. What about the first line?

src/test/core.ts
```diff
@@ -145,6 +147,7 @@ describe('snabbdom', function () {
       assert(!elm.classList.contains('not'));
     });
     it('receives classes in selector when namespaced', function () {
+      if (!hasSvgClassList) return
```
kuraga: Why `return` not `skip`?

kuraga:
@mightyiam thanks! Add `IE11 workaround` comment everywhere?

src/test/attributes.ts
mightyiam: Just keeping it consistent and loyal to the purpose of this specific test. "Not omitted".

src/test/attributes.ts
mightyiam: `hasAttribute` is all that's necessary for a boolean attribute, I figure.

src/test/core.ts
mightyiam: What API includes, `skip`, please? Is that a mocha thing?

mightyiam:
@kuraga, I prefer to refrain from name-calling in test skipping. Feature detection seems to be a better idea anyway. Whatever browser doesn't have CSS variables, for example. I don't care what browser that is.

kuraga:
@mightyiam I meant to highlight a trick. To remove it (trick) when we would decide to drop IE11 support.

src/test/core.ts
kuraga: https://mochajs.org/#inclusive-tests, `this.skip(...)`

mightyiam:
I don't know what are our supported browsers. It could be that it's not only IE11 that lacks CSS variables, for example. The, what would the comment say? It doesn't seem important enough to figure this all out, to me. If it ever bothers anyone, they could check whether it still passes CI and that would be it. Or if you'd like to come up with the correct comments then I'd support that.

kuraga:
@mightyiam ,

> It could be that it's not only IE11

there is a probability there will be much code for _only_ one browser. And we drop support of this browser it would be nice to remove code tricks for it (to make code simpler). Comments could help to find out this code.

But I see your position, question seems to be settled.

src/test/core.ts
mightyiam: Good tip. Done.

src/test/core.ts
jvanbruegge: you can also load a classlist polyfill for this case

mightyiam: Thanks, @jvanbruegge. I actually did that and then changed to this. Have a feeling that this is better. Not sure why.

PR #555, opened 2020-01-28
test: enable IE10

### LINKED ISSUE #472: IE10 fails in CI
https://travis-ci.org/snabbdom/snabbdom/builds/611279225#L415-L477

mightyiam:
Closes #472.


mreinstein:
I just want to say thank you for continuing to support this browser. 

This is one of many reasons why snabbdom makes a lot of sense instead of other approaches, for anyone doing serious web development that has to reach the widest audience possible on the web.

PR #558, opened 2020-02-20
Enable some ESLint rules

mightyiam:


paldepind:
PR seems to do more than just enable the `one-var` rule. Semicolons are gone? 🤔

mightyiam:
Had a glitch there. Ready for review!

paldepind:
I don't think we should remove semicolons. That's pretty atypical.

mightyiam:
@paldepind changed semicolons to always. Rebased.

mightyiam:
Hold on. There's an error.

mightyiam:
OK cool. I much prefer no semicolons but that should not hold up the rest of the rules in this PR.

PR #569, opened 2020-03-29
test: benchmark core

mightyiam:


mightyiam:
The CI run fails due to missing environment variables. @paldepind could you please provide me with access to add those?

PR #582, opened 2020-04-12
chore: renovate no automerge

mightyiam:


mightyiam:
This should disable automerge. Automerge should be agreed on by @paldepind.

Issue #593, opened 2020-05-01
Renovate updates `.nvmrc` & `.node-version` separately from `.travis.yml`


mightyiam:
See #588. Only the Travis file is updated.

Perhaps removing the less popular `nodenv` file would solve it? Or maybe removing the `v` prefix from these files?

mightyiam:
According to #573, renovate does update `.node-version` and `.nvmrc`. It is `nodenv` that does not get updated. But that's #609.

This issue is about renovate not updating Travis-CI Node.js version and the `.nvmrc` type of files together. Here's a PR that updates only the Travis-CI Node.js: #588. And #573 updated only the `.nvmrc` type of files.

mightyiam:
https://github.com/renovatebot/renovate/issues/6313

mightyiam:
#646

PR #594, opened 2020-05-01
chore: remove v prefix from node version files

mightyiam:


mightyiam:
This is an attempt to solve #593. It should be clear whether it fixes the issue only after merging. Could I get an approval, please?

PR #596, opened 2020-05-01
an es modules only re-factor. fixes #516  DO NOT MERGE

### LINKED ISSUE #516: Snabbdom and ECMAScript Modules
I think it's time to discuss: Snabbdom and ECMAScript Modules.

My thoughts: Snabbdom should be standard-compliant as much as possible.
UPD: About current state of the art, see for instance [1](https://2ality.com/2019/04/nodejs-esm-impl.html), [2](https://2ality.com/2019/10/hybrid-npm-packages.html).

All issues about this ATM: #200, #225, #247, #261, #312, #319, #338, #437, #522, #523, #542, #543, cyclejs/cyclejs#925, #568, #601.

/cc @paldepind , @mightyiam , @nojvek

mreinstein:


mreinstein:
also fixes #525 since we don't need build tools for the examples anymore.

examples/reorder-animation/script.js
```diff
@@ -1,11 +1,16 @@
...
-  require('../../modules/style').default,
-  require('../../modules/eventlisteners').default,
+import * as snabbdom from '../../es/snabbdom.js';
+import clazz from '../../es/modules/class.js';
+import h from '../../es/h.js';
+import props from '../../es/modules/props.js';
+import style from '../../es/modules/style.js';
+import listeners from '../../es/modules/eventlisteners.js';
+
+const patch = snabbdom.init([
```
mightyiam, line 8: `var`

examples/reorder-animation/script.js
```diff
@@ -25,6 +30,7 @@ var originalData = [
   { rank: 9, title: 'The Lord of the Rings: The Return of the King', desc: 'Gandalf and Aragorn lead the World of Men against Sauron\'s army to draw his gaze from Frodo and Sam as they approach Mount Doom with the One Ring.', elmHeight: 0 },
   { rank: 10, title: 'Fight Club', desc: 'An insomniac office worker looking for a way to change his life crosses paths with a devil-may-care soap maker and they form an underground fight club that evolves into something much, much more...', elmHeight: 0 },
 ];
+
```
mightyiam, line 33: Unrelated.

examples/svg/script.js
```diff
@@ -1,12 +1,14 @@
...
+const patch = snabbdom.init([
+  attributes
 ]);
-var h = require('../../h.js').default;
 
 window.addEventListener('DOMContentLoaded', () => {
-  var container = document.getElementById('container');
-  var vnode = h('div', [
+  const container = document.getElementById('container');
+  const vnode = h('div', [
```
mightyiam, line 11: `var`s

package.json
```diff
@@ -2,9 +2,21 @@
...
+    "./h": "./es/h.js",
+    "./attributes": "./es/modules/attributes.js",
+    "./class": "./es/modules/class.js",
+    "./dataset": "./es/modules/dataset.js",
+    "./eventlisteners": "./es/modules/eventlisteners.js",
+    "./props": "./es/modules/props.js",
+    "./style": "./es/modules/style.js",
+    "./tovnode": "./es/tovnode.js"
+  },
+  "typings": "./es/snabbdom.d.ts",
```
mightyiam, line 19: @mreinstein could you please make a PR that fixes #597?

src/h.ts
```diff
@@ -1,5 +1,5 @@
-import { vnode, VNode, VNodeData } from './vnode';
-import * as is from './is';
+import { vnode, VNode, VNodeData } from './vnode.js';
+import * as is from './is.js';
```
mightyiam, line 2: It seems that all of the source code TypeScript import paths in the source are changed from implicit/no extension to `.js`. I am assuming that this is to affect the output compiled modules' import paths to be the same for the purpose of being able to use them as ES modules in the browser.

Yes, TypeScript's support for resolution of extensions is not ES modules spec compliant. And yes, providing ES modules means that any import paths in them must have paths with extensions.

But this is not how I would do it. The source files should remain unchanged. Instead, use [this project](https://github.com/dropbox/ts-transform-import-path-rewrite) to transform the import paths conditionally. Use `ttypescript`. There's an example there that uses `ttypescript`. The `alias` object needs to be something like 
```json
{
  "^\\.(.*)(?<!\\.js)&": "$1.js"
}
```

tsconfig.json
```diff
@@ -4,7 +4,9 @@
         "strictFunctionTypes": false,
         "strictBindCallApply": false,
         "noImplicitThis": false,
-        "target": "ES5",
+        "target": "ES6",
+        "module": "ES6",
```
mightyiam, line 8: The value of `module` does not require overriding here:
https://github.com/mightyiam/tsconfigs/blob/7cfbd4100e10f6c6bf0bd605a3ace616036b3e96/browser-module.json#L5

mreinstein:
To be honest, I think you should just come up with your own PR for this. There are so many criticisms of the requested change, I think it would just be easier to get it formatted exactly how you want it the first time. :) Best of luck!

PR #601, opened 2020-05-09
feat: only esm and correct import paths

### LINKED ISSUE #263: Please document how to run examples locally
I'd like to experiment with a Custom elements example. Perhaps you could help me ;)
Trying with cycle.js "on top" is causing too many problems...
Thanks.

### LINKED ISSUE #437: open to a PR that updates all README examples to es modules?
using commonjs is a pattern that the community is trying to move away from. Given snabbdom already supports es modules, would you be open to a PR that stops referencing `require` in the `README.md` in favor of esm? Would look something like this for the inline example:

```javascript
import { h, init } from 'snabbdom';
import classModule from 'snabbdom/modules/class';
import propsModule from 'snabbdom/modules/props';
import styleModule from 'snabbdom/modules/style';
import eventModule from 'snabbdom/modules/eventlisteners';


// Init patch function with chosen modules
const patch = init([
	classModule,  // makes it easy to toggle classes
	propsModule,  // for setting properties on DOM elements
	styleModule,  // handles styling on elements with support for animations
	eventModule   // attaches event listeners
]);
```

Happy to send a PR if this is a welcome addition!

### LINKED ISSUE #516: Snabbdom and ECMAScript Modules

mightyiam:
BREAKING CHANGE: CommonJS module are no longer provided.
BREAKING CHANGE: import paths in ES modules include file name
extensions.
BREAKING CHANGE: Compiled to ES2015 (was ES5).

Fixes #516.
Fixes #437.
Fixes #263. Kind of. Because there is no build step.

Thanks to @mreinstein on starting this work.


mreinstein:
Sorry for not following this change through to completion; I wasn't trying to be stubborn, just figured this change would probably be easier for you to make, in exactly how you want it. 

Great work!

README.md
```diff
@@ -58,14 +58,20 @@ performance, small size and all the features listed below.
...
 
 ```javascript
-var snabbdom = require('snabbdom');
-var patch = snabbdom.init([ // Init patch function with chosen modules
-  require('snabbdom/modules/class').default, // makes it easy to toggle classes
-  require('snabbdom/modules/props').default, // for setting properties on DOM elements
-  require('snabbdom/modules/style').default, // handles styling on elements with support for animations
-  require('snabbdom/modules/eventlisteners').default, // attaches event listeners
+import { init } from 'snabbdom';
+import clazz from 'snabbdom/class';
```
jvanbruegge, line 62: Why still split? With tree shaking we could have everything just as snabbdom package

package.json
```diff
@@ -2,9 +2,30 @@
   "name": "snabbdom",
   "version": "0.7.4",
   "description": "A virtual DOM library with focus on simplicity, modularity, powerful features and performance.",
-  "main": "snabbdom.js",
+  "type": "module",
+  "main": "./es/snabbdom.js",
```
jvanbruegge, line 6: Why not use rollup here? This will break for everyone with a node less than 14.

mreinstein: That's kind of the whole point of this pure es modules work. To remove the need for bundlers and simplify workflow. It's intentionally breaking with the old way, that's why this work isn't being done on master branch.

README.md
mreinstein: pure es modules is potentially enabling a bundless workflow; tree shaking is an artifact of bundling.

mightyiam: If we place everything in one module, we gain a more concise import (1 statement with multiple names). But we lose "tree shaking" in the case of not using a bundler. I would pay the price of the multiple import statements for split modules without a bundler.

package.json
mightyiam: If we provide as is in this branch (ES modules with `.js` extensions and extensions in import paths), then Node.js consumers could transpile it themselves. Whoever goes to the lengths of server side rendering or testing in Node.js instead of in a browser, should be able to solve this. And this is the future of packages, isn't it? Publish in the spec compliant form and let the consumer figure out the rest. No?

README.md
kuraga: And shorter files...

kuraga:
:+1: :+1: :+1:

mightyiam:
I am investigating the reason for the commitlint failure in CI. It involves force pushing and examining build log.

kuraga:
:tada: :tada: :tada:

PR #611, opened 2020-05-10
fix: renovate node supportPolicy array

mightyiam:


mightyiam:
In hope that it would solve #593.

PR #624, opened 2020-05-14
fix(props): do not attempt to delete node properties

### LINKED ISSUE #283: Props with falsy values are deleted
Description: If a prop has a key that has a falsy value, the key is deleted.

Reproducer: https://jsfiddle.net/0q6uueou/

Affected Browsers: Chrome 42 and prior on windows (possibly others).

Seems due to this if condition: https://github.com/snabbdom/snabbdom/blob/master/src/modules/props.ts#L15

I ran into a problem while debugging snabbdom in older versions of chrome. scrollTop starts at 0, and in chrome 42 and prior you were able to delete scrollTop from an element.

This is still a bug in latest chrome in certain situations, as shown in the reproducer.

### LINKED ISSUE #415: Props are not reseted properly
I would like to know if this is a bug or is it the intended behaviour:

https://codesandbox.io/s/ppn1y5okjq

Please enlighten me about how to use snabbdom. I use it for little toy language here https://github.com/batiste/blop-language and I am running into funny bugs

### LINKED ISSUE #623: props module should not attempt to delete DOM node props
Here is the whole props module.

https://github.com/snabbdom/snabbdom/blob/a42c31a300b7d689ad8a0b424afac67636da54d7/src/modules/props.ts#L1-L34

On line 21 it attempts to delete DOM node properties. This does not actually work. It should not be attempted. The properties are not deleted, nor their values affected. They remain unchanged.

This is at best extraneous code. It also confuses users who happen to read the props module code.

mightyiam:
BREAKING CHANGE: props module does not attempt to delete node
properties. This may affect you if you are using the props module
to add non-native (custom) properties to DOM nodes. Instead, it is
recommended to use _data-* attributes_.
https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

Fixes #623.
Fixes #283.
Closes #415.
Closes #307.
Closes #151.
Closes #416.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,6 +371,14 @@ Allows you to set properties on DOM elements.
 h('a', {props: {href: '/foo'}}, 'Go to Foo');
 ```
 
+Properties can only be set. Not removed. Even though browsers allow addition and
+deletion of custom properties, deletion will not be attempted by this module.
+This makes sense, because native DOM properties cannot be removed. And
+if you are using custom properties for storing values or referencing
+objects on the DOM, then please consider using
+[data-\* attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes)
+instead. Perhaps via [the dataset module](#the-dataset-module).
+
 ### The attributes module
 
 Same as props, but set attributes instead of properties on DOM elements.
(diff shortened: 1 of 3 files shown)


mightyiam:
@batiste I guess people do run into weird bugs. This PR closes some open issues. I would love to merge it soon.

PR #634, opened 2020-05-15
perf(class): prevent redundant calls to classList.remove

### LINKED ISSUE #633: Redundant calls to classList.remove in class module
Here's the entire class module:

https://github.com/snabbdom/snabbdom/blob/a42c31a300b7d689ad8a0b424afac67636da54d7/src/modules/class.ts#L1-L32

The iteration on line 18 seems to be for the purpose of identifying classes that existed on the previous vnode and do not exist on the current vnode.

But the condition is a general `!klass[name]`. This works, but it produces redundant calls to `classList.remove`. False positives. In three cases:

1. The previous is `false` and the current is not provided. Wrong, because the class should not be on the DOM node.
1. The previous is `false` and the current is `false`. Wrong, because the class should not be on the DOM node.
1. The previous is `true` and the current is `false`. Wrong, because this case is handled in iteration on the current class object (line 23).

mightyiam:
Fixes #633.

### DIFF
--- a/src/modules/class.ts
+++ b/src/modules/class.ts
@@ -16,7 +16,11 @@ function updateClass (oldVnode: VNode, vnode: VNode): void {
   klass = klass || {}
 
   for (name in oldClass) {
-    if (!klass[name]) {
+    if (
+      oldClass[name] &&
+      !Object.prototype.hasOwnProperty.call(klass, name)
+    ) {
+      // was `true` and now not provided
       elm.classList.remove(name)
     }
   }


mightyiam:
I have observed a considerable performance improvement applying this fix in production and no adverse effects.

src/modules/class.ts
```diff
@@ -16,7 +16,11 @@ function updateClass (oldVnode: VNode, vnode: VNode): void {
   klass = klass || {};
 
   for (name in oldClass) {
-    if (!klass[name]) {
+    if (
+      oldClass[name] &&
+      !Object.prototype.hasOwnProperty.call(klass, name)
```
paldepind, line 21: Is using `Object.prototype.hasOwnProperty.call(klass, name)` better than just `klass.name`? If so, then should we not do the same thing for `oldClass[name]` as well? I.e. change it to `Object.prototype.hasOwnProperty.call(oldClass, name)`? It's less readable though so unless there's a good reason I think just `klass[name]` whould be good.

mightyiam: > Is using `Object.prototype.hasOwnProperty.call(klass, name)` better than just `klass.name`?

It's not the same thing. Just `klass.name` would be checking for a falsey value. We don't want to check the value at all. Only whether the property exists.

> If so, then should we not do the same thing for `oldClass[name]` as well? I.e. change it to `Object.prototype.hasOwnProperty.call(oldClass, name)`? It's less readable though so unless there's a good reason I think just `klass[name]` whould be good.

Can't do that, because there we do need to check the value.

mightyiam: Edited my previous comment.

paldepind: I'm probably wrong, but what if `oldClass[name]` is `true` and `klass[name]` is `false`. Shouldn't the class be removed in that case as well?

staltz: Alternatives:
- `typeof klass[name] === 'undefined'`
- `klass[name] == null` (double equality checks for `null` and `undefined`)

staltz: My guess is that the performance benefit comes from applying `classList.remove` only when `oldKlass` and `klass` differ from each other, not because `klass[name]` is undefined or falsey. Also for the sake of conservatism (we don't know what other adverse affects the `Object.prototype.hasOwnProperty` choice could bring), I would keep it `!klass[name]`, so that we are just adding one more condition (i.e. `oldClass[name]` truthy). It makes sense that the more conditions you have, the more rare the conclusion will trigger, which to us means touching the DOM less.

mightyiam: > I'm probably wrong, but what if `oldClass[name]` is `true` and `klass[name]` is `false`. Shouldn't the class be removed in that case as well?

That case is handled in the iteration `for (name in klass) {`.

> My guess is that the performance benefit comes from applying `classList.remove` only when `oldKlass` and `klass` differ from each other, not because `klass[name]` is undefined or falsey. Also for the sake of conservatism (we don't know what other adverse affects the `Object.prototype.hasOwnProperty` choice could bring), I would keep it `!klass[name]`, so that we are just adding one more condition (i.e. `oldClass[name]` truthy). It makes sense that the more conditions you have, the more rare the conclusion will trigger, which to us means touching the DOM less.

In iterating over `for (name in oldClass) {` we would like to check exactly one thing. And that thing is whether `old` is `true` AND `new` does not exist at all. That way we iterate less times. "It was defined as true and now it is not defined at all. Remove the class.". That's the logic. That's the sole purpose for that iteration. Because immediately after iteration on the old properties is complete, there is iteration over new properties. And there, if this class exists and is falsey, it will be removed.

I feel that all the cases are explained in  the issue description, #633.

mightyiam: Also, what adverse affects could the `Object.prototype.hasOwnProperty` introduce, please? It is a prototype function that is expected to exist in a valid JavaScript environment since IE5.5, Chrome 1, Safari 3 and Firefox 1.

mightyiam: Also the `Function.prototype.call` that is used is as old as that.

staltz: "What adverse effects?"

Unpredicted ones

mightyiam: @staltz do you have a general direction? Are you suggesting we refrain from it? But this is the most accurate expression for this check, as far as I can tell. Checking `!klass[name]` means that we would be checking twice in the case that `name` exists on `klass`.

staltz: I guess `Object.prototype.hasOwnProperty` is fine. It's also faster than `typeof`: https://jsperf.com/typeof-versus-hasownproperty

paldepind:
I think we're all on board with this now. Really great change! 👍. Thanks a lot @mightyiam.

PR #637, opened 2020-05-17
fix(attrs): do not redeclare Element.setAttribute(NS)

### LINKED ISSUE #615: `Element.setAttribute` global is re-declared for the user
https://github.com/snabbdom/snabbdom/blob/a42c31a300b7d689ad8a0b424afac67636da54d7/src/modules/attributes.ts#L4-L10

mightyiam:
BREAKING CHANGE: Types exported by this package have re-declared
the global `Element.setAttribute` and `Element.setAttributeNS` to
accept `number` and `boolean` for the `value` parameter. This
change removes that re-declaration and thus the only valid value is
`string`. If your code provides `number` and/or `boolean`, then it
may now fail to compile.

Fixes #615.

### DIFF
--- a/src/modules/attributes.ts
+++ b/src/modules/attributes.ts
@@ -1,14 +1,6 @@
 import { VNode, VNodeData } from '../vnode'
 import { Module } from './module'
 
-// because those in TypeScript are too restrictive: https://github.com/Microsoft/TSJS-lib-generator/pull/237
-declare global {
-  interface Element {
-    setAttribute(name: string, value: string | number | boolean): void
-    setAttributeNS(namespaceURI: string, qualifiedName: string, value: string | number | boolean): void
-  }
-}
-
 export type Attrs = Record<string, string | number | boolean>
 
 const xlinkNS = 'http://www.w3.org/1999/xlink'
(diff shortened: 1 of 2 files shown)


src/modules/attributes.ts
```diff
@@ -1,14 +1,6 @@
...
 import { Module } from './module';
 
-// because those in TypeScript are too restrictive: https://github.com/Microsoft/TSJS-lib-generator/pull/237
-declare global {
-  interface Element {
-    setAttribute(name: string, value: string | number | boolean): void
-    setAttributeNS(namespaceURI: string, qualifiedName: string, value: string | number | boolean): void
-  }
-}
-
```
mightyiam, line 11: If you look below, you may find the reason for this re-declare. What I did not realize is that it would be re-declared for the users, as well. That's global pollution. It's a no-no.

So, you may ask why can this re-declaration be removed without the calls to `Element.setAttribute` below failing to type check? Good question. That's because this is still re-declared, even though this is removed. See, this is not the only place where this is re-declared. The other place is by importing our own project under a different name, in the benchmark test:

https://github.com/snabbdom/snabbdom/blob/ef719f41ca9eb5b9c08a5d2aa2790f8f3280a54f/src/benchmark/core.ts#L6
https://github.com/snabbdom/snabbdom/blob/ef719f41ca9eb5b9c08a5d2aa2790f8f3280a54f/package.json#L45

In light of that, you could claim that this PR is not complete, because the calls below should not pass type checking only due to the global pollution of the last published version of our package. Good point. Then, the more complete PR would also take care to receive the compiler's agreement on those calls. But I prefer to wait until this change is in a published package and then update that import so that the lack of the global pollution would actually assist me in making that more complete change with the correct type checking. It's kind of awkward but harmless and convenient.

kuraga: If we'd not remove that we should add a describing comment, IMHO.

src/modules/attributes.ts
```diff
@@ -37,6 +29,14 @@ function updateAttrs (oldVnode: VNode, vnode: VNode): void {
...
         elm.removeAttribute(key);
       } else {
+        /**
+         * The following `setAttribute` calls pass type checking
+         * with the `value` parameter possibly a `number` due to
+         * the import of `latest-snabbdom-release` in
+         * `src/benchmark/core.ts`, which re-declares `setAttribute`.
+         * For background, read
+         * https://github.com/snabbdom/snabbdom/issues/615
+         */
```
mightyiam, line 39: Here's a comment for you, @kuraga

PR #639, opened 2020-05-17
chore(ci): increase browser concurrency to 5

### LINKED ISSUE #638: 1 parallel browser in CI while quota is 5
According to [this page](https://www.browserstack.com/open-source), we should have 5 browsers available in CI. The CI build would go much faster.

Screenshot from 2020-05-17 21-46-08

mightyiam:
Closes #638

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -6,6 +6,9 @@ const ip = 'bs-local.com'
 
 const browserstack = require('./browserstack-karma.js')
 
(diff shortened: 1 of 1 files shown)


mightyiam:
It doesn't seem like a big time saver if judging by a single build... but it does mention "concurrency 5" where previously it said "concurrency 1" in the log.

PR #671, opened 2020-06-14
dx(git): ignore each test artifact specifically

mightyiam:


kuraga:
@mightyiam , why?

mightyiam:
@kuraga I like knowing when an unexpected build artifact is present.

PR #688, opened 2020-06-17
docs(style): `transition-property:all` not supported

### LINKED ISSUE #38: Some nodes not delay-removed when `transition-property: all`
While doing some tests, I noticed that DOM nodes don't get removed after transitioning when I define the transition-property as 'all' instead of the more specific properties (I know it's lower perf, but I like being lazy during testing).

Is it correct to assume that checking the specific property transitionend events is what allows the library to wait for all of them to complete before node removal?

mightyiam:
Closes #38.

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,6 +460,8 @@ h('span', {
 
 This makes it easy to declaratively animate the entry of elements.
 
+The `all` value of `transition-property` is not supported.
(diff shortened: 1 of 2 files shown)


paldepind:
Looks good to me 😄

PR #689, opened 2020-06-18
Removed duplicate 'h' import in README.md

grimalschi:
It looks like there is two imports with no reasons for this


paldepind:
Oops 😄 . Thanks for the PR 👍

mightyiam:
@paldepind the CI build failed due to the non-compliant commit message. Not a big deal. But it's easy to compose a new one if using squash and merge.

paldepind:
Ah, ok. Thanks for letting me know.

PR #715, opened 2020-06-27
ux(package): remove stale directories

niklasf:
These directories are not actually included in the package.


mightyiam:
Thank you, @niklasf.

PR #729, opened 2020-07-10
fix(scope):  Fix error when deleting the last element

2604150210:
When deleting the last data, an error will be reported: Cannot read property 'offset' of undefined


2604150210:
fix(scope): Fix error when deleting the last element

PR #731, opened 2020-07-10
auto(eslint): error in deleting the last element

2604150210:
When deleting the last data, an error will be reported:Cannot read property 'offset' of undefined


kuraga:
1. Let's use ternary operator instead?
2. Add a test, please?

mightyiam:
Thank you for this pull request, @2604150210.

Since there is no existing test coverage, I would not expect this pull request to add a test. Manual testing should suffice. And I will review this when I get around to it.

examples/reorder-animation/index.js
```diff
@@ -90,7 +90,11 @@ function render () {
     m.offset = last ? last.offset + last.elmHeight + margin : margin
     return acc.concat(m)
   }, [])
-  totalHeight = data[data.length - 1].offset + data[data.length - 1].elmHeight
+  if (data.length === 0) {
+    totalHeight = 0
+  } else {
+    totalHeight = data[data.length - 1].offset + data[data.length - 1].elmHeight
+  }
```
mightyiam, line 97: Thank you, @2604150210. Just changing this to a ternary, if you don't mind. Awesome change. Good catch.

```suggestion
  totalHeight = data.length === 0
    ? 0
    : data[data.length - 1].offset + data[data.length - 1].elmHeight
```

PR #783, opened 2020-08-13
chore(karma): decouple webpack from karma

patomation:


package.json
```diff
@@ -58,7 +59,6 @@
     "karma-firefox-launcher": "1.3.0",
     "karma-mocha": "2.0.1",
     "karma-mocha-reporter": "2.2.5",
-    "karma-webpack": "4.0.2",
```
mightyiam, line 61: Why do we do these things to ourselves as developers? Why couple karma and webpack? Build the pipeline yourself. Be happy. Know what is going on. That's how I feel.

PR #804, opened 2020-08-24
feat(eventlisteners): remove loaded listeners feature

### LINKED ISSUE #802: Ambiguity in eventlisteners API
```js
const vnode = h('div', { on: { click: [
  function a () {},
  function b () {},
] } })
```

According to the documentation, it seems that `a` and `b` could be one of
1. two different event listeners
2. one loaded / carrier event listener, which would be called as `a.call(vnode, b, event, vnode)`

patomation:
BREAKING CHANGE: loaded/carrying event listeners are no longer supported.
For alternatives see issue #802.

Fixes #802

### DIFF
--- a/src/package/modules/eventlisteners.ts
+++ b/src/package/modules/eventlisteners.ts
@@ -12,22 +12,9 @@ function invokeHandler (handler: any, vnode?: VNode, event?: Event): void {
     // call function handler
     handler.call(vnode, event, vnode)
   } else if (typeof handler === 'object') {
-    // call handler with arguments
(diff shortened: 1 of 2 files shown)


mightyiam:
I'll give a few days for comments from users on this breaking change.

mightyiam (approved):

kuraga:
:+1: . I think we have to follow semantic versioning and publish 2.0 on breaking change.

mightyiam:
@kuraga definitely major version.

PR #829, opened 2020-09-05
Support custom elements v1

### LINKED ISSUE #141: Support custom elements
Hi, do you have plans for custom elements support?

I gave a look at the code and should be a painless implementation (just handle the `is` attribute and add an argument to the `createElement` functions).

If you glad, I can submit a PR.
Thanks

the-dr-lazy:
This PR resolves #141.


paldepind:
Looks good.

What about putting the `is` attribute in the "root" of the data object instead of under `attrs`. This thing is used differently than all the other properties on the `attrs` property. Additionally, the attributes module will pick the `is` attribute up as well and set it as an attribute, which, as far as I can tell, is unnecessary.

mightyiam:
I need to study about custom elements.

the-dr-lazy:
> That is, if is is the only interesting property? Are there others? If there are we should probably add a new property to the root with those properties as "subproperties".

There is no other options in `document.createElement` based on current [W3C standard for custom elements v1](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement).

the-dr-lazy:
Is there any issue in tests? Looks like they are not consistent on my machine (maybe resource limitation!).

mightyiam:
> Is there any issue in tests? Looks like they are not consistent on my machine (maybe resource limitation!).
> BTW I can run the unit/core.ts in isolation without any issue.

CI for PRs runs only after explicit approval.
https://github.com/snabbdom/snabbdom/blob/master/CONTRIBUTING.md#allowing-ci-builds-of-pull-requests-from-forks

the-dr-lazy:
> CI for PRs runs only after explicit approval.

So if I understood correctly, there is nothing for me to do. The PR should become trusted by the members of the repository, Am I right?

mightyiam:
> > CI for PRs runs only after explicit approval.
> 
> So if I understood correctly, there is nothing for me to do. The PR should become trusted by the members of the repository, Am I right?

That's correct.

paldepind:
This PR looks good to me.

@mreinstein, you mentioned support for custom elements in another thread. Does this look good to you too?

mreinstein:
@paldepind it looks ok, but I still have that question around using `vnode1.data?.is === vnode2.data?.is` Does the optional chaining get compiled out by whatever typescript config is currently in place? Because 15% of the browsers don't support this yet. Assuming it gets transpiled down to something like es6 then LGTM.

jvanbruegge:
Yes, optional chaining is transpiled

paldepind:
@mreinstein Yes, as @jvanbruegge says it gets transpiled away. Also, we have been using `?.` for 1-2 years now, so it's nothing new to this PR.

@the-dr-lazy Thanks a lot for the PR. Please resolve the conflicts and then we'll merge it 🙂

src/package/vnode.ts
```diff
@@ -33,6 +33,7 @@ export interface VNodeData {
   ns?: string // for SVGs
   fn?: () => VNode // for thunks
   args?: any[] // for thunks
+  is?: string // for custom elements v1
```
paldepind, line 36: What does the `v1` here mean? Perhaps it should be removed?

src/package/init.ts
```diff
@@ -16,8 +16,20 @@ type VNodeQueue = VNode[]
...
+  return vnode1.sel === vnode2.sel
+}
+
+function sameVnodeKey (vnode1: VNode, vnode2: VNode): boolean {
+  return vnode1.key === vnode2.key
+}
+
+function sameVnodeIs (vnode1: VNode, vnode2: VNode): boolean {
+  return vnode1.data?.is === vnode2.data?.is
+}
```
paldepind, line 29: Since these three functions are all only used a single time I think it would be better to not have them.

src/package/init.ts
```diff
@@ -109,8 +121,8 @@ export function init (modules: Array<Partial<Module>>, domApi?: DOMAPI) {
       const dot = dotIdx > 0 ? dotIdx : sel.length
       const tag = hashIdx !== -1 || dotIdx !== -1 ? sel.slice(0, Math.min(hash, dot)) : sel
       const elm = vnode.elm = isDef(data) && isDef(i = data.ns)
-        ? api.createElementNS(i, tag)
-        : api.createElement(tag)
+        ? api.createElementNS(i, tag, vnode.data)
+        : api.createElement(tag, vnode.data)
```
paldepind, line 125: Here you can just use `data` as defined above instead of `vnode.data`.

src/test/unit/core.ts
```diff
@@ -181,6 +190,11 @@ describe('snabbdom', function () {
       assert(elm.firstChild.classList.contains('has'))
       assert(elm.firstChild.classList.contains('classes'))
     })
+    it('can create custom elements', function () {
+      var vnode1 = h('p', { is: 'p-a' })
```
paldepind, line 194: Please use `const` instead of `var` here.

src/test/unit/core.ts
```diff
@@ -344,6 +358,15 @@ describe('snabbdom', function () {
       patch(vnode1, vnode2)
       assert.strictEqual((elm as any).a, 'foo')
     })
+    it('handles changing is attribute', function () {
+      var vnode1 = h('p', { is: 'p-a' })
+      var vnode2 = h('p', { is: 'p-b' })
```
paldepind, line 363: Please use `const` for these two variables.

paldepind (commented):
Looks really good @the-dr-lazy. I've added a few comments on some minor things.

src/package/vnode.ts
mreinstein: I assume it's referring to the spec (custom elements v0 is deprecated) https://caniuse.com/?search=custom%20elements

src/package/init.ts
mreinstein: these nested ternaries are very hard for humans to read. At least, this human struggles with reading them. :)

src/package/vnode.ts
the-dr-lazy: Yeah, it's for spec.

paldepind (approved):
LGTM 👍

the-dr-lazy:
Thank you.

src/package/init.ts
```diff
@@ -17,7 +17,11 @@ type VNodeQueue = VNode[]
 const emptyNode = vnode('', {}, [], undefined, undefined)
 
 function sameVnode (vnode1: VNode, vnode2: VNode): boolean {
-  return vnode1.key === vnode2.key && vnode1.sel === vnode2.sel
+  const isSameKey = vnode1.key === vnode2.key
+  const isSameIs = vnode1.data?.is === vnode2.data?.is
+  const isSameSel = vnode1.sel === vnode2.sel
+
+  return isSameSel || isSameKey || isSameIs
```
alfonsogarciacaro, line 24: Is this intended? Actually I wanted to write a thunk version that lets me change `sel` on the init hook, so it'd be great if key equality is enough when they're both defined. But I was surprised to find this when checking the current code: using `||` here means two nodes with same selector will be considered the same even if they have different keys, which doesn't look right.

the-dr-lazy: No. I think this was a mistake introduced by 39c9da4615d248265ff8de376a95adab799ca981.

alfonsogarciacaro: Thanks to you for adding the feature and the quick reply!

PR #885, opened 2020-10-15
Fix style.ts for headless environments

chekoopa:
The current state of style module can crash if running in some environments which have no `window.requestAnimationFrame` function. So, we should add an additional check for it.


paldepind:
Thanks for the PR @chekoopa and for reporting this bug. The PR #1082, which is based on this PR and which fixes the same issue, has been merged.

chekoopa:
That's neat, though I've already forgot 😅

PR #886, opened 2020-10-16
add prettier config

wclr:
This adds prettier config and formats files. 

Passes linting. Some issues with tests, that actually should not be related to this PR.


paldepind:
Looks good! I'm a big fan of Prettier so this change is fine for me.

It might be nice to include `"prettier": {}` in the configuration file. Just because some formatters can be configured to only run Prettier if a configuration is present (and an empty config is fine).

whitecolor:
> "prettier": {} in the configuration file.

In what configuration file?

I added config in [package.json](https://github.com/snabbdom/snabbdom/pull/886/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R85)
```json
"prettier": {
   "semi": false,
    "singleQuote": true
}
```

because these settings seemed to be in compliance with the existing codebase (no semicolons and single quote mostly).

paldepind:
@whitecolor Yes, in the package.json is exactly what I meant. I think we should just use the default Prettier config. I.e. `prettier: {}`. I don't think we use set the `"semi"` option to `false`. That seems to remove a lot of existing semicolons.

whitecolor:
@paldepind so you want default config with semicolons and double quotes, right?

whitecolor:
@paldepind made prettier default config

whitecolor:
@paldepind hello

paldepind:
Hi @wclr. Thanks a lot for working on this. I apologize for dropping the ball on this.

The PR looks good. I'd appreciate it if you could fix the conflicts and then I will merge 👍

wclr:
@paldepind I've made a [new version of PR](https://github.com/snabbdom/snabbdom/pull/945) because it turns to be easier than resolving conflicts with the recent updates.

cc @mightyiam please check existing (hanging) PRs before planning updates changes.

mightyiam:
Didn't mean to cause more work.

PR #897, opened 2020-11-06
remove use of loaded/carrying event listeners in examples

xifeiwu:


mreinstein:
thanks for the PR! The examples have been updated to remove loaded events recently, obsoleting these changes.

PR #903, opened 2020-12-03
Fix webpack tests

wclr:
Based on https://github.com/snabbdom/snabbdom/pull/886

This fixes running webpack tests. It was not working because of wrong globby usage.


jvanbruegge:
closing in favor of #948 
Webpack is not needed any more then

PR #907, opened 2020-12-10
fix(examples):Click the sort button without responding

EliChen95:
There is no response after clicking the sort button since the loaded listener feature has been removed.


paldepind:
Thanks a lot @EliChen95. This PR looks good. Unfortunately, I saw #923, which does the same thing, before this one.

PR #923, opened 2021-02-02
chore(examples): update examples

gzg1023:
Hi , when I was learning snabbdom, I ran the examples, but found that it was a project based on the old grammar, and some functions could not be run.
I adjusted the example according to the new grammar. Can better help everyone learn


paldepind:
Thank you @gzg1023. Looks good :)

Btw, the documentation for the eventlisteners module is outdated as well.

Issue #640, opened 2020-05-17
Is there any place where we catalog snabbdom-compatible UI components?


mreinstein:
I'm starting to build up a catalog of pure snabbdom based UI elements, and I'm wondering if:

a) we have any place where we collect/share these
b) anyone else out there building snabbdom elements that are shareable

to give an example, here is one I've been working on very recently, exposing a chrome devtools inspired, re-usable timeline component:

https://github.com/mreinstein/snabbdom-timeline

If we don't have one, maybe something like https://bit.dev/ could be used to curate a collection of high quality ui components?  Thoughts?

mreinstein:
some more modules:

https://github.com/mreinstein/snabbdom-drawer

https://github.com/mreinstein/snabbdom-property

mreinstein:
https://github.com/mreinstein/snabbdom-tabs

mreinstein:
https://github.com/mreinstein/snabbdom-valve

paldepind:
Those can be added to the readme in [this section](https://github.com/snabbdom/snabbdom#structuring-applications). It would probably be a good idea to re-structure that section with a few sub-sections etc.

mightyiam:
How about in a wiki here in the repo?

mreinstein:
The github wikis don't get much traffic and they often end up being neglected. It might be better to just put them in a markdown file in the project root.

mreinstein:
https://github.com/mreinstein/snabbdom-knob

PR #946, opened 2021-02-17
docs(docs): readme section on related packages

mightyiam:


mightyiam:
@mreinstein do you feel this would close #640 ?

paldepind:
I think we should collapse this with the current "Structuring Applications" section. What I mean is, keep the text that this PR adds, rename the "Structuring Applications" section to "Related Projects" to broaden its scope, and then perhaps sort the list into a few rough categories.

mreinstein:
@paldepind @mightyiam I think for the sake of making the readme shorter and easier to read, we should move this stuff to a separate markdown doc. Maybe just like `related-projects.md`.

Making the initial documentation shorter is almost always a good thing. :)

paldepind:
I think mentioning the npm tag and linking to the search results for the tag on npm is nice. I still agree with my comment above, but merging this PR for now as it's still an improvement.

Issue #517, opened 2019-12-24
Remove the hero module


mightyiam:
Out of all of the [modules](https://github.com/snabbdom/snabbdom/tree/v0.7.4/src/modules) to me stands out the hero module as being too specific. I feel it shouldn't be in this project.

mreinstein:
maybe we could move this to the `examples/hero/` directory. I think it's a reasonable module, I just suspect that it's not something people are actually using in their own code. Probably shouldn't be a built-in module.

mightyiam:
Yes. Move to examples. Good idea.

Issue #913, opened 2021-01-10
RFC: Export everything from the main 'snabbdom' module


jvanbruegge:
When updating dependencies, I (as have many others) stumbled across the issue of #723, in my case with typescript itself.
Is there any reason other than historical to split the package into multiple files that need to be imported seperately?
Why not simply:
```ts
import { VNode, init } from 'snabbdom';
```
IMO, this would be the cleaner API and it would resolve the issues with bundlers/typscript not supporting the `export` field

jvanbruegge:
Wow, even better, TypeScript does not support `export`, so I have to use `build/package/h`, but mocha does so it complains that `build/package/h` is not in the exports field, making it impossible to import in a test

kuraga:
Issues listed in https://github.com/snabbdom/snabbdom/issues/516#issue-541125637 are related.

wclr:
It's time to fork snabbdom ;)

jvanbruegge:
Incidentally, I am just starting with the new Cycle.js Neo DOM driver :thinking:

mreinstein:
I ran into this too. the first thing I tried doing was `import snabbdom from 'https://cdn.skypack.dev/snabbdom'` and that failed with a big scary error message. It would be _really_ nice if the default thing that everyone will try to import actually worked out of the box.  It sounds like all we need to do is add a `".":` as the first entry in the `exports` section of `package.json`.

https://github.com/skypackjs/skypack-cdn/issues/125

mreinstein:
see the first line of the `"exports": {` block from the node docs: https://nodejs.org/api/packages.html#packages_package_entry_points

mreinstein:
This is not a disruptive change but it does technically break the API, so it makes sense to do this in the 3.0.0 release.

mightyiam:
There decision to go with the `exports` field is solid. TypeScript can be made to behave, by defining some `paths`:
https://www.typescriptlang.org/tsconfig#paths

It covers for the lack of `exports` field support.

Please try that, @jvanbruegge, and let me know. I'll assume this answers your problem and close this.

mreinstein:
@mightyiam come on man, you can't just disappear, drop the project, and then come back and start making unilateral decisions.

mightyiam:
There's one problem here, and it is the lack of support for the `exports` field in TypeScript. A reasonable solution could be to document the `paths` workaround in the readme. That should take care of the issue. @mreinstein would you like to do that?

[...]

Issue #932, opened 2021-02-15
isDef?


mreinstein:
At the risk of starting a flamewar (I definitely do _not_ want to start one,) I can't help but look at some parts of the code written in typescript and wonder why we are doing this. For example:

```typescript
type NonUndefined<T> = T extends undefined ? never : T

function isUndef (s: any): boolean {
  return s === undefined
}
function isDef<A> (s: A): s is NonUndefined<A> {
  return s !== undefined
}
```

It seems to me like a lot over overhead, compared to doing `if (s === undefined)`. Typescript could be limiting contributor-ship.

Again, I'm not trying to start a fight or demand that we change this. I'm just taking the temperature to see what the general sentiment on this is. I don't know if it's just me that finds it to be over-engineered or if others feel that way too?

wclr:
How do you want to rewrite this?

```ts
    if (isUndef(vnode.text)) {
      if (isDef(oldCh) && isDef(ch)) {
        if (oldCh !== ch) updateChildren(elm, oldCh, ch, insertedVnodeQueue);
      } else if (isDef(ch)) {
        if (isDef(oldVnode.text)) api.setTextContent(elm, "");
        addVnodes(elm, null, ch, 0, ch.length - 1, insertedVnodeQueue);
      } else if (isDef(oldCh)) {
        removeVnodes(elm, oldCh, 0, oldCh.length - 1);
      } else if (isDef(oldVnode.text)) {
        api.setTextContent(elm, "");
      }
    } else if (oldVnode.text !== vnode.text) {
      if (isDef(oldCh)) {
        removeVnodes(elm, oldCh, 0, oldCh.length - 1);
      }
      api.setTextContent(elm, vnode.text!);
    }
```

Functions are used to make code terser and easier to read. As for performance, I don't think that it does really matter.

> Typescript could be limiting contributor-ship.

This is funny. You should learn typescript, this is just lightweight types over js.

mightyiam:
There may be pieces of TypeScript in this codebase that are not as modern as they can be, because they were written for older versions of both TypeScript and JavaScript. Improvements and even pointers are welcome.

mightyiam:
But we're not ditching TypeScript, obviously. It helps with type safety for this project and some of its consumers. They would not be happy about us ditching TypeScript. And, if anything, TypeScript seems quite popular these days. So I can't imagine it deterring contribution more than attracting it.

mreinstein:
> How do you want to rewrite this?

[...]

Issue #723, opened 2020-07-02
 Cannot resolve de pendency 'snabbdom/init'


f43310:
import { init } from 'snabbdom/init'
node log file:
<pre>0 info it worked if it ends with ok
1 verbose cli [
1 verbose cli   'D:\\Program Files\\nodejs\\node.exe',
1 verbose cli   'D:\\Program Files\\nodejs\\node_global\\node_modules\\npm\\bin\\npm-cli.js',
1 verbose cli   'run',
1 verbose cli   'build'
1 verbose cli ]
2 info using npm@6.14.5
3 info using node@v12.18.2
4 verbose run-script [ 'prebuild', 'build', 'postbuild' ]
5 info lifecycle snabbdom-demo@1.0.0~prebuild: snabbdom-demo@1.0.0
6 info lifecycle snabbdom-demo@1.0.0~build: snabbdom-demo@1.0.0
7 verbose lifecycle snabbdom-demo@1.0.0~build: unsafe-perm in lifecycle true
8 verbose lifecycle snabbdom-demo@1.0.0~build: PATH: D:\Program Files\nodejs\node_global\node_modules\npm\node_modules\npm-lifecycle\node-gyp-bin;I:\temp\code\study\frantEnd\vuecode\snabbdom-demo\node_modules\.bin;D:\Program Files (x86)\NetSarang\Xshell 6\;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;E:\adb;C:\Program Files (x86)\NVIDIA Corporation\PhysX\Common;D:\Program Files\Git\cmd;F:\phpStudy\php72n\php-7.2.25-nts-Win32-VC15-x64\;C:\ProgramData\ComposerSetup\bin;D:\Program Files\MongoDB\Server\4.2\bin;M:\Program Files\OpenSSH\bin;D:\Program Files\nodejs\;C:\Users\linda\AppData\Local\Programs\Python\Python36\Scripts\;C:\Users\linda\AppData\Local\Programs\Python\Python36\;D:\Program Files\nodejs\node_global;J:\Program Files\JetBrains\PhpStorm 2018.3\bin;;J:\Program Files\Microsoft VS Code\bin;D:\Program Files\Fiddler;C:\Users\linda\AppData\Roaming\Composer\vendor\bin;C:\Users\linda\AppData\Roaming\npm
[...]

PR #948, opened 2021-02-17
New tooling

jvanbruegge:
This PR is to simplify the tooling for snabbdom. Roadmap for this PR:
- [x] Drop webpack
- [x] Drop custom typescript compiler plugin
- [x] Implement #913 (needed for simpler tooling, especially karma)
- [x] Offer cjs and esm bundles for easier use with nodejs
- [x] Just use karma for typescript tests
- [x] Add prettier
- [x] Port over jsx tests
- [ ] ~~Port over benchmark tests~~


jvanbruegge:
The CI won't pass, but the karma tests for all *.ts (not tsx) pass already locally

mightyiam:
@jvanbruegge you can approve deploys now.

karma.conf.js
```diff
@@ -15,25 +15,36 @@ const browsers = ci
...
     : watch
       ? ['Chrome']
-      : ['Chrome', 'Firefox']
+      : ['ChromeHeadless', 'FirefoxHeadless']
 
 module.exports = function (config) {
   config.set({
     basePath: '.',
-    frameworks: ['mocha'],
+    frameworks: ['mocha', 'karma-typescript'],
```
mightyiam, line 23: I don't like karma-typescript or most of these kinds of creations. I see them as unnecessary couplings. I like chaining my tools myself. It seems simpler to me and with less of the pipeline under the hood of some dependency.

mightyiam (commented):
Wow, this is a big change. It's like... a lot of my work is going down the drain. I don't agree with most of this. And this PR is a bit too large to review comfortably for me. But if you'd like to take the reigns on this project then...

jvanbruegge:
To make contributions easier and ease maintenance, I want to get rid of as much custom tooling as possible and replace it with standard tools. The main reason is that custom tooling has to be maintained by us, while more eyes are on the standard stuff.

For me, ideally we would just have `karma`, `typescript`, `prettier` and `rollup` and nothing else.

If you want to look more towards rust and only do small maintenance (which I completely understand), then yes, I'd like to take over the reigns, if you are ok with that. I won't start making major, publicly visible decisions on my own, but consult and discuss with the rest of the maintainers, be it you or @mreinstein. If this is fine by you, could you make me an owner on the snabbdom repo? I want to adjust CI a bit, so we can run a stripped down version (no browserstack, thus no secrets) even without approval

mreinstein:
I don't know if I'm alone in this feeling but similar to typescript, I find the output of `prettier` to be ironically very un-pretty. 

While I understand the herd's movement towards these automated tools, I do lament the decrease in readability they incur.

mreinstein:
>  It's like... a lot of my work is going down the drain.

@mightyiam  I would encourage you to re-think your perspective on this.  I don't think any of the maintenance stuff you've done has gone down the drain, it's just an ongoing task, and things inevitably change. The ecosystem has changed and it's more of an evolution than any kind of deletion.

Something separate to consider is that commonjs is on the way out. Every production version of node currently supports es modules except for node v10, and that loses maintenance support in April (see https://nodejs.org/en/about/releases/)

So while I don't think we should outright break commonjs support today, the writing is on the wall, it's on the way out. We should operate under the assumption that we need to support cjs today but that it's a temporary thing, and maybe a year from now it might be ok to just drop it (depending on how fully the ecosystem embraces it.)

mreinstein:
2 concerns:

1-  There is no included documentation around the changes to how one`require` / `import` s snabbdom. That should be included with any changes to that stuff, so the commit is atomic.

2- It's unclear to me why we need to run `rollup` for the es module build. Can't we just point at the es module directly, and only use `rollup` to produce a temporary `cjs` build? The nice thing about this is if/when commonjs becomes unneeded we can just drop rollup.

thoughts?

wclr:
> I find the output of prettier to be ironically very un-pretty.

This is just about standardization of the code convenience for developers when you should not care about formatting at all. For self-sufficient small lib, it really doesn't matter what auto-formatting is used, the thing that it should be used not to create additional work and hassle.

mreinstein:
> For self-sufficient small lib, it really doesn't matter what auto-formatting is used

I agree with this statement, in isolation. It is a small library, and given how critical it is, making it readable by humans is really, really high priority in my opinion.

While DOM diff/patching is not rocket science it is critically important and underpins several high value frameworks (vue, cyclejs, snabby.)

wclr:
> making it readable by humans is really, really high priority in my opinion

the prettier output is absolutly readable even with semicolons 😄

mreinstein:
> the prettier output is absolutly readable even with semicolons

Semicolons are a trivial non-issue. The horrific way it indents and spaces lines in some cases is of more concern.

jvanbruegge:
> There is no included documentation around the changes to how onerequire / import s snabbdom

Updated the README now in the same commit

>  Can't we just point at the es module directly, and only use rollup to produce a temporary cjs build?

Did that now too

README.md
```diff
@@ -157,8 +152,7 @@ takes a list of modules and returns a `patch` function that uses the
 specified set of modules.
 
 ```mjs
-import { classModule } from 'snabbdom/modules/class'
-import { styleModule } from 'snabbdom/modules/style'
+import { classModule, styleModule } from 'snabbdom'
```
mreinstein, line 155: In the interests of making all of the example code sections smaller and more readable, I think we can omit all of the `import` statements _after_ the first example. We don't need to demo the import process in every example snippet.

src/index.ts
```diff
@@ -0,0 +1,30 @@
...
+// types
+export * from './hooks';
+export { Module } from './modules/module';
+
+// modules
+export { Attrs, attributesModule } from './modules/attributes';
+export { Classes, classModule } from './modules/class';
+export { Dataset, datasetModule } from './modules/dataset';
+export { On, eventListenersModule } from './modules/eventlisteners';
+export { Hero, heroModule } from './modules/hero';
```
mreinstein, line 28: let's kill this thing immediately. Will make the module a little smaller and also resolve https://github.com/snabbdom/snabbdom/issues/517

jvanbruegge: Wanted to do that in a seperate PR

mreinstein: I guess that's fine, since we're not actually pushing this out to a new npm version anyway. Just figured it's a little silly to carry forward this piece we know we don't want. but it's not a big deal.

jvanbruegge:
So, eslint is back at it, now together with prettier. Next step for me is fixing the JSX stuff

mightyiam:
Could we do small PRs, instead of this?

mightyiam:
Seriously. Is the tooling that I perceive as superbly set up so aweful that they must be replaced in one fell swoop?

mreinstein:
@jvanbruegge thank you for pushing these changes forward! I don't agree with everything but I'm glad to see the project evolve.

kuraga:
Drop karma, too would be brilliant :)

kuraga:
> Drop karma, too would be brilliant :)

BTW, I meant to use something lightweightweight like [tman](https://github.com/thunks/tman).

jvanbruegge:
karma != mocha
karma allows us to run the tests in the browser fairly easily, even with Browserstack

jvanbruegge:
> Could we do small PRs, instead of this?

I don't want to leave the master in a broken state. IMO the commits are cleanly separated regardless if they sit in 4 or 1 PR

kuraga:
> karma != mocha
> karma allows us to run the tests in the browser fairly easily, even with Browserstack

Tman too, but not so easy. But lightweight :)

paldepind:
Thank you for working on this @jvanbruegge.

> To make contributions easier and ease maintenance, I want to get rid of as much custom tooling as possible and replace it with standard tools. The main reason is that custom tooling has to be maintained by us, while more eyes are on the standard stuff.

I agree with this sentiment. If we can get an equally good result for end users but while mainting fewer lines of configuration then that seems like a net benefit to me.

I agree with @mightyiam that this PR is a bit difficult to review given the number of things changed. It also makes it harder for people to voice their opinions on individual things. But, I also understand that doing this at once is probably easier for you @jvanbruegge. If it's not a big problem I think it would be nice to split this up a bit, but alternatively you could maybe just summarise briefly the various changes?

> Seriously. Is the tooling that I perceive as superbly set up so aweful that they must be replaced in one fell swoop?

I don't think anyone considers the current tooling to be awful 🙂. I think your should share your opinion on the changes supported with technical arguments, then we can all take them into consideration and decide on what is best going forward.

paldepind:
I think this seems very reasonable overall. And most of it does seem to simplify the tooling.

The only part of the PR that I'm a bit worried about is the "export everything from 'snabbdom'" part. I've added a comment regarding that over in the related issue https://github.com/snabbdom/snabbdom/issues/913#issuecomment-786480470.

kuraga:
> karma != mocha
> karma allows us to run the tests in the browser fairly easily, even with Browserstack

One more comment about Tman. I know we don't need it here (as tooling works already) but for your knowledge.

It's brilliant because it's lightweight (as such as Snabbdom :smile: ). It's not mocha, it's just (only!) test runner.

To run tests in browser I [just use Rollup](https://github.com/kuraga/rotorjs/blob/master/scripts/testBrowser.js) without big issues (but yes sometimes they exist).

wclr:
I'm on board with this. The tooling that is currently used for tests etc doesn't seem to be well-tuned, some times tests fail locally. And overall config seemed to be a little bit complicated.

paldepind:
@wclr

> Easy to ask when you are not performing the task.

Please stop the aggressive attitude and be nice to everybody here. I will not tolerate comments like that and this is not the first one (https://github.com/snabbdom/snabbdom/issues/932#issuecomment-779917547). @mightyiam has made tremendous contributions to Snabbdom and was making a perfectly reasonable comment.

Everybody, please keep the comments constructive and on-topic. Thank you 🙏

paldepind:
To be clear. I think we (and by "we" let's acknowledge that @jvanbruegge is the one doing the work) should absolutely continue with this PR. It brings a bunch of great changes with it.

I think we should do a release after this PR is merged. Exporting everything from the main module is going to be a small breaking change, but it will also simplify the whole import process (thanks to tree-shaking) and fix issues such as #723.

jvanbruegge:
Yeah, I will continue, I just have an exam today and on Friday, so that kind of hogged my free time a bit 😅

paldepind:
@jvanbruegge No problem. It was not to rush you in any way 😄. My comment was to make sure that we all agreed on that and to make sure you didn't feel discouraged.

Good luck with your exam 🤞

jvanbruegge:
I finally found out how typescript wants you to JSX typings. Now there is no global declarations file, but instead it is added to the factory function itself.

jvanbruegge:
The benchmark tests test the current version against the last released snabbdom version. For now I would not port those over, until we have release v3.0 because there are two issues:
- The way snabbdom exports stuff, makes it hard to get it to work with karma
- After releasing v3 all the workarounds would be useless and I don't really want to waste a lot of time there now
Is that ok?

paldepind:
@jvanbruegge 

> The benchmark tests test the current version against the last released snabbdom version. For now I would not port those over, until we have release v3.0 because there are two issues: [...]

Yes, that seems very reasonable to me.

jvanbruegge:
I've swapped `commitlint` for my new package `commithelper`. The main reason is that commitlint can't provide the user with a convenient prompt to create a fitting commit message, while commitizen is not able to lint an existing commit message. commithelper does both.

jvanbruegge:
I've rebased and reworded all commit messages to match the linting. This includes adding explainatory bodies and mentioning issues closed and breaking changes

jvanbruegge:
@paldepind Can you make me owner of the snabbdom repo? Always having to approve CI is annoying

jvanbruegge:
So, this would be done from my side except for CI which I can't do until I have owner access to this repo. Once version 3.0 is released I will tackle the benchmark tests as explained earlier.
Some statistics of this PR:
|                      | master | new tooling | difference |
|----------------------|--------|-------------|------------|
| config files         | 12     | 10          | -2         |
| direct dependencies  | 52     | 40          | -12        |
| overall dependencies | 1483   | 1069        | -414       |

jvanbruegge:
ping @paldepind

mreinstein:
@jvanbruegge great work, man! Thanks for doing this!

paldepind:
@jvanbruegge Sorry for the delay. I've made you an owner of the Snabbdom organization and sent you an invitation to become a maintainer of the Snabbdom package on npm. On a related note, do you know if the ["cycle" npm account](https://www.npmjs.com/~cycle) is used for anything? Or can we remove that account from the npm package?

PR looks great to me! 👍

jvanbruegge:
Thanks!
I am not aware that the cycle account is used, it should be save to remove it

paldepind:
> I am not aware that the cycle account is used, it should be save to remove it

Thanks, great. I've removed it.

jvanbruegge:
So, CI works, runs on every commit for every PR now. I was also able to reactivate old browsers in the browserstack config. @paldepind @mreinstein please do a final review

Just a side note, this PR closes 8 open issues :tada:

jvanbruegge:
Once this is merged I can release version 3.0

package.json
```diff
@@ -2,203 +2,112 @@
   "name": "snabbdom",
   "version": "2.1.0",
   "description": "A virtual DOM library with focus on simplicity, modularity, powerful features and performance.",
-  "type": "module",
```
mreinstein, line 5: I'm a little sad to see this. `type: "module"` is the node sanctioned way of declaring this module to be es based. Going back to cjs feels a bit regressive. :(

mreinstein (approved):

package.json
jvanbruegge: The problem is AFAICT that node will never pick the entry from `module`, always `main`. So if we set `type: module` we would need to throw out cjs completely. And at least I have often issues with testing and ES modules

paldepind:
Thanks a lot for doing this work @jvanbruegge.

jvanbruegge:
@paldepind would it be possible to undo the merge and do a merge commit? having one gigantic commit on master that touches almost 100 files is not really helpful. Of you want I can also do it

jvanbruegge:
Oh, I also see that this totally breaks the changelog because the commit messages do not have subject fields any more. I will change this now

kuraga:
🥳

PR #949, opened 2021-03-04
Fix node similarity checking bug

the-dr-lazy:
The bug introduced by 39c9da4615d248265ff8de376a95adab799ca981 on #829 and reported by @alfonsogarciacaro on https://github.com/snabbdom/snabbdom/pull/829#discussion_r587475354.


paldepind:
Fantastic that you spotted this @alfonsogarciacaro and thanks for the quick fix @the-dr-lazy 👍

PR #953, opened 2021-03-28
Remove the hero module

jvanbruegge:
As discussed, it is too specific. I moved it to the examples instead


mreinstein (approved):
looks great!

PR #954, opened 2021-03-28
feat: allow symbols as keys

jvanbruegge:


mreinstein (approved):
This is a breaking change. Might want to do this in v3 too, while we're updating the API pretty drastically.

jvanbruegge:
Why is this a breaking change? It only allows you to also use symbols as keys. 100% of the previous functionality is still there

mreinstein:
> Why is this a breaking change? 

What happens if you do `h("span", { key: Symbol() })` prior to this commit? My assumption is it would throw an Error. After this commit, it will no longer throw an error on that input.

Probably this is never relied upon in user facing code, but technically this does change the behavior of the public API.

jvanbruegge:
in typescript you would have gotten a Compiler error, but in javascript it would have worked. In the PR I only change types and add a test

kuraga:
> My assumption is it would throw an `Error`.

No, unless there docs or tests about this (or this is a bug). Until this it would be an undefined behavior.

mreinstein:
It's a moot point since we're adding this into a major version which makes api breaking changes anyway.

jvanbruegge:
Not completely, as a description of a breaking change has to be includen in the commit message, so it appears in the release notes and changelog

jvanbruegge:
but yes, I released v3.0.0 with this included now

PR #951, opened 2021-03-07
Resolve checkbox race condition

Fresheyeball:
We should check if the key is "checked" as well as "value" to see if the dom got out of sync.

Resolves: #950

PR #955, opened 2021-04-10
update checked property based on current element value. fixes #950

### LINKED ISSUE #950: Race condition between native browser state and VDom
When hitting a checkbox, there is race condition where the native browser state will set the checkbox to `checked: true` and Snabbom fails to set `checked: false`. Here is a minimal complete example

```javascript
var inBetween; // holds the intermediate vnode

const vnode1 = hh('div', {}, [ // vnode1 has the div with 2 checkboxes set to false
  hh('input', { on : { change : () => patch(inBetween,vnode2)}, // when the first checkbox is changed, we change to vnode2
                props : { checked : false, type : "checkbox" } }, []),
  hh('input', { props : { checked : false, type : "checkbox" } }, [])
])

const vnode2 = hh('div', {}, [
  hh('input', { props : { checked : false, type : "checkbox" } }, []) // vnode2 has the div containing 1 checkbox set to false
])

inBetween = patch(container, vnode1); // works as expected
```

## Reproduction

You can see this bug occur live here: https://plnkr.co/edit/TyvcMBT0F3TQ5YDq

```javascript
document.getElementsByTagName("input").length == 2
document.getElementsByTagName("input")[0].checked == false
document.getElementsByTagName("input")[1].checked == false
```

1. Click the first checkbox

## Expected Result

```javascript
document.getElementsByTagName("input").length == 1
document.getElementsByTagName("input")[0].checked == false
```

## Actual Result

```javascript
document.getElementsByTagName("input").length == 1
document.getElementsByTagName("input")[0].checked == true
```

mreinstein:
This is the same exact logic that @Fresheyeball added in #951 but refactored in an attempt to make the logic more clear around _why_ the `value` and `checked` properties are treated a bit differently.

### DIFF
--- a/src/modules/props.ts
+++ b/src/modules/props.ts
@@ -25,8 +25,6 @@ function updateProps(oldVnode: VNode, vnode: VNode): void {
     cur = props[key];
     old = oldProps[key];
 
-    shouldUpdate = false;
-
     // DOM INPUT.value should never be updated when it matches the new vnode.
     // typing into an INPUT with a sync event handler cause a render (such as for
     // validation feedback) where setting INPUT.value also pushes the cursor to the end
--- a/src/modules/props.ts
+++ b/src/modules/props.ts
@@ -7,21 +7,44 @@ function updateProps(oldVnode: VNode, vnode: VNode): void {
   let key: string;
   let cur: any;
   let old: any;
+  let shouldUpdate: boolean;
   const elm = vnode.elm;
   let oldProps = (oldVnode.data as VNodeData).props;
   let props = (vnode.data as VNodeData).props;
 
-  if (!oldProps && !props) return;
-  if (oldProps === props) return;
+  if (!oldProps && !props)
+    return;
+
+  if (oldProps === props)
+    return;
+
   oldProps = oldProps || {};
   props = props || {};
 
   for (key in props) {
     cur = props[key];
     old = oldProps[key];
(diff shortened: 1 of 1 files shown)


jvanbruegge:
I did not know that there is not way to pass secrets to PRs from forks. Well then I guess it will be an env var instead (no real risk with the browserstack access key, the snabbdombot user has no admin privileges)

jvanbruegge:
Wait, there is! Apparently I need to user `pull_request_target` instead of `pull_request`.

jvanbruegge:
can you update your fork to the latest master and try again?

mreinstein:
This repo is really a pain in the ass to work in with all of these automated guard systems. I'm trying to send a clean PR, and it won't let me commit. Husky seems to be failing now?

Screen Shot 2021-04-13 at 8 57 16 AM

mreinstein:
I've just deleted `.husky` because it's failing for some reason, and that at least lets me commit/push. Will open a PR for this shortly

jvanbruegge:
The commits don't follow the conventions (which we use to generate a changelog). If you use `git commit` without `-m` you will get help with creating a valid commit message

jvanbruegge:
wait, why does your npm not have an `exec` command? That is the issue

jvanbruegge:
i see, apparently that command was added in npm v7. I will change the scripts to not use `exec` then

jvanbruegge:
fixed on master

mreinstein:
closing in favor of #957

PR #957, opened 2021-04-13
update checked property based on current element value. fixes #950

### LINKED ISSUE #950: Race condition between native browser state and VDom

mreinstein:
This is the same exact logic that @Fresheyeball added in #951 but refactored in an attempt to make the logic more clear around why the value and checked properties are treated a bit differently.

### DIFF
--- a/src/modules/props.ts
+++ b/src/modules/props.ts
@@ -7,6 +7,7 @@ function updateProps(oldVnode: VNode, vnode: VNode): void {
   let key: string;
   let cur: any;
   let old: any;
+  let shouldUpdate: boolean;
   const elm = vnode.elm;
   let oldProps = (oldVnode.data as VNodeData).props;
   let props = (vnode.data as VNodeData).props;
@@ -19,9 +20,25 @@ function updateProps(oldVnode: VNode, vnode: VNode): void {
   for (key in props) {
     cur = props[key];
     old = oldProps[key];
-    if (old !== cur && (key !== "value" || (elm as any)[key] !== cur)) {
+
+    // DOM INPUT.value should never be updated when it matches the new vnode.
+    // typing into an INPUT with a sync event handler cause a render (such as for
+    // validation feedback) where setting INPUT.value also pushes the cursor to the end
+    // of the input. fixes #71, partially fixes #63
+    if (key === 'value')
+      shouldUpdate = (old !== cur) && ((elm as any)[key] !== cur);
+
+    // DOM INPUT.checked is a boolean property. The element's internal value should be
+    // used for comparison. fixes #950
+    else if (key === 'checked')
+      shouldUpdate = (elm as any)[key] !== cur;
+
+    else
+      shouldUpdate = (old !== cur);
+
+
+    if (shouldUpdate)
       (elm as any)[key] = cur;
-    }
   }
 }
 


jvanbruegge:
Good, CI works now also for forks

src/modules/props.ts
```diff
@@ -19,9 +20,25 @@ function updateProps(oldVnode: VNode, vnode: VNode): void {
...
+    // typing into an INPUT with a sync event handler cause a render (such as for
+    // validation feedback) where setting INPUT.value also pushes the cursor to the end
+    // of the input. fixes #71, partially fixes #63
+    if (key === 'value')
+      shouldUpdate = (old !== cur) && ((elm as any)[key] !== cur);
+
+    // DOM INPUT.checked is a boolean property. The element's internal value should be
+    // used for comparison. fixes #950
+    else if (key === 'checked')
+      shouldUpdate = (elm as any)[key] !== cur;
```
jvanbruegge, line 34: Could you just use `(elm as any).checked` here? It's clearer that way

jvanbruegge (changes requested):
Looks good, but you need to run prettier (the git commits are fixed in the latest master even for older npm version, otherwise just run `npm run format`).
The commit message I can fix while merging

jvanbruegge:
Also, do we have a test for this? I would be ok with merging it without one, just asking

mreinstein:
> do we have a test for this? 

That's a good question. I poked around at creating a test for this but it wasn't clear to me how we might automate it.

The problem is the test case the original submitter provided requires clicking a checkbox, and I didn't see any ui automation for doing these kinds of interactions programmatically. 


> I would be ok with merging it without one, just asking

Before you merge anything, please read the last comment in #950  I'm curious about if this is a larger class of problems involving any props that change their state due to user interaction.

paldepind:
I think it would be nicer to have `shouldUpdate` as a `const` and setting it in one place. Keeping the comments and using the ternary operator should make things just as readable.

mreinstein:
> Keeping the comments and using the ternary operator should make things just as readable.

But there are 3 cases here. Are you suggesting using a nested ternary? Because nesting those makes them hard to read. :(

paldepind:
Yes, exactly. I'm suggesting a nested ternary. I don't think they are hard to read (but that of course depends on what one is used to seeing). Prettier's formatting of them isn't my favourite but it's pretty good.

```js
const shouldUpdate =
  key === "value"
    ? old !== cur && (elm as any)[key] !== cur
    : key === "checked"
    ? (elm as any)[key] !== cur
    : old !== cur;
```

Unnecessary `let`s are in my opinion much more confusing 😅 (though admittedly we do have a lot of those in the codebase as it predates both `const` and block scoping).

mreinstein:
I personally find a prettier, typescript, and this style of writing perl-ish looking code difficult to read, but maybe that's just my age showing. 🤷 

Anyway, we should probably avoid a syntax war, it's not as important as the underlying question of "does this need to be handled for other property types?" and I don't want to distract from that. If this is how we want to write the code in snabbdom, I'm all for it. 👍

mreinstein:
Hey, I'm sorry that I dropped off the globe.  I'd like to re-visit this issue and see if we can the fix merged in. I have time I can dedicate to this.  @paldepind @jvanbruegge do you recall where we left off with this?  It sounds like we agreed on strategy but got hung up on specific syntax?

mreinstein:
> do we have a test for this?

@jvanbruegge  worked on a proof-of-concept way to automate the test for this:  https://github.com/snabbdom/snabbdom/issues/1079

iambumblehead:
possibly related https://github.com/snabbdom/snabbdom/pull/418

iambumblehead:
@mreinstein maybe the test here can be copy-pasted into this PR? https://github.com/snabbdom/snabbdom/pull/418/files

mreinstein:
Possibly, but I'm waiting for feedback from Jan.  There's already a lot of stuff flying around today. I don't want to jam more activity in until he's had a chance to catch up.

paldepind:
I've left a comment over on the original issue #950.

PR #964, opened 2021-05-18
fix(npm script): karma config path

heng1025:


jvanbruegge:
The benchmark is not ported to the latest version yet, I am currently working on that. To run the unit tests instead use `npm test`.

heng1025:
> The benchmark is not ported to the latest version yet, I am currently working on that. To run the unit tests instead use `npm test`.

Got it.

PR #967, opened 2021-05-26
docs: add additional JSX usage information

geotrev:
Motivation: I recently tried using the JSX pragma in a project and it wasn't clear why Snabbdom wasn't processing my element props via the appropriate module (class, style, props, attributes, etc). But I eventually figured it out, and figured it'd be worth adding here.

Main additions are:
- adds JSX examples to module sections
- removes file extension requirement in babel config section (babel only cares about the pragma, not the extension, to my knowledge).

Is it worth adding these? Should there be other clarifications/additions I didn't cover? Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,7 +395,7 @@ not the class should stay or go on the vnode.
 h("a", { class: { active: true, selected: false } }, "Toggle");
 ```
 
-In JSX, you can use the `class` key.
+In JSX, you can use the `class` prop.
 
 ```jsx
 <div class={{ "foo": true, "bar": true }} />
@@ -410,7 +410,7 @@ Allows you to set properties on DOM elements.
 h("a", { props: { href: "/foo" } }, "Go to Foo");
 ```
 
-In JSX, you can use the `props` key.
+In JSX, you can use the `props` prop.
 
 ```jsx
 <input props={{ name: "foo" }} />
@@ -433,7 +433,7 @@ Same as props, but set attributes instead of properties on DOM elements.
 h("a", { attrs: { href: "/foo" } }, "Go to Foo");
 ```
 
-In JSX, you can use the `attrs` key.
+In JSX, you can use the `attrs` prop.
 
 ```jsx
 <div attrs={{ 'aria-label': "I'm a div" }} />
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -395,6 +395,13 @@ not the class should stay or go on the vnode.
 h("a", { class: { active: true, selected: false } }, "Toggle");
 ```
 
+In JSX, you can use the `class` prop.
```
paldepind, line 398: I think "attribute" is more appropriate than "prop" in a JSX context?

paldepind:
Thanks a lot for the PR @geotrev. If you found this to be unclear then I'm sure other people will as well. I think it's really good, in general, to improve documentation base on "real world experience" like this.

I've left a single comment but otherwise it looks good to me :)

README.md
geotrev: I've been thinking about this since you commented yesterday and I'm not sure! From data model perspective, the "key" (for lack of a better term) is neither a property nor an attribute, but can produce either of those as side-effects.

My original verbiage was "key"; does that make more sense? Or should it be labeled in relation to its side-effect? E.g., "attribute" for `class` and `style`, "prop" for `dataset`, and "handler" for `on`. 

Thanks for the review!

paldepind: I was thinking "attribute" because in JSX something like `<div foo=bar>` looks like HTML/XML and in those languages "foo" would be called an attribute. But, I see your point, those "attributes" may not actually correspond to the "real" "attributes" on the actual HTML/DOM.

In the [JSX specification](https://facebook.github.io/jsx/) it seems that these things are called attributes, regardless of what they eventually end up meaning to whoever is using JSX.

But this doesn't really matter, I think people will get the point anyway 😅. What about sidestepping the issue an writing something like

> In JSX, you can use `class` like this:

?. Alternatively, I think "key" would also be fine. But the abbreviated "prop" sounds weird to me.

geotrev: Updated using your suggested verbiage:

> In JSX, you can use <module> like this:

I also added to the event listeners section, forgot to do that before.

geotrev: My commit message wasn’t helpful on that last commit, but I suppose you will probably squash commits when merging for simplicity. :)

jvanbruegge:
Thank you! Merged it now

README.md
paldepind: Looks good. Thanks for the PR 😀

Issue #560, opened 2020-02-29
support array of elements without needing a wrapper div


jimfranke:
Does Snabbdom has support for JSX fragments to return an array of elements? If not, is this on the roadmap, or what does it take to implement it? Thanks.

mightyiam:
@jimfranke thank you for your interest. Could you please describe an API for this?

jimfranke:
Basicly it comes down to a component being able to return an array of elements which are then flattened into the parent component, without the need for a wrapping div. The diff algorithm needs to support array based diffs for this. See https://reactjs.org/docs/fragments.html for an example.

Leland-Kwong:
This would be nice. I recently setup my own implementation by doing a patch on the fragment's parent vnode and replacing the old children with the new ones. Its not the most ideal, but it works reasonably well.

jimfranke:
@Leland-Kwong Can you share your implementation and does it have any downsides patching the parent?

Leland-Kwong:
@jimfranke Turns out theres a much simpler solution. The way I've done it is to auto-flatten the children in my vnode factory function. As an example:

```js
const vnodeFactory = (selector, props, children) => {
  return {
    sel: selector,
    children: children.flat(Infinity)
  }
}

const vtree = vnodeFactory('div', [1, [1,2,3]]);
```

The only caveat is this doesn't work for the root vnode. I'd imagine you can also have it work for the root vnode by creating based off of the initial dom node that you're patching to.

nojvek:
If you use https://github.com/nojvek/snabbdom-jsx-lite

Fragment support comes out of the box. I really would have loved to not create a new npm module but land snabbdom jsx support here but @paldepind has been AWOL for so long, it was better to create a new module.

https://github.com/nojvek/snabbdom-jsx-lite#jsx-fragments

mreinstein:
@jimfranke I renamed the issue in an attempt to make it more accurate in terms of what is being asked for.

If I've got this wrong I'm happy to change it back though. Definitely not looking to re-write history here.

ryota-ka:
Hi @jimfranke, I've submitted #974 that introduces a support for JSX fragments.
Hope it'll be merged and meet your needs.
Feedback welcomed!

PR #974, opened 2021-07-01
Support JSX fragments

ryota-ka:
Addresses #560

Added support for JSX fragments.
A fragment can be created with `<>...</>`, and it will be converted into a VNode without `sel`.

Please check the test cases to see how it works.
(In fact they are almost copy-and-paste from ones for `jsx` function, but looks reasonable for me.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -780,6 +780,9 @@ significant computational time to generate.
 
 ## JSX
 
+Note that JSX fragments are still experimental and must be opted in.
+See [`fragment`](#fragment-experimental) section for details.
+
 ### TypeScript
 
 Add the following options to your `tsconfig.json`:
@@ -788,20 +791,28 @@ Add the following options to your `tsconfig.json`:
 {
   "compilerOptions": {
     "jsx": "react",
-    "jsxFactory": "jsx"
+    "jsxFactory": "jsx",
+    "jsxFragmentFactory": "Fragment"
   }
 }
 ```
 
-Then make sure that you use the `.tsx` file extension and import the `jsx` function at the top of the file:
+Then make sure that you use the `.tsx` file extension and import the `jsx` function and the `Fragment` function at the top of the file:
 
 ```tsx
-import { jsx, VNode } from "snabbdom";
+import { Fragment, jsx, VNode } from "snabbdom";
 
 const node: VNode = (
   <div>
     <span>I was created with JSX</span>
   </div>
+);
+
+const fragment: VNode = (
+  <>
+    <span>JSX fragments</span>
+    are experimentally supported
+  </>
 );
 ```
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -398,7 +398,7 @@ h("a", { class: { active: true, selected: false } }, "Toggle");
 In JSX, you can use `class` like this:
 
 ```jsx
-<div class={{ "foo": true, "bar": true }} />
+<div class={{ foo: true, bar: true }} />
```
ryota-ka, line 401: These kind of changes has been introduced by Prettier kicked by `lint-staged` when I try to make a commit.
Is it okay to commit them? 🤔

jvanbruegge: Yeah, that is fine

src/jsx.ts
```diff
@@ -25,6 +25,30 @@ export type FunctionComponent = (
   children?: VNode[]
 ) => VNode;
 
+export function Fragment(
```
jvanbruegge, line 28: Any special reason why this is not lower case?

test/unit/jsx.tsx
```diff
@@ -272,4 +272,275 @@ describe("snabbdom", function () {
       });
     });
   });
+
+  describe("Fragment", function () {
```
jvanbruegge, line 276: The unit tests are good, but can you please add one or two browser tests to verify that the (flattened) children are also patched correctly?

jvanbruegge (changes requested):
Looks good, just needs some browser tests

src/jsx.ts
ryota-ka: I believe it's the convention in the community.

Please refer these documentations:
- https://www.typescriptlang.org/tsconfig#jsxFragmentFactory
- https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#pragmafrag

jvanbruegge: I see, thanks

test/unit/jsx.tsx
ryota-ka: Hi @jvanbruegge,
While working on writing the test case, I noticed that a VNode like `{ sel: undefined, children: [...] }` is not a valid one while it typechecks.

I've submitted #975 that will treat such a VNode as a virtual `DocumentFragment`.
So could you please review that PR first?
Thanks!

ryota-ka: @jvanbruegge I've added a test case that ensures that virtual nodes made by JSX fragments can correctly be patched.
Please check it out!
https://github.com/snabbdom/snabbdom/pull/974/files#diff-2a05a44523d99a42362a290d489f088b0d55f4e74be980bb43ea5d1bc055e9b2R546-R598

ryota-ka:
I've refined the implementation based on #975 and it should now be ready for review again!

jvanbruegge (approved):
Looks good now!

jvanbruegge:
Published this as `v3.3.0`

PR #975, opened 2021-07-14
Support DocumentFragment

ryota-ka:
# Synopsis

This pull request introduces an ability to handle [`DocumentFragment`](https://developer.mozilla.org/docs/Web/API/DocumentFragment), which is another notable subclass of [`Node`](https://developer.mozilla.org/en-US/docs/Web/API/Node) other than [`Element`](https://developer.mozilla.org/docs/Web/API/Element) nor [`Text`](https://developer.mozilla.org/docs/Web/API/Text).
It allows a function to return multiple VNodes without an extra wrapper element, resolving #560.

`fragment` function is now exported from the entry point.
There is, however, a room for discussion on the design of the API.

The second commit will allow us to use a `DocumentFragment` as a container, in addition to an `Element`.

# Possible design choices

## a) Public API

In the current implementation I chose a2), but of course it's okay to change it along with the discussion before merging.

### a1) Extend `h` function so that it can create a fragment

Looking around the documents and test cases, there is no doubt that the most popular way to produce a VNode is to call `h` function.
This is why adding such an ability to `h` function can be a good choice.

There are still some possible APIs:

- a11) `h(children)`
  - [`react-hyperscript`](https://github.com/mlmorg/react-hyperscript) uses this.
- a12) `h(undefined, children)`
- a13) `h('', children)`

#### Pros

- No extra import is needed other than `h`

#### Cons

- AFAIK this doesn't exist in the original hyperscript notation
  - **UPDATE:** I've noticed that we already have a syntactic extension to the original hyperscript: `h('!', ...)`

### a2) Provide a function other than `h`

As the original hyperscript notation does not consider fragments, it may be undesirable to change the API of `h` function.
We may want to export different function other than `h` for this reason.

In the current implementation, I named it `fragment`, but to be honest I'm not sure what is the good name for this function.

If we also want to support JSX fragments (which I'm working on in #974), `Fragment` (with a capital "F") should be avoided, as it is a conventional name for it.

#### Pros

- Does not break the original API of `h` function

#### Cons

- It must be named
- Consumers have to add an extra import for it

### a3) Do nothing

Of course VNodes can also be created with `vnode` function, `vnode(undefined, {}, children, undefined, undefined)` will just do.

#### Pros

- No new API have to be introduced
  - That is, we can defer the decision. Maybe we can merge this time and continue to discuss what the desirable API look like later on.

#### Cons

- It may look like a "hidden" API
- Children must be normalized manually before passing them to `vnode`

## b) Internal representation

It is relatively minor topic compared to what the public API should look like, but we also have to decide how a VNode representing a `DocumentFragment` is encoded.

In the current implementation I chose b1), but there are no profound reasons for this choice.

Please note that both b1) and b2) are assignable to the current implementation of `VNode` type.

### b1) `{ sel: undefined, children: [...] }`

Currently `VNode` type is defined like this:
https://github.com/snabbdom/snabbdom/blob/27e9c4d5dca62b6dabf9ac23efb95f1b6045b2df/src/vnode.ts#L12-L19

In spite of its assignability to `VNode` type, `{ sel: undefined, children: [...] }` is not a first-class citizen of Snabbdom, and will cause a runtime error.
This option will give a correct semantics to such an empty seat.

### b2) `{ sel: '', children: [...] }`

I saw @staltz passing an empty string to the first argument of `vnode` function in #320, perhaps meaning "whatever".
As this line is still alive in the code base, it may be more conventional.

https://github.com/snabbdom/snabbdom/blob/27e9c4d5dca62b6dabf9ac23efb95f1b6045b2df/src/tovnode.ts#L35

# Breaking changes

This is technically breaking, as `DOMAPI` now requires `createDocumentFragment` and `isDocumentFragment` method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ patch(vnode, newVnode); // Snabbdom efficiently updates the old view to the new 
   - [`patch`](#patch)
     - [Unmounting](#unmounting)
   - [`h`](#h)
+  - [`fragment`](#fragment-experimental) (experimental)
   - [`tovnode`](#tovnode)
   - [Hooks](#hooks)
     - [Overview](#overview)
@@ -229,6 +230,27 @@ const vnode = h("div", { style: { color: "#000" } }, [
   h("h1", "Headline"),
   h("p", "A paragraph"),
 ]);
+```
+
+### `fragment` (experimental)
+
+Caution: This feature is currently experimental and must be opted in.
+Its API may be changed without an major version bump.
+
+```mjs
+const patch = init(modules, undefined, {
+  experimental: {
+    fragments: true,
+  },
+});
+```
+
+Creates a virtual node that will be converted to a document fragment containing the given children.
+
+```mjs
+import { fragment, h } from "snabbdom";
+
+const vnode = fragment(["I am", h("span", [" a", " fragment"])]);
 ```
 
 ### `tovnode`
@@ -398,7 +420,7 @@ h("a", { class: { active: true, selected: false } }, "Toggle");
 In JSX, you can use `class` like this:
 
 ```jsx
-<div class={{ "foo": true, "bar": true }} />
+<div class={{ foo: true, bar: true }} />
 // Renders as: <div class="foo bar"></div>
 ```
 
@@ -436,7 +458,7 @@ h("a", { attrs: { href: "/foo" } }, "Go to Foo");
 In JSX, you can use `attrs` like this:
 
 ```jsx
-<div attrs={{ 'aria-label': "I'm a div" }} />
+<div attrs={{ "aria-label": "I'm a div" }} />
 // Renders as: <div aria-label="I'm a div"></div>
 ```
 
@@ -492,11 +514,13 @@ h(
 In JSX, you can use `style` like this:
 
 ```jsx
-<div style={{
-  border: "1px solid #bada55",
-  color: "#c0ffee",
-  fontWeight: "bold",
-}} />
+<div
+  style={{
+    border: "1px solid #bada55",
+    color: "#c0ffee",
+    fontWeight: "bold",
+  }}
+/>
 // Renders as: <div style="border: 1px solid #bada55; color: #c0ffee; font-weight: bold"></div>
 ```
 
--- a/src/h.ts
+++ b/src/h.ts
@@ -86,3 +86,28 @@ export function h(sel: any, b?: any, c?: any): VNode {
   }
   return vnode(sel, data, children, text, undefined);
 }
+
+/**
+ * @experimental
+ */
+export function fragment(children: VNodeChildren): VNode {
+  let c: any;
+  let text: any;
+
+  if (is.array(children)) {
+    c = children;
+  } else if (is.primitive(c)) {
+    text = children;
+  } else if (c && c.sel) {
+    c = [children];
+  }
+
+  if (c !== undefined) {
+    for (let i = 0; i < c.length; ++i) {
+      if (is.primitive(c[i]))
+        c[i] = vnode(undefined, undefined, undefined, c[i], undefined);
+    }
+  }
+
+  return vnode(undefined, {}, c, text, undefined);
+}
--- a/src/htmldomapi.ts
+++ b/src/htmldomapi.ts
@@ -30,6 +30,11 @@ export interface DOMAPI {
   isElement: (node: Node) => node is Element;
   isText: (node: Node) => node is Text;
   isComment: (node: Node) => node is Comment;
+  /**
+   * @experimental
+   * @todo Make it required when the fragment is considered stable.
+   */
+  isDocumentFragment?: (node: Node) => node is DocumentFragment;
 }
 
 function createElement(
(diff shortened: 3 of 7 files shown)


ryota-ka:
@mreinstein @mightyiam @jvanbruegge I would appreciate it if you could give some feedbacks on the API. Thanks!

mreinstein:
I think the proposal is well written, thank your for taking the time to think about and write it!

I don't understand how this works when document fragments are nested and changed. Doesn't this break the current diffing algorithm?

I think I'd need to see a fairly comprehensive set of tests to have some confidence that this works for nesting.

It seems like an ambitious fix for #560 

That's my initial thoughts anyway, I don't speak for others. @jvanbruegge @paldepind what are your thoughts?

ryota-ka:
@mreinstein As we've gotten no responses from other maintainers after 4 months, I'd like to make it clear what is the next step and what is have to be done in order to ship this feature.

I love Snabbdom's zen of minimalism, and understand it's hard to decide to add a new stuff to the core, but I also often feel awkward that a virtual DOM library cannot handle the Web API itself well.
Without this, we will continue sufferring from a bunch of extra `<div>` elements that breaks CSS styles (e.g. `:first-child`), or branching against `VNode | VNode[]`.

---

By the way, I'm also not very sure if the diffing algorithm works well for complexly nested fragments, and this indeed makes it harder to merge this into the core.
There are, however, several plans to mitigate the pain.

One idea is to ship it with marking it as an "experimental" feature, without a guarantee of major version bumps on breaking changes.
This make is it much easier for us to add a breaking changes to the API, or even to remove it in the future if really needed.

The second one is, it is not orthogonal to the first one though, to make it completely opt-in.
The point here is that this PR only introduces a single change to the core: yet another `else-if` branch shown below.

https://github.com/snabbdom/snabbdom/blob/d1965fb85493061c6044b2b1239fe2e9b52daca6/src/init.ts#L165-L177

And the proposed API here is something like this:

```typescript
// Currently the `init` function takes the optional DOM API implementation
// so we might also have to put it into the options object, but anyway:
const patch = init(modules, {
  experimentalFragments: true, // default: false
})
```

By changing the condition to `else if (options.experimentalFragments && vnode.children)`, we can ship it as an opt-in feature without any breakage.
When we consider it to be battle-tested enough and no longer unstable someday in the future, then we can simply turn the default value to `true`.

Done is better than perfect, and we even can get feedbacks from the early adopters in the community, which will help us incrementally improve the library.
In fact I'm very eager to test it on our production environment.
(We @herp-inc heavily uses Cycle.js w/ Snabbdom for our software development.)

What do you think?

jvanbruegge:
Sorry for not being active for so long, I had several time consuming projects. I support the idea of explicitly marking this as experimental, with an explicit opt-in. I will take a look over the code later this week and then go ahead and merge it, given no objections from @mreinstein

ryota-ka:
Thank you @jvanbruegge!
I understand your situation where you're working very hard on some tough projects (such as research on the row polymorphism).
I really appreciate your contribution.

mreinstein:
Sorry for the reply delay.  I also support pushing this out as an experimental opt-in feature. PR looks ok to me I think.

mreinstein:
My only gripe with this PR is there's no updates to the documentation explaining this capability or how to use it.

Other than that it looks good.

ryota-ka:
Thank you again @mreinstein!

So my next tasks are:
- Make it experimental
- Add some docs about `fragment` function

I hope I can accomplish them by the next weekend, but there might be a delay.

ryota-ka:
Hi @mreinstein and @jvanbruegge. Today I've added 3 commits.

The first one prevents `DOMAPI` from facing a breaking change.
It's achieved by making `.createDocumentFragment()` and `.isDocumentFragment()` optional.
Due to this change, this pull requests introduces no breaking changes, so a minor version bump should be sufficient.
Instead, you'll face a runtime error when you try to play with fragments on an unsupported platform.

The second one makes this feature experimental and opt-in.
The `init` function now takes the third argument: `options`.
I suppose, however, the second argument `domapi` should be put into it in the next major version bump

The last one adds a documentation for `fragment` function.

Please ping me if it looks good. I'll squash the "fixup!" commits.

jvanbruegge:
The code looks good for merge, just needs a rebase to fix the merge conflict

ryota-ka:
@jvanbruegge @mreinstein I've rebased the branch to resolve a merge conflict and squash the commits, and then force-pushed it.
The pull request should be ready for the review again!

jvanbruegge:
I've published 3.2.0 now

mreinstein:
Great work @ryota-ka thanks for seeing this through!

ryota-ka:
Thanks again @jvanbruegge and @mreinstein!
I really appreciate your dedication.

Then I'll work on #974 next.

ryota-ka:
While I was trying to adopt this in Cycle.js, I found that it is impossible to pass the option to turn an experimental feature on to the `init` function, as the `makeDOMDriver` function does not accept such an argument at this moment.

So I'm thinking of doing these of two changes:

- Export `Options` type from the entrypoint of `snabbdom`
  - Making it public, there is a room for discussion of its name (`Options`, `InitOptions`, or some other cool name?)
- Update `makeDOMDriver` function to accept the options for the Snabbdom's `init` function.

If this looks okay, I'll submit pull requests.

cc: @staltz

PR #979, opened 2021-08-17
allow javascript String obj to be used as children

### LINKED ISSUE #977: h doesn't handle String Objects
h('div', new String('y')) doesn't render, my guess is because the primitive check only does typeof x == 'string'

this could be fixed by doing in the primitive check

if (typeof x === 'string' || x instanceof String)

tobymao:
fixes #977


kuraga:
:+1:

PR #981, opened 2021-08-27
fix lint

tobymao:


jvanbruegge:
There was something weird with GH actions, it only checked out the master branch for some reason. I fixed the lint and some other issues myself now

PR #983, opened 2021-09-03
chore(test): fix tests for string object children

jvanbruegge:


jvanbruegge:
For some reason, github checks out the master branch instead of this branch. Tests do pass

PR #994, opened 2021-12-30
fix: resolve 'VNode.data' may be undefined

### LINKED ISSUE #993: VNode.data can be undefined and should be used normally
```
  const patched = patch(document.getElementById("container"), [
    vnode("h1", undefined, undefined, "hello", undefined),
    vnode("h1", undefined, undefined, "snabbdom", undefined),
  ]);
  console.log(patched);
```
image

dmoosocool:
resolve #993


jvanbruegge:
Thanks for the PR. I think it might be easier to tests once in `patch` if the `data` is undefined and assign an empty object in this case. All those `?.` operators will introduce a `if` which could be avoided that way

PR #995, opened 2021-12-30
Fix issue 976

### LINKED ISSUE #976: toVNode parses datasets as attributes
This feels misleading:

```javascript
const element = document.createElement("div");
element.dataset.example = "test";
const tree = toVNode(element);

// {"sel":"div","data":{"attrs":{"data-example":"test"}},"children":[],"elm":{}}
```

I would have expected to see:

```javascript
// {"sel":"div","data":{"dataset":{"example":"test"}},"children":[],"elm":{}}
```

VNodes generated by `toVNodes` can only have datasets added to them by parsing datasets out of attributes while traversing the tree, or else they're lost during the patch:

```javascript
addDataset(tree) {
    if (Object.keys(tree.data.attrs).length > 0) {
        Object.keys(tree.data.attrs).forEach((key: string) => {
            if (key.includes("data-")) {
                if ("dataset" in tree.data) {
                    tree.data["dataset"][key.replace("data-", "")] =
                        tree.data.attrs[key];
                } else {
                    tree.data["dataset"] = {};
                    tree.data["dataset"][key.replace("data-", "")] =
                        tree.data.attrs[key];
                }
            }
        });
    }
    tree.children.forEach((child) => {
        if (child.data && "attrs" in child.data) {
            this.addDataset(child);
        }
    });
}
```

This feels like unecessary surface area for bugs.

I wanted to check if this is intended behaviour, as I'm aware that support for datasets comes from a separate module. If not, I would be happy to raise a PR.

dmoosocool:
fix #976

### DIFF
--- a/src/tovnode.ts
+++ b/src/tovnode.ts
@@ -11,6 +11,9 @@ export function toVNode(node: Node, domApi?: DOMAPI): VNode {
     const c = cn ? "." + cn.split(" ").join(".") : "";
     const sel = api.tagName(node).toLowerCase() + id + c;
     const attrs: any = {};
+    const datasets: Record<string, string> = {};
+    const data: Record<string, any> = {};
+
     const children: VNode[] = [];
     let name: string;
     let i: number, n: number;
@@ -18,14 +21,19 @@ export function toVNode(node: Node, domApi?: DOMAPI): VNode {
     const elmChildren = node.childNodes;
     for (i = 0, n = elmAttrs.length; i < n; i++) {
       name = elmAttrs[i].nodeName;
-      if (name !== "id" && name !== "class") {
+      if (name.startsWith("data-")) {
+        datasets[name.slice(5)] = elmAttrs[i].nodeValue || "";
+      } else if (name !== "id" && name !== "class") {
         attrs[name] = elmAttrs[i].nodeValue;
       }
     }
     for (i = 0, n = elmChildren.length; i < n; i++) {
       children.push(toVNode(elmChildren[i], domApi));
     }
-    const data = { attrs };
+
+    if (Object.keys(attrs).length > 0) data.attrs = attrs;
+    if (Object.keys(datasets).length > 0) data.datasets = datasets;
+
     if (
       sel[0] === "s" &&
       sel[1] === "v" &&
--- a/test/unit/core.ts
+++ b/test/unit/core.ts
@@ -549,8 +549,32 @@ describe("snabbdom", function () {
         assert.deepEqual(vnode.data, { attrs: { data: "value" } });
         const children = vnode.children as [VNode, VNode];
         assert.strictEqual(children[0].sel, "x-h2#hx");
-        assert.deepEqual(children[0].data, { attrs: { "data-env": "xyz" } });
+        assert.deepEqual(children[0].data, { datasets: { env: "xyz" } });
         assert.strictEqual(children[1].text, "Foobar");
+      });
+
+      it("can parsing datasets and attrs", function () {
+        const onlyAttrs = document.createElement("div");
+        onlyAttrs.setAttribute("foo", "bar");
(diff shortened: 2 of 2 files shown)


src/tovnode.ts
```diff
@@ -15,6 +24,21 @@ export function toVNode(node: Node, domApi?: DOMAPI): VNode {
...
+      if (attr.name.startsWith("data-")) {
+        datasets[attr.name.replace("data-", "")] = attr.value;
+      }
+    });
+
+    // remove attributes from datasets
+    getAttrs(elmAttrs).forEach((attr) => {
+      Object.keys(datasets).includes(attr.name.replace("data-", "")) &&
+        elmAttrs.removeNamedItem(attr.name);
+    });
```
jvanbruegge, line 40: Creating a new array twice, just to use the `forEach` method is not really necessary. Please refactor this to use a single `for` loop. Also, instead of `replace` please use `slice`, we are only interested in removing stuff from the beginning of the string anyways

src/tovnode.ts
```diff
@@ -1,6 +1,15 @@
 import { vnode, VNode } from "./vnode";
 import { htmlDomApi, DOMAPI } from "./htmldomapi";
 
+// get real array from NamedNodeMap
+function getAttrs(map: NamedNodeMap) {
```
jvanbruegge, line 5: This function should not exists (see other comment)

jvanbruegge (changes requested):
Thanks for the PR, but to merge it please refactor the code to not unnecessarily allocate arrays.

dmoosocool:
> Thanks for the PR, but to merge it please refactor the code to not unnecessarily allocate arrays.

Thanks for taking the time to help with the code review, you are right. i am learned a lot form it.

src/tovnode.ts
```diff
@@ -11,21 +11,29 @@ export function toVNode(node: Node, domApi?: DOMAPI): VNode {
...
         attrs[name] = elmAttrs[i].nodeValue;
       }
     }
     for (i = 0, n = elmChildren.length; i < n; i++) {
       children.push(toVNode(elmChildren[i], domApi));
     }
-    const data = { attrs };
+
+    if (Object.keys(attrs).length) data.attrs = attrs;
+    if (Object.keys(datasets).length) data.datasets = datasets;
```
jvanbruegge, line 35: Please no implicit number <-> boolean conversion
```suggestion
    if (Object.keys(datasets).length > 0) data.datasets = datasets;
```

test/unit/core.ts
```diff
@@ -539,18 +539,30 @@ describe("snabbdom", function () {
         h2.setAttribute("data-env", "xyz");
         const text = document.createTextNode("Foobar");
         const elm = document.createElement("div");
+        const elm2 = document.createElement("div");
         elm.id = "id";
         elm.className = "class other";
         elm.setAttribute("data", "value");
         elm.appendChild(h2);
         elm.appendChild(text);
+        elm.appendChild(elm2);
```
jvanbruegge, line 548: Please move this to a new test case.

jvanbruegge (commented):
A lot better now, just two minor things left.

dmoosocool:
I found new problem, parser datasets is needs to be consistent with the DOM API?

eg:  
```js
element.setAttribute('data-my-name', 'cool');  // { datasets: { myName: 'cool' } }
```

If don't need to deal with it: 
```js
element.setAttribute('data-my-name', 'cool'); // { datasets: { 'my-name':  'cool' } }
```

jvanbruegge:
Sorry it took me so long to get back to you. No I would not dual the DOM Api, this is a rather confusing behavior. The way it works at the moment is good. I've removed the merge commit you have on this branch and rebased it. I could not push to your branch, so I created a new PR for that: #1008
The commit in the PR is still yours, thanks for the PR :)

Issue #388, opened 2018-08-07
toVNode doesn't parse SVGs correctly


michael-england:
`toVNode` doesn't account for namespace when parsing an SVG.  The resulting vnode is missing `data.ns` causing it not to get patched correctly, resulting with the SVG not rendering.

Here's a fiddle demonstrating this behavior: http://jsfiddle.net/michael_england/07pqtbfu/47/

Uncommenting the lines to manually add `data.ns` fixes this, but manually adding them shouldn't be required.  Seems like any fix should account for the SVG not being a root node.

mreinstein:
I'm running into this problem when I attempt to inline svg into my app with snabbdom. manually updating the `vnode.data.ns = 'http://www.w3.org/2000/svg'` value in my code doesn't seem to solve the problem. Wondering if the code changed such that this workaround in the fiddle is no longer effective.

Help please? :(

macarc:
This needs to be added to line 26 of `tovnode.ts`
```ts
const data = { attrs };
if (
    sel[0] === 's' && sel[1] === 'v' && sel[2] === 'g' &&
    (sel.length === 3 || sel[3] === '.' || sel[3] === '#')
  ) {
    addNS(data, children, sel)
  }
    return vnode(sel, data, children, undefined, node)
```

(Copied from the end of `h`)
I think that fixes it, but I can't check right now.

macarc:
That does indeed fix it. @mightyiam OK for me to go ahead and submit a PR with this?

toSayNothing:
@mightyiam hello, do you have a plan to fix it? or can I submit a PR with @macarc 's code?😅

> This needs to be added to line 26 of `tovnode.ts`
> 
> ```ts
> const data = { attrs };
> if (
>     sel[0] === 's' && sel[1] === 'v' && sel[2] === 'g' &&
>     (sel.length === 3 || sel[3] === '.' || sel[3] === '#')
>   ) {
>     addNS(data, children, sel)
>   }
>     return vnode(sel, data, children, undefined, node)
> ```
> 
> (Copied from the end of `h`) I think that fixes it, but I can't check right now.

macarc:
Ha, I forgot about this entirely. I'll make a PR shortly.

PR #996, opened 2022-01-09
Fix SVG bugs

### LINKED ISSUE #867: thunk doesn't seem to work with svg child elements
My render functions looks like this:

```javascript

function renderSvg (lines) {
    return h(
        'svg',
        {
            style: {
                height: '200px',
                width: '200px',
                'background-color': 'white'
            },
            attrs: {
                xmlns: 'http://www.w3.org/2000/svg',
                viewBox: '0 0 200 200'
            }
        },
        renderLineGroup(lines)
    )
}


function renderLineGroup (lines) {
    const strokeWidth = 1
    const dashArray = '4 2'

    return h(
        'g',
        {},
        lines.map((l, idx) => thunk('line', `line_${idx}`, renderLine, [ l.x1, l.x2, l.y1, l.y2, strokeWidth, dashArray ]))
        //lines.map((l, idx) => renderLine(l.x1, l.x2, l.y1, l.y2, strokeWidth, dashArray))
    )
}


function renderLine (x1, x2, y1, y2, strokeWidth, dashArray) {
    return h('line', {
        style: {
            'stroke-width': strokeWidth,
            stroke: 'red'
        },
        attrs: { x1, x2, y1, y2, 'stroke-dasharray': dashArray }
    })
}

```

This constructs the svg tag as expected, but it doesn't actually draw. So strange! Here's what it looks like when rendering via `thunk`:
thunk

when I comment out the `lines.map` line and uncomment the one that doesn't use thunk, I see what I'd expect:

non-thunk

macarc:
Fixes #867 and #388

### DIFF
--- a/src/thunk.ts
+++ b/src/thunk.ts
@@ -16,7 +16,7 @@ export interface ThunkFn {
 }
 
 function copyToThunk(vnode: VNode, thunk: VNode): void {
-  const ns = thunk.data ? thunk.data.ns : null;
+  const ns = thunk.data?.ns;
   (vnode.data as VNodeData).fn = (thunk.data as VNodeData).fn;
   (vnode.data as VNodeData).args = (thunk.data as VNodeData).args;
   thunk.data = vnode.data;
--- a/src/thunk.ts
+++ b/src/thunk.ts
@@ -1,5 +1,5 @@
 import { VNode, VNodeData } from "./vnode";
-import { h } from "./h";
+import { h, addNS } from "./h";
 
 export interface ThunkData extends VNodeData {
   fn: () => VNode;
@@ -16,12 +16,14 @@ export interface ThunkFn {
 }
 
 function copyToThunk(vnode: VNode, thunk: VNode): void {
+  const ns = thunk.data ? thunk.data.ns : null;
   (vnode.data as VNodeData).fn = (thunk.data as VNodeData).fn;
   (vnode.data as VNodeData).args = (thunk.data as VNodeData).args;
   thunk.data = vnode.data;
   thunk.children = vnode.children;
   thunk.text = vnode.text;
   thunk.elm = vnode.elm;
+  if (ns) addNS(thunk.data, thunk.children, thunk.sel);
 }
 
 function init(thunk: VNode): void {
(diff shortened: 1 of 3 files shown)


macarc:
I could also move `addNS` to its own file, since it's not really specific to `h.ts` anymore. Let me know.

src/thunk.ts
```diff
@@ -16,12 +16,14 @@ export interface ThunkFn {
 }
 
 function copyToThunk(vnode: VNode, thunk: VNode): void {
+  const ns = thunk.data ? thunk.data.ns : null;
```
jvanbruegge, line 19:
```suggestion
  const ns = thunk.data?.ns;
```

src/tovnode.ts
```diff
@@ -24,7 +25,16 @@ export function toVNode(node: Node, domApi?: DOMAPI): VNode {
     for (i = 0, n = elmChildren.length; i < n; i++) {
       children.push(toVNode(elmChildren[i], domApi));
     }
-    return vnode(sel, { attrs }, children, undefined, node);
+    const data = { attrs };
+    if (
+      sel[0] === "s" &&
+      sel[1] === "v" &&
+      sel[2] === "g" &&
+      (sel.length === 3 || sel[3] === "." || sel[3] === "#")
```
jvanbruegge, line 33:
```suggestion
      sel === "svg" ||
      sel.startsWith("svg.") ||
      sel.startsWith("svg#")
```
A bit easier to read

jvanbruegge (commented):
Thanks for the PR!
Code looks good, just two minor style comments.

src/tovnode.ts
macarc: Do you want me to change that here too? https://github.com/snabbdom/snabbdom/blob/9de2307dd9bec6106f587a18eb7b0c994e24722b/src/h.ts#L79-L84

jvanbruegge: yes, please

macarc: Cool, do you want me to create a new commit or to squash?

jvanbruegge: Thinking a bit more about this, this version is a bit more performant. I applied my other suggestion and will merge it directly now

jvanbruegge:
Released as `v3.3.1`

PR #1001, opened 2022-01-18
fix #1000

### LINKED ISSUE #1000: A question for style module
Hi，I see the following description on the readme:
> Note that the style module does not remove style attributes if they
are removed as properties from the style object. To remove a style,
you should instead set it to the empty string.   

and I tried to remove the properties from the style object, like this:
```js
const container = document.getElementById("app");
const patch = init([styleModule, attributesModule]);
const vnode1 = h("section", [
  h("span", { style: { color: "red" }, attrs: { id: "span_1" } }, "Hello")
]);
patch(container, vnode1);

const vnode3 = h("section", [
  h("span", { style: {},attrs: { id: "span_1" } }, "Hello")
]);
patch(vnode1, vnode3);
```
It is working.
Could somebody provide an incorrect example?

Censwin:
fix [#1000](https://github.com/snabbdom/snabbdom/issues/1000)

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,20 +524,6 @@ In JSX, you can use `style` like this:
 // Renders as: <div style="border: 1px solid #bada55; color: #c0ffee; font-weight: bold"></div>
 ```
 
-Note that the style module does not remove style attributes if they
-are removed as properties from the style object. To remove a style,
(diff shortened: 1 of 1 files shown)


jvanbruegge:
Thanks

PR #1005, opened 2022-02-20
Export `Options` type from the entry point

ryota-ka:
This is a speculative pull request addressing https://github.com/snabbdom/snabbdom/pull/975#issuecomment-997180772, so please feel free to close this if you don't need this.

---

Exporting the `Options` type, so that other libraries in the ecosystem (e.g. `@cycle/dom`) can depend on it.


jvanbruegge:
Thanks!

Issue #1007, opened 2022-03-19
Patch fragment VNode Faild


cqh963852:
https://codesandbox.io/s/focused-dan-g6xck2?file=/src/index.ts

![image](https://user-images.githubusercontent.com/17702287/159131174-2b06704a-74c7-44b0-929c-65b8f040c9d1.png)

While patch a fragment node. It will show a common error. 

Can I patch a subtree that not from root? 
It seems work if node created by h function.

cqh963852:
https://codesandbox.io/s/naughty-gauss-ocj5cf?file=/src/index.ts

a demo from https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment

the fragment node's parentNode is null

cqh963852:
While add a new element to fragment. 
![image](https://user-images.githubusercontent.com/17702287/159171185-9a621f33-1f6f-45f9-af2f-83179c36188c.png)
The parentEl is fragment. and before is null.
So ,the new node can't be added.

cqh963852:
![image](https://user-images.githubusercontent.com/17702287/159172164-ed5d6428-4c66-487c-aebe-786f26b28ac9.png)
While referenceNode is not null,It can work.
But if adding a node to fragment's 0 index. the referenceNode will be null.

jvanbruegge:
Thanks for the investigation. Can you open a PR with your solution and add a test to catch the error?
Also cc @ryota-ka because you wrote the fragment support :)

rosostolato:
@cqh963852 any luck on this?

rosostolato:
I found a solution for this and added a commit for the [PR](https://github.com/snabbdom/snabbdom/pull/1015) of the other [issue](https://github.com/snabbdom/snabbdom/issues/1014) I opened yesterday.

cqh963852:
> @cqh963852 any luck on this?

Going to closing this issue. Because the maintainer has no response to my PR

kuraga:
@cqh963852, I understand your feelings. But I think @jvanbruegge will find a time for this in early May.

@cqh963852, I understand your feelings. But let's reopen this and #1009...

Thanks!

PR #1009, opened 2022-03-21
fix: change parent elm, if fragment

cqh963852:
Select parent from fragment's children[0]'s parentNode.

Because fragmentDocument's parent always null.


cqh963852:
relate to https://github.com/snabbdom/snabbdom/issues/1007

cqh963852:
The logic of isDocumentFragment seems not right. So I just use `elm?.nodeType` determine whether it is fragment

image

Issue #1014, opened 2022-04-11
[Bug] Can't patch a rendered fragment with an actual vnode.


rosostolato:
Fragments can't be patched with another vnode.

```js
// If it was not a fragment, it would work.
const fnode = fragment([h('div', 'test')]);
patch(container, fnode);

const node = h('h1', 'Hello World!');
patch(fnode, node);
```

The example above, the dom remains as "<div>test</div>"

example in [stackblitz](https://stackblitz.com/edit/typescript-9y66o1?file=index.ts)

rosostolato:
I found a potential fix for the bug and added a pull request on https://github.com/snabbdom/snabbdom/pull/1015.

kuraga:
@rosostolato, is #1007 the same as this?

/cc @cqh963852

rosostolato:
@kuraga the two reports are different. This one is referent to a bug that doesn't allow you to replace an element node with a fragment and vice versa. The other one is an error that is throwed on console when working with fragments.

But the pull request that I have opened covers these 2 bugs. The other PR was closed by the author.

PR #1015, opened 2022-04-12
fix(fragment): fixed patching fragment nodes

rosostolato:
Fix for the bug mentioned on https://github.com/snabbdom/snabbdom/issues/1014.

To fix the bug, it was needed to keep the Ref of the parent that the `DocumentFragment` was appended to. So the easiest fix was to create a `parent` property on fragment ref inside the `insertBefore` hook (maybe we'll need to add it to other domApi hooks too, but it's working just with this one).

The tests for fragment patch weren't right, even with success it was bugging on real browsers. So I fixed the tests to reproduce the bug.

I will be happy to help with any questions that come out.

### DIFF
--- a/src/htmldomapi.ts
+++ b/src/htmldomapi.ts
@@ -75,6 +75,14 @@ function insertBefore(
   newNode: Node,
   referenceNode: Node | null
 ): void {
+  if (isDocumentFragment(parentNode)) {
+    let node: Node | null = parentNode;
+    while (node && isDocumentFragment(node)) {
+      const fragment = parseFragment(node);
+      node = fragment.parent;
+    }
+    parentNode = node ?? parentNode;
+  }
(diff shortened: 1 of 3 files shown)


rosostolato:
These are examples in stackblitz to repro the bugs:
https://stackblitz.com/edit/typescript-9y66o1
https://stackblitz.com/edit/typescript-63rtg9
https://stackblitz.com/edit/typescript-gdptsp

jvanbruegge:
Thanks for the PR, but I sadly won't have time to look over it until the beginning of May

rosostolato:
@jvanbruegge that's fine. In the meantime I'm using the fork on my project and I can find more bugs and add to this PR.

jvanbruegge:
Sorry it took so long, I was ill. The PR looks good, thanks.

PR #1020, opened 2022-05-13
fix: datasets -> dataset

nexteditorjs:
bug fix: toVNode use datasets bug dataset module use dataset


jvanbruegge:
Thanks for the PR!

PR #1028, opened 2022-07-19
Make delayed styles a bit more robust

tjenkinson:
I added some more tests for the `delayed` property on styles and these are the ones that fail on master

```
  style
    delayed
      ✖ handles delayed style change that gets removed before next frame
        Chrome Headless 103.0.5060.114 (Mac OS 10.15.7)
        Firefox 100.0 (Mac OS 10.15)
      uncaught exception: AssertionError: expected '16px' to equal '14px' (/var/folders/rx/zl6tqvx16d1_rr9g8n34wc_n3qycv_/T/karma-typescript-bundle--59926-6twtE7difV0f-.js:1342)

      ✖ applies base style immediately and reschedules delayed change
        Chrome Headless 103.0.5060.114 (Mac OS 10.15.7)
        Firefox 100.0 (Mac OS 10.15)
      uncaught exception: AssertionError: expected '18px' to equal '16px' (/var/folders/rx/zl6tqvx16d1_rr9g8n34wc_n3qycv_/T/karma-typescript-bundle--59926-6twtE7difV0f-.js:1342)

      ✖ reverts a delayed style synchronously
        Chrome Headless 103.0.5060.114 (Mac OS 10.15.7)
        Firefox 100.0 (Mac OS 10.15)
      uncaught exception: AssertionError: expected '16px' to equal '14px' (/var/folders/rx/zl6tqvx16d1_rr9g8n34wc_n3qycv_/T/karma-typescript-bundle--59926-6twtE7difV0f-.js:1342)

      ✖ removes a delayed style synchronously
        Chrome Headless 103.0.5060.114 (Mac OS 10.15.7)
        Firefox 100.0 (Mac OS 10.15)
      uncaught exception: AssertionError: expected '16px' to equal '' (/var/folders/rx/zl6tqvx16d1_rr9g8n34wc_n3qycv_/T/karma-typescript-bundle--59926-6twtE7difV0f-.js:1342)
```

So as above this makes the behaviour the following:
- base props are added and removed synchronously
- delayed props are scheduled to apply on the next frame when they change
- base prop changes are always applied synchronously even when they have a matching prop in `delayed`
- if a base prop is changed and is in `delayed`, the delayed prop is scheduled to apply again on the next frame
- when a prop is removed from `delayed` we synchronously switch back to the base one, or remove the prop if it's not in the base

So now the actual classes should be completely deterministic based on the `style` property.

I also updated the build config to target es5, meaning we could start safely using arrow functions and some other features. This is needed because I started using object destructuring and needed that to be compiled down to es5

And finally I removed the `destroy` property because not sure what the value of it was? Given the element is being deleted any style change wouldn't be visible anyway? It also wasn't tested

### DIFF
--- a/test/unit/style.ts
+++ b/test/unit/style.ts
@@ -192,6 +192,41 @@ describe("style", function () {
         });
       });
     });
+    it("handles base style getting removed before next frame", function (done) {
+      const vnode1 = h("i", {
+        style: { fontSize: "16px", delayed: { fontSize: "18px" } as any },
+      });
+      const vnode2 = h("i", {
+        style: { delayed: { fontSize: "18px" } as any },
+      });
+      elm = patch(vnode0, vnode1).elm;
+      assert.strictEqual(elm.style.fontSize, "16px");
+      elm = patch(vnode1, vnode2).elm;
+      assert.strictEqual(elm.style.fontSize, "");
+      requestAnimationFrame(() => {
+        requestAnimationFrame(() => {
+          assert.strictEqual(elm.style.fontSize, "18px");
+          done();
+        });
+      });
+    });
+    it("handles base style getting removed when it's also in delayed", function (done) {
+      const vnode1 = h("i", {
+        style: { fontSize: "16px", delayed: { fontSize: "18px" } as any },
+      });
+      const vnode2 = h("i", {
+        style: { delayed: { fontSize: "18px" } as any },
+      });
+      elm = patch(vnode0, vnode1).elm;
+      requestAnimationFrame(() => {
+        requestAnimationFrame(() => {
+          assert.strictEqual(elm.style.fontSize, "18px");
+          elm = patch(vnode1, vnode2).elm;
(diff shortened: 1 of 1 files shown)


src/modules/style.ts
```diff
@@ -1,114 +1,215 @@
...
-  }
-  const compStyle = getComputedStyle(elm as Element);
-  const props = (compStyle as any)["transition-property"].split(", ");
-  for (; i < props.length; ++i) {
+  Object.keys(style).forEach((key) => {
+    applied.push(key);
+    elm.style[key as any] = style[key]!;
+  });
+  const compStyle = getComputedStyle(elm);
+  const props = compStyle.transitionProperty.split(", ");
```
tjenkinson, line 205: this used to be `transition-property`, which I think was a mistake? Typescript thinks `transition-property` doesn't exist anyway and the tests still pass

tjenkinson: this also goes away with https://github.com/snabbdom/snabbdom/pull/1030

PR #1048, opened 2023-04-21
NIT: fix README's example

kraktus:
Thank you for maintaining snabbdom.

A small fix to make the README's example compilable by default when copy-pasting.

Feel free to close if too niche.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ const patch = init([
 
 const container = document.getElementById("container");
 
-const vnode = h("div#container.two.classes", { on: { click: someFn } }, [
+const vnode = h("div#container.two.classes", { on: { click: () => console.log("div clicked") } }, [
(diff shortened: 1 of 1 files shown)


paldepind:
Hi @kraktus. Thanks for taking the time to submit a PR. I definitely see the point that it would be nice if the example was working code that could be copy-pasted.

Have you tried running the code with the suggested changes? I would think that the `console.log`s should be wrapped in a function (a la `() => console.log("...")`) in order for things to work?

Maybe some more descriptive output could be logged? What do you think about logging "div clicked" for the initial `div` and then logging "updated div clicked" for the changed `div`?

kraktus:
Thanks for the nice review, sorry for the type errors, was just transpiring instead of type-checking.

paldepind:
Looks great! Thanks for the contribution.

PR #1037, opened 2022-11-15
fix: optimization judge node attribute

Yuiai01:
use name.startsWith("data-") replace  name[0] === "d" && name[1] === "a"...

PR #1033, opened 2022-09-04
fix: optimization judge svg

jingjing20:
use sel.startsWith('svg') replace sel[0] === s && sel[1] === v &&
sel[2] === g

PR #1057, opened 2023-10-25
feat: optimise detection of data and svg names

iambumblehead:
using these regexp tests would close two outstanding PRs
 1. https://github.com/snabbdom/snabbdom/pull/1037, optimize data attribute
 2. https://github.com/snabbdom/snabbdom/pull/1033, optimize svg selector

### DIFF
--- a/.github/workflows/steps.yml
+++ b/.github/workflows/steps.yml
@@ -9,12 +9,9 @@ on:
 jobs:
   ci:
     runs-on: ubuntu-latest
-
     steps:
-      - uses: actions/checkout@v2
-
+      - uses: actions/checkout@v3
       - run: npm ci
-
       - run: npm test
         env:
           BROWSER_STACK_USERNAME: ${{ secrets.BROWSER_STACK_USERNAME }}


kuraga:
(there is a smile of the old man)

https://github.com/snabbdom/snabbdom/pull/996#discussion_r786320401

paldepind:
Out of the possible solutions, I would say that say that the one using `startsWith` is the easiest to read, but I think we should go with the most efficient solution.

Is there any evidence to suggest that using regexes is the fastest solution?

paldepind:
I don't see how the regex test not throwing on non-strings is of much use. Passing anything but a string is an error anyway and does not have any use.

Since all three options are easy to understand and don't have any ramifications for the rest of the code, I think the key concern here is to pick the most performant version.

Thanks a lot for creating a benchmark, I think that is very useful to guide the decision here. Correct me if I'm wrong, but what I'm seeing is that the regex version is the slowest option. The benchmark tool outputs operations/second, so a higher number is better. This also matches what we see, for instance, `equalityFull` is clearly going to be slower than `equality`, and it has a lower number. So, it seems that using `startsWith` is the fastest option.

iambumblehead:
> The last number is the rate of operations measured in ops/sec (higher is better).

Excuse the confusing summary I gave originally, I was reading the results wrong myself...  Yes, you are correct: regexp is least performant and startsWith is most performant. I'll update this PR to use startsWith.

iambumblehead:
I didn't expect the PR to be closed...

paldepind:
Ok, great, then we agree on the benchmark 🙂. I think using `startsWith` is best as well.

Dunno why the PR got closed, but GitHub won't let me reopen it, so I think you'll have to create a new one.

PR #1058, opened 2023-10-25
feat: optimise detection of data and svg names

iambumblehead:
per https://github.com/snabbdom/snabbdom/pull/1057, use startsWith to optimise the detection of data and svg names

this PR would satisfy two outstanding PRs
 1. https://github.com/snabbdom/snabbdom/pull/1037, optimize data attribute
 2. https://github.com/snabbdom/snabbdom/pull/1033, optimize svg selector

### DIFF
--- a/src/h.ts
+++ b/src/h.ts
@@ -16,7 +16,7 @@ export type VNodeChildren = ArrayOrElement<VNodeChildElement>;
 export function addNS(
   data: any,
   children: Array<VNode | string> | undefined,
-  sel: string | undefined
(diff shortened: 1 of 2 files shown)


paldepind:
Thanks for the PR @iambumblehead and especially for the benchmark which shows that this is in fact an optimization.

The only "downside" to this is that it doesn't work in Internet Explorer. But at this point, I don't think we have to worry about that anymore.

iambumblehead:
thank you @paldepind if I am useful that makes me happy

Issue #1059, opened 2023-10-25
`npm test` hangs indefinitely


iambumblehead:
```bash
snabbdom$ npm test

> snabbdom@3.5.1 test
> npm run build && npm run lint && npm run unit


> snabbdom@3.5.1 build
> tsc


> snabbdom@3.5.1 lint
> eslint --ext .ts,.tsx,.js --ignore-path .gitignore .


> snabbdom@3.5.1 unit
> web-test-runner "test/unit/*.{ts,tsx}" --node-resolve --coverage


Chrome: |                              | 0/9 test files | 0 passed, 0 failed

Running tests...
```
Please tell me what I am doing wrong, `npm test` always hangs here for me.

I have also followed the steps here https://github.com/snabbdom/snabbdom/issues/960#issuecomment-862594658

iambumblehead:
For me this patch resolves the issue and tests complete when using a system with chrome
```diff
diff --git a/web-test-runner.config.js b/web-test-runner.config.js
index 527fc7d..60faf8e 100644
--- a/web-test-runner.config.js
+++ b/web-test-runner.config.js
@@ -14,7 +14,7 @@ const sharedCapabilities = {
 
 export default {
   concurrentBrowsers: 2,
-  concurrency: ci ? 2 : undefined,
+  concurrency: ci ? 2 : 1,
   browsers: !ci
     ? undefined
     : Object.values(browsers).map((cap) =>
```

PR #1060, opened 2023-10-25
chore(test): use default concurrency 1, so local unit tests do not hang

iambumblehead:
closes https://github.com/snabbdom/snabbdom/issues/1059


jvanbruegge:
whoops, that was not intended, thanks for the PR

Issue #972, opened 2021-06-14
Exception thrown when updating a text node


cjohansen:
Related t #971. I'm not sure what the intended behavior for text nodes is, but I had hoped this would work:

```js
var el = document.createElement('div');
document.body.appendChild(el);

// Works as expected
var vdom = patch(el, snabbdom.vnode(undefined, {}, [], 'Text', null));

// Throws exception
var vdom2 = patch(vdom, snabbdom.vnode(undefined, {}, [], 'Text 2', null));
```

The second `patch` call throws an exception where it tries to do `getAttribute` on `null`. Is this even supposed to work?

Changing an `h2` to a bare text works:

```js
var vdom = patch(el, snabbdom.vnode("h2", {}, [], 'Text', null));
patch(vdom, snabbdom.vnode(undefined, {}, [], 'Text 2', null));
```

But the other way around does not:

```js
var vdom = patch(vdom, snabbdom.vnode(undefined, {}, [], 'Text 2', null));
patch(el, snabbdom.vnode("h2", {}, [], 'Text', null));
```

Issue #971, opened 2021-06-12
Can't patch document fragment created from a text node


cjohansen:
I'm trying to patch in a text node, using a document fragment:

```js
var patch = snabbdom.init([
  snabbdom.propsModule
]);

var el = document.createElement('div');
document.body.appendChild(el);

var fragment = document.createDocumentFragment();
fragment.appendChild(document.createTextNode("Text"));

var vdom = patch(el, snabbdom.toVNode(fragment));
```

This throws an exception when snabbdom tries to `createElement` and the vnode `sel` is an empty string. The changelog mentioned document fragments being supported, but perhaps some other way? Am I doing it wrong?

jvanbruegge:
Document fragments should work, but sadly I can't have a deeper look at this at the moment as I am busy the next two weeks

cjohansen:
After looking into this some more I realize that the initial report is a little inaccurate. The problem is that calling `toVNode` with a document fragment created from only a text node creates a vnode that snabbdom can't render.

PR #1062, opened 2023-10-26
fix: patch should apply textNodes (#972)

iambumblehead:
closes https://github.com/snabbdom/snabbdom/issues/972
closes https://github.com/snabbdom/snabbdom/issues/971

Feel free to leave criticism. Changes are not smoke-tested, however all unit-tests are passing.

Some parts of createElm were rearranged to support these scenarious,
 * `vnode("", {}, [], "Test Text 1", null)` if there is no selector, a textNode is created as one of the first conditions
 * `vnode("h1", {}, [], "Test Text h1", null)` if there is a selector, a primitive vnode text AND an empty child list, a textNode is childed to the element

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -160,6 +160,9 @@ export function init(
         vnode.text = "";
       }
       vnode.elm = api.createComment(vnode.text!);
+    } else if (sel === "") {
+      // textNode has no selector
+      vnode.elm = api.createTextNode(vnode.text!);
     } else if (sel !== undefined) {
       // Parse selector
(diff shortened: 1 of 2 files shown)


iambumblehead:
```diff
-     if (
-       is.primitive(vnode.text) &&
-       (!is.array(children) || children.length === 0)
-     ) {
+     if (is.primitive(vnode.text)) {
```
I prefer the above change but used the array check to making a change that would deviate too much from the existing behaviour

Issue #958, opened 2021-05-02
Patching DOM elements loses img src attribute values


jarmo:
I'm having a html similar to this:
```html
<body>
  <main>
    <div class="image-1">
      <img src="xxx">
    </div>
    <div class="image-2">
      <img src="yyy">
    </div>
  </main>
</body>
```

After some user interaction I will perform a request to the server which returns new `innerHTML` for `main`. Since most of that innerHTML is the same then I decided to give Virtual DOM a try to avoid changing too much. Here's how I tried to do it:
```javascript
fetch("/new-content")
  .then(response => response.text())
  .then(html => {
    var newContentEl = document.querySelector("main").cloneNode(false)
    newContentEl.innerHTML = html
    patch(toVNode(document.querySelector("main")), toVNode(newContentEl))
  })
```

It will result in html similar to like this:
```html
<body>
  <main>
    <div class="image-1">
      <img>
    </div>
    <div class="image-2">
      <img>
    </div>
  </main>
</body>
```

Notice that both `img` tag `src` attributes have been removed although I can see them being there in the `html` variable got from `fetch`. All other seems to be in place, the only problem seems to be `img` elements `src` attributes which will change after each `fetch`.

If I don't use Virtual DOM then everything works as expected:
```javascript
fetch("/new-content")
  .then(response => response.text())
  .then(html => {
    var newContentEl = document.querySelector("main")
    newContentEl.innerHTML = html
  })
```

What am I doing wrong?

iambumblehead:
This test fails because toVNode does not capture the src property from the original dom element.
```javascript
it("can process img src attributes", function () {
  const prevElm = document.createElement("img");
  const newElm = document.createElement("img");
  prevElm.src = "http://other/";
  newElm.src = "http://localhost/";

  const vnode2 = toVNode(newElm)

  // test passes, when the below line is un-commented
  // vnode2.data.props = { src: "http://localhost/" }
  
  elm = patch(toVNode(prevElm), vnode2).elm;
  assert.strictEqual(elm.src, "http://localhost/");
});
```

iambumblehead:
This issue seems to touch a number of overlapping things,

Currently, the below test incorrectly nests img inside another empty vnode,
```javascript
it("can patch img vnode attributes", function () {
  const vnode1 = toVNode((e => (
    e = document.createElement("img"),
    e.src = "http://other/",
    toVNode(e)
  ))())
  const vnode2 = toVNode((e => (
[...]

PR #1066, opened 2023-11-09
fix: img src attribute patching (#958)

iambumblehead:
closes https://github.com/snabbdom/snabbdom/issues/958

**possibly there is no issue** and this PR should be closed, see comment https://github.com/snabbdom/snabbdom/pull/1066#issuecomment-1804737446 Essentially, src attribute is patched correctly when the patch function is composed around the attributes module/plugin

### DIFF
--- a/src/tovnode.ts
+++ b/src/tovnode.ts
@@ -24,7 +24,12 @@ export function toVNode(node: Node, domApi?: DOMAPI): VNode {
       if (name.startsWith("data-")) {
         dataset[name.slice(5)] = elmAttrs[i].nodeValue || "";
       } else if (name !== "id" && name !== "class") {
-        attrs[name] = elmAttrs[i].nodeValue;
+        if (name === "src") {
+          data.props = data.props || {};
+          data.props[name] = elmAttrs[i].nodeValue;
+        } else {
+          attrs[name] = elmAttrs[i].nodeValue;
+        }
       }
     }
     for (i = 0, n = elmChildren.length; i < n; i++) {
--- a/test/unit/core.ts
+++ b/test/unit/core.ts
@@ -449,6 +449,18 @@ describe("snabbdom", function () {
       }
     });
     describe("using toVNode()", function () {
+      it("can patch the src attribute of an image elment", function () {
+        const prevElm = document.createElement("img");
+        const newElm = document.createElement("img");
+        prevElm.src = "http://other/";
+        newElm.src = "http://localhost/";
+        const vnode1 = toVNode(prevElm);
+        const vnode2 = toVNode(newElm);
+        assert.strictEqual(vnode2.data.props.src, "http://localhost/");
+        patch(vnode0, vnode1);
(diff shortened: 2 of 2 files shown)


iambumblehead:
this PR depends on another PR https://github.com/snabbdom/snabbdom/pull/1062 for reasons outlined in this comment https://github.com/snabbdom/snabbdom/issues/958#issuecomment-1804645696

iambumblehead:
feel free to leave any suggestions or criticisms or anything at all

kuraga:
@iambumblehead , thanks!

Are there any additional to `src` attributes need to be handled this way?

iambumblehead:
@kuraga experimenting a little bit here... maybe there is no bug.

When the patch function is composed with the attributes module, src-patching behaviour works correctly. For example, when the test is moved from `test/unit/core.ts` to `test/unit/attributes.ts`, the test passes. 

What do you think?

> Are there any additional to src attributes need to be handled this way?

Searching for canonical information about the src attribute hasn't returned anything useful so far.

jarmo:
@iambumblehead I've totally forgotten about #958 (it's been over 2.5 years), but I don't understand how moving test into different file make this not a bug/unexpected behavior?

I mean, I have not tried the same situation described in #958 with latest changes in this repo (decided to use something else at the time), but if it still behaves the way described in there then I would say this is a bug regardless of the test location.

iambumblehead:
@jarmo the failing test, inside `test/unit/core.ts` constructs `patch` this way,
```javascript
const patch = init(
  [classModule, propsModule, eventListenersModule],
  undefined,
  { experimental: { fragments: true } },
);
```

The same test passes, when it is moved to `test/unit/attributes.ts` which constructs `patch` this way
```javascript
const patch = init([attributesModule]);
```

In other words, the patch function must use the `attributes` module to handle attributes like "src" and so the issue can be solved on the user's end by including the `attributes` module in the patch function.

iambumblehead:
@jarmo thanks for messaging here over two years later :)

jarmo:
@iambumblehead a-ha! Thanks for the explanation. As I understand then the issue was probably an user error in this case. Unfortunately I don't have code anymore to check how was `patch` created at the time, but it seems that `attributesModule` was not included as per your example (and test). I would not be too surprised if I copied it from here as it is the first example under _documentation_ https://github.com/snabbdom/snabbdom#init without noticing that some module is not there.

Thanks for putting your time into it and I guess I can close #958 now thanks to you!

paldepind:
Thank you for looking into this @iambumblehead. I have only skimmed the conversation. Is this PR still relevant?

iambumblehead:
> Thank you for looking into this @iambumblehead. I have only skimmed the conversation. Is this PR still relevant?

The PR can be closed however you may want a different PR updated the first README example to include the "attributes" module. If the attributes module were included, it would increase the chance of success for new users who start from copy-pasting the example.

PR #1070, opened 2023-11-09
chore(tools): added github package publishing job, runs when release is created

iambumblehead:
This job publishes snabbdom to github's npm package repository and it will only run when a release is created. Github's package repository requires packages to be "scoped". Un-scoped package names like "snabbdom" need to have a scope added to them, so this part applies a scope to the name before publishing, so the published name becomes "@snabbdom/snabbdom"
```javascript
const fs = require('fs').promises;
fs.readFile('package.json', 'utf8').then((data) => JSON.parse(data)).then((json) => {
  json.name = '@$(echo "$GITHUB_REPOSITORY" | sed 's/\/.\+//')/' + json.name;
  console.info('Package name changed to %s', json.name);
  return fs.writeFile('package.json', JSON.stringify(json), 'utf8');
}).catch(error => {
  console.error(error);
  process.exit(1);
});
```

An example package published to github's package repository is https://github.com/iambumblehead/esmock/pkgs/npm/esmock

Mainly, this adds some small visual "bling" to the side of snabbdom's main github page https://github.com/snabbdom/snabbdom because when the package is published to github, a descriptive little box linking to the package will appear there at the side, for example see the little box at the side of this page https://github.com/iambumblehead/esmock

related link https://docs.github.com/en/packages/quickstart

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,33 @@
+name: npm package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  publish-gpr:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 20
+          registry-url: https://npm.pkg.github.com/
+      - name: Insert repository owner as scope into package name
+        run: |
+          node <<EOF
+          const fs = require('fs').promises;
+          fs.readFile('package.json', 'utf8').then((data) => JSON.parse(data)).then((json) => {
+            json.name = '@$(echo "$GITHUB_REPOSITORY" | sed 's/\/.\+//')/' + json.name;
+            console.info('Package name changed to %s', json.name);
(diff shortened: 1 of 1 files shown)


paldepind:
So with this, we will have a package on npm with the name "snabbdom" and a package on Githubs package repository with the name "@snabbdom/snabbdom"?

Is having a package in Githubs repo something that you need? I'm asking because I have never used it.

kuraga:
I'm uneasy about _automatic_ publishing... Thoughts?

jvanbruegge:
This does not publish automatically. When a new release is created (manually) then this will also upload the package to github's package registry. I never used the github registry, but it does not hurt

kuraga:
> This does not publish automatically. When a new release is created (manually) then this will also upload the package to github's package registry.

Aha, _release_ != _tag_. Then ok.

Just a tag is not necessarily an application's new version. It could be a preliminary one or accidentally pushed.

iambumblehead:
> Is having a package in Githubs repo something that you need?

I've never used a github package and do not know if there is any benefit to publishing one. Afaik it just adds bling to the side of the main repo page and maybe pleases the microsoft github gods.

PR #1072, opened 2023-11-20
chore(tools): remove devdependency cross-env; un-maintained, not used ~86kB

iambumblehead:
This PR removes the cross-env devDependency, which is un-maintained. Snabbdom sources do not appear to use cross-env anywhere,
 * https://www.npmjs.com/package/cross-env
 * https://packagephobia.com/result?p=cross-env


paldepind:
Nicely spotted 👍

PR #1073, opened 2023-11-20
chore(tools): remove devdependency lodash-es; used once, ~622kB

iambumblehead:
Removes lodash-es devdependency, ~622kB
 * https://packagephobia.com/result?p=lodash-es

### DIFF
--- a/test/unit/core.ts
+++ b/test/unit/core.ts
@@ -1,5 +1,4 @@
 import { assert } from "@esm-bundle/chai";
-import { shuffle } from "lodash-es";
 
 import {
   init,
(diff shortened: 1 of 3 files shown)


test/unit/core.ts
```diff
@@ -848,7 +847,7 @@ describe("snabbdom", function () {
               return spanNumWithOpacity(n, "1");
             }),
           );
-          const shufArr = shuffle(arr.slice(0));
+          const shufArr = arr.slice(0).sort(() => 0.5 - Math.random());
```
paldepind, line 850: This "[Microsoft shuffle](https://www.robweir.com/blog/2010/02/microsoft-random-browser-ballot.html)" is not a good way of shuffling an array. See also [the comments on this SO answer](https://stackoverflow.com/a/18650169/556072).

If we want to remove the Lodash dependency I think we should define our own `shuffle` function that does a proper shuffle. The implementation in the bottom of [this SO answer](https://stackoverflow.com/a/12646864/556072) seems like a fine candidate.

iambumblehead: the SO shuffle is added here, but maybe lodash-shuffle was better than using a custom function.

feel free to close this PR or merge either way

test/unit/core.ts
```diff
@@ -31,6 +30,17 @@ const patch = init(
   { experimental: { fragments: true } },
 );
```
paldepind, line 32:
```suggestion
/** Shuffle an array using Durstenfeld's version of the Fisher–Yates shuffle. */
```

iambumblehead: this was added locally, rather than through the github interface to preserve the git commit message formatting

paldepind (approved):
Looks good! 🎉

PR #1077, opened 2023-11-22
chore(tools): apply prettier-format to all files

iambumblehead:
This PR applies prettier formatting to all files eg `prettier --write .`


paldepind:
Thanks.

PR #487, opened 2019-11-22
Removed broken link of  Benchmark from readme

labtocat:
closes #402

PR #1078, opened 2023-11-23
chore(docs): add attributesModule to first README example (#958)

iambumblehead:
This PR updates the first README example to use the attributesModule, to increase the chance of success for people who start by copy-pasting the README example.

Related https://github.com/snabbdom/snabbdom/pull/1066

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ patch(vnode, newVnode); // Snabbdom efficiently updates the old view to the new 
 - [Virtual Node](#virtual-node)
   - [sel : String](#sel--string)
   - [data : Object](#data--object)
-  - [children : Array<vnode>](#children--arrayvnode)
+  - [children : Array\<vnode\>](#children--arrayvnode)
   - [text : string](#text--string)
   - [elm : Element](#elm--element)
   - [key : string | number](#key--string--number)
@@ -270,25 +270,22 @@ server-side generated content.
 ```mjs
 import {
   init,
-  classModule,
-  propsModule,
   styleModule,
-  eventListenersModule,
+  attributesModule,
   h,
   toVNode
 } from "snabbdom";
 
 const patch = init([
-  // Init patch function with chosen modules
-  classModule, // makes it easy to toggle classes
-  propsModule, // for setting properties on DOM elements
+  // Init patch function with the modules used by `toVNode`
   styleModule, // handles styling on elements with support for animations
-  eventListenersModule // attaches event listeners
+  attributesModule // handles attributes, for example an image src attribute
 ]);
 
 const newVNode = h("div", { style: { color: "#000" } }, [
   h("h1", "Headline"),
-  h("p", "A paragraph")
+  h("p", "A paragraph"),
+  h("img", { attrs: { src: "sunrise.png", alt: "morning sunrise" } })
 ]);
 
 patch(toVNode(document.querySelector(".container")), newVNode);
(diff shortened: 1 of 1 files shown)


jvanbruegge:
It's a bit weird that the attributesModule is added but not actually used in the example. Maybe you could add an image with a source attribute there?

iambumblehead:
@jvanbruegge the test pipeline is stuck waiting for a "ci" job not sure why

An image with src and alt attributes is added to the example.

paldepind:
I think it's a good idea to address the issue in #958 through documentation. The problem there was related to using `toVNode` without having the attributes module enabled.

With that in mind, perhaps it would be more prudent to update [the example in the documentation for `toVNode`] and make it clear which modules `toVNode` depends on. The current documentation and example do not make this clear at all. It imports the props module which is not used at all by `toVNode`. From a quick look, it seems that `toVNode` only relies on the attributes module and the dataset module. I would suggest updating the example for `toVNode` to use only these two modules and write (perhaps in the comments) that `toVNode` constructs vnodes that make use of these modules.

iambumblehead:
The README is updated. Feel free to nitpick this PR if you wish any other change.

<details>
  <summary>what the updated tovnode section looks like in this PR now</summary>

### `tovnode`

Converts a DOM node into a virtual node. Especially good for patching over an pre-existing,
server-side generated content.

```mjs
import {
  init,
  classModule,
  styleModule,
  attributesModule,
  eventListenersModule,
  h,
  toVNode
} from "snabbdom";

const patch = init([
  // Init patch function with chosen modules
  classModule, // makes it easy to toggle classes
  propsModule, // for setting properties on DOM elements
  styleModule, // handles styling on elements with support for animations
  attributesModule, // handles attributes, for example an image src attribute
  eventListenersModule // attaches event listeners
]);

const newVNode = h("div", { style: { color: "#000" } }, [
  h("h1", "Headline"),
  h("p", "A paragraph"),
  h("img", { attrs: { src: 'sunrise.png', alt: 'morning sunrise' } })
]);

patch(toVNode(document.querySelector(".container")), newVNode);
```

</details>

paldepind:
> Feel free to nitpick this PR if you wish any other change.

Ok, here are my suggestions.

* Make no changes to the first example. This is to keep that example simple.
* When building `init` in the tovnode section only provide the attributes module and the dataset module as these are the ones that `toVNode` relies on.
* Change the comment for `init` in the tovnode section to "// Init patch function with the modules used by `toVNode`". This will make it explicit why those modules are there.

This I feel will precisely address the original issue about `toVNode` not being documented clearly enough.

iambumblehead:
the requested changes are applied

iambumblehead:
A few things which are not directly related to this PR. 

I tried to replace 'husky' with 'simple-git-hooks' in order to replace the `.husky` directory and files with single-line hook-definitions inside package.json. Using 'simple-git-hooks' did not succeed, because commithelper fails with a cryptic error. 

The performance benchmarks for inferno at https://github.com/krausest/js-framework-benchmark/ are looking really good and I have two thoughts about this; 1) maybe I should continue using inferno and 2) why haven't snabbdom maintainers added snabbdom to the benchmark suite there? The Snabbdom README makes this claim "Splendid performance. Snabbdom is among the fastest virtual DOM libraries." but Snabbdom developers seem disinterested in verifying that claim.

As evidenced by open PRs and issues, Snabbdom has some important bugs and is missing crucial tests, some of which could have been obtained by merging some contributed PRs.

paldepind:
@iambumblehead As you can see [here](https://github.com/snabbdom/snabbdom/pull/487) there used to be a link to a benchmark that included Snabbdom. But that benchmark suite was deprecated and replaced with something else.

PR #1082, opened 2023-11-29
fix: detect window.requestAnimationFrame function before calling

iambumblehead:
essentially, this is the same as https://github.com/snabbdom/snabbdom/pull/885

### DIFF
--- a/src/modules/style.ts
+++ b/src/modules/style.ts
@@ -8,9 +8,10 @@ export type VNodeStyle = Record<string, string> & {
 
 // Binding `requestAnimationFrame` like this fixes a bug in IE/Edge. See #360 and #409.
 const raf =
-  (typeof window !== "undefined" &&
-    window.requestAnimationFrame.bind(window)) ||
-  setTimeout;
+  typeof window?.requestAnimationFrame === "function"
+    ? window.requestAnimationFrame.bind(window)
+    : setTimeout;
+
 const nextFrame = function (fn: any) {
   raf(function () {
     raf(fn);


src/modules/style.ts
```diff
@@ -9,6 +9,7 @@ export type VNodeStyle = Record<string, string> & {
 // Binding `requestAnimationFrame` like this fixes a bug in IE/Edge. See #360 and #409.
 const raf =
   (typeof window !== "undefined" &&
+    typeof window.requestAnimationFrame === "function" &&
```
mreinstein, line 12: does typescript have the optional chaining feature yet? If so, this might be cleaner:
```typescript
const raf = (typeof window?.requestAnimationFrame === "function" && …
```

iambumblehead: okay :) and there is a precedent for using that way as well
```bash
snabbdom水$ grep -ir \?\\. ./src/
./src/thunk.ts:  const ns = thunk.data?.ns;
./src/init.ts:  const isSameIs = vnode1.data?.is === vnode2.data?.is;
./src/init.ts:    const key = children[i]?.key;
./src/init.ts:      const init = data.hook?.init;
./src/init.ts:        hook.create?.(emptyNode, vnode);
./src/init.ts:    } else if (options?.experimental?.fragments && vnode.children) {
./src/init.ts:      data?.hook?.destroy?.(vnode);
./src/init.ts:          const removeHook = ch?.data?.hook?.remove;
./src/init.ts:    const hook = vnode.data?.hook;
./src/init.ts:    hook?.prepatch?.(oldVnode, vnode);
./src/init.ts:      vnode.data?.hook?.update?.(oldVnode, vnode);
./src/init.ts:    hook?.postpatch?.(oldVnode, vnode);
```

just a moment and I'll update this PR

iambumblehead: it is updated to look this way
```javascript
// Binding `requestAnimationFrame` like this fixes a bug in IE/Edge. See #360 and #409.
const raf = typeof window?.requestAnimationFrame === "function"
  ? window.requestAnimationFrame.bind(window)
  : setTimeout
```

iambumblehead:
@mreinstein I _think_ there is a project-wide setting to require the "ci" job to pass and the "ci" job is not longer there because it was renamed "test"

mreinstein:
Is this the commit that broke it?  https://github.com/snabbdom/snabbdom/commit/856f26a52ea3d92add1d65dc3e5758a54e1fe51e If so can we put it back? Having the ci system broken is very bad.

iambumblehead:
@mreinstein I'm confident the workflow issues will be resolved :)

Let's see what @jvanbruegge thinks. Reverting the changes is one option and another option is to move forward with the name "test" for the test pipeline. 

https://docs.github.com/en/actions/using-workflows/required-workflows
> Note: GitHub is deprecating support for required workflows for GitHub Actions.

paldepind:
LGTM! Thanks for the PR.

PR #1083, opened 2023-11-29
fix: allow innerHTML to replace non-empty node, credit @tokichie

iambumblehead:
Copy-pasted changes from @tokichie https://github.com/snabbdom/snabbdom/pull/347 re-applied to the current master branch

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -138,7 +138,9 @@ export function init(
     return function rmCb() {
       if (--listeners === 0) {
         const parent = api.parentNode(childElm) as Node;
-        api.removeChild(parent, childElm);
+        if (parent !== null) {
+          api.removeChild(parent, childElm);
+        }
       }
     };
   }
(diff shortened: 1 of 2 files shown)


iambumblehead:
happy new year bump!

src/init.ts
```diff
@@ -138,7 +138,9 @@ export function init(
     return function rmCb() {
       if (--listeners === 0) {
         const parent = api.parentNode(childElm) as Node;
-        api.removeChild(parent, childElm);
+        if (parent) {
```
paldepind, line 141:
```suggestion
        if (parent !== null) {
```

Just to be more explicit about what is checked here.

paldepind:
LGTM, except for the one nitpick comment :)

src/init.ts
iambumblehead: thank you, the suggested change is amended to the commit

paldepind:
Thanks!

PR #1080, opened 2023-11-28
fix: generate valid javascript modules in build. fixes #963.

mreinstein:

PR #1085, opened 2023-11-30
fix: correct minimum node version to one that supports modules

mreinstein:
Currently our `package.json` says minimum node engine version is `8`, but we have `"type": "module"` in our `package.json` and that requires node v12+ as per https://nodejs.org/dist/latest-v20.x/docs/api/esm.html#modules-ecmascript-modules

Screenshot 2023-11-30 at 9 42 15 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "url": "https://github.com/snabbdom/snabbdom/issues"
   },
   "engines": {
-    "node": ">=12.0.0"
+    "node": ">=12.17.0"
   },
   "files": [
     "build"
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "url": "https://github.com/snabbdom/snabbdom/issues"
   },
   "engines": {
-    "node": ">=8.3.0"
+    "node": ">=12.0.0"
   },
   "files": [
     "build"


iambumblehead:
The current published project is type "commonjs" and not type "module" and that is seen at the package.json here https://www.npmjs.com/package/snabbdom?activeTab=code

Maybe the published Snabbdom sources have type "module" removed from them, I don't know...

kuraga:
@mreinstein , why was there 12.18? https://github.com/snabbdom/snabbdom/pull/1080#issuecomment-1832076246

mreinstein:
It depends on how one defines support. technically it will work in node > 12.0, but that requires a command line flag. 12.17 and up work without the command line flag.

iambumblehead:
note: if dependents use other files than the main exported file, those other files must be explicitly exported at the package.json.

Should exports for these files be added to the package.json?
```
snabbdom木$ tree build/helpers/
build/helpers/
├── attachto.d.ts
├── attachto.js
└── attachto.js.map
```

mreinstein:
> Should exports for these files be added to the package.json

Out of scope for this PR, I want to keep this focused on fixing one singular problem.

iambumblehead:
I'm sharing this here as a sort of starting point in case the package.json must define exports. There are many different ways of doing something like this, but the below (basic) pattern would allow another package to do something like `import attachto from 'snabbdom/build/helpers/attachto.js'`
```json
{
  "name": "snabbdom",
  "exports": {
    ".": {
      "import": "./build/index.js",
      "types": "./build/index.d.ts",
      "default": "./build/index.js"
    },
    "./build/helpers/*.js": "./build/helpers/*.js",
    "./build/modules/*.js": "./build/modules/*.js"
  }
}
```

mreinstein:
Please put the exports stuff it into a separate issue/PR.  It's orthogonal.  I don't want to clog up this issue with separate but related items.

paldepind:
The minimum Node version is a bit artificial as Snabbdom is a browser library and doesn't work in Node. 

That being said if a version of Node can't import Snabbdom then it certainly doesn't support Snabbdom, so bumping the version seems reasonable. I do wonder though if removing the "engine" entry altogether would have any ramifications? It was added in #948.

mreinstein:
> the minimum Node version is a bit artificial as Snabbdom is a browser library

`"type": "module"` in `package.json` is also a purely node feature.  With these 2 pieces of node-only metadata we're declaring a config that isn't possible.   Converted to English our node metadata says something like "We use es modules. Make sure you are running this module on a version of node that existed way before es modules were supported, or higher"


>  and doesn't work in Node.

You have to be careful with assumptions.  Server side rendering is one case where snabbdom _might_ be invoked outside of a browser. The only parts of snabbdom that actually interface with the browser are neatly encapsulated in `src/htmldomapi.ts`, and it's really quite trivial to stub out that very limited api on `window.` and ``document.` globals.


> I do wonder though if removing the "engine" entry altogether 

Can you explain why there is a hesitancy around bumping the node version up to `12.something` ?  This seems like a non-issue unless I'm missing something.

paldepind:
Those are very good points 👍. I didn't think about server-side rendering – proving that you're absolutely right about not making too many assumptions.

Thanks for the PR.

> `"type": "module"` in package.json is also a purely node feature.

It also serves as info that web bundlers can use. Webpack, for instance, uses that field to determine which module format to use.

PR #1092, opened 2024-01-01
Enhance JSX and VNode types

WorldMaker:
- Use lib.dom's `HTMLTagNameMap` for HTML element property names in JSX
- Use lib.dom's `ElementCSSInlineStyle` for style property names in all VNodes

The meta-typing for HTML element property names could also be applied to `h` overloads as well and more VNode situations in general (`HTMLTagNameMap` exists for mapping overloads to among other things `document.createElement(elementName: string)`) but I didn't already have a working equivalent on hand, I just had the JSX specific types, so that isn't included in this PR.

This was JSX meta-typing that I discovered/built for [Butterfloat](https://worldmaker.net/butterfloat/), but having used Snabbdom in the past and may use Snabbdom in the future I thought would be useful to share the meta-programming wealth for better auto-complete in more JSX environments.

### DIFF
--- a/src/jsx.ts
+++ b/src/jsx.ts
@@ -106,11 +106,11 @@ export namespace jsx {
   /*
   IfEquals/WritableKeys: https://stackoverflow.com/questions/52443276/how-to-exclude-getter-only-properties-from-type-in-typescript/52473108#52473108
   */
-  export type IfEquals<X, Y, A = X, B = never> = (<T>() => T extends X
+  export type IfEquals<X, Y, Output> = (<T>() => T extends X
     ? 1
     : 2) extends <T>() => T extends Y ? 1 : 2
-    ? A
-    : B
+    ? Output
+    : never
 
   export type WritableKeys<T> = {
     [P in keyof T]-?: IfEquals<
--- a/src/jsx.ts
+++ b/src/jsx.ts
@@ -121,7 +121,7 @@ export namespace jsx {
   }[keyof T]
 
   export type ElementProperties<T> = {
-    [Property in WritableKeys<T> as T[Property] extends string | number ? Property : never]?: T[Property]
+    [Property in WritableKeys<T> as T[Property] extends string | number | null | undefined ? Property : never]?: T[Property]
   }
 
   export type VNodeProps<T> = ElementProperties<T> & Props
--- a/src/jsx.ts
+++ b/src/jsx.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-namespace, import/export */
 import { Key, vnode, VNode, VNodeData } from "./vnode";
 import { h, ArrayOrElement } from "./h";
+import { Props } from "./modules/props";
 
 // for conditional rendering we support boolean child element e.g cond && <tag />
 export type JsxVNodeChild =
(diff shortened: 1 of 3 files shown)


src/modules/style.ts
```diff
@@ -1,9 +1,13 @@
 import { VNode, VNodeData } from "../vnode";
 import { Module } from "./module";
 
-export type VNodeStyle = Record<string, string> & {
-  delayed?: Record<string, string>;
-  remove?: Record<string, string>;
+export type ElementStyle = {
+  [Property in keyof ElementCSSInlineStyle]?: ElementCSSInlineStyle[Property]
+}
```
paldepind, line 6: This doesn't seem right. The type `ElementCSSInlineStyle` only contains the keys `style` and `attributeStyleMap`. But what we want is the CSS style names themselves (`background`, `fontStyle`, etc.). It seems that `CSSStyleDeclaration` is the type we want instead. When I change the type to that I get auto-completions for the CSS style names.

Also, `Partial` can be used instead of the explicit map to create a partial version of an object.

paldepind:
Thanks for this PR. These typing improvements look great! 👍

> The meta-typing for HTML element property names could also be applied to h overloads as well and more VNode situations in general (HTMLTagNameMap exists for mapping overloads to among other things document.createElement(elementName: string)) but I didn't already have a working equivalent on hand, I just had the JSX specific types, so that isn't included in this PR.

If you're up to it you're very welcome to add these improvements to `h` as well. If not, that is completely fine as well.

src/jsx.ts
```diff
@@ -101,7 +102,35 @@ export function jsx(
 // See https://www.typescriptlang.org/docs/handbook/jsx.html#type-checking
 export namespace jsx {
   export type Element = VNode;
-  export interface IntrinsicElements {
+
+  /*
+  IfEquals/WritableKeys: https://stackoverflow.com/questions/52443276/how-to-exclude-getter-only-properties-from-type-in-typescript/52473108#52473108
+  */
+  export type IfEquals<X, Y, A = X, B = never> = (<T>() => T extends X
```
paldepind, line 109: I would prefer this type without the defaults. The first default is never used and I think no defaults make things a bit clearer and more explicit.

src/modules/style.ts
WorldMaker: `ElementCSSInlineStyle` I thought I copied from working code, so I'll triple check it there too when I get a chance. I hope it is not something silly like different lib.dom versions, but it might be a silly mistake of a different sort.

Good catch on the `Partial<>`, that mapped type has an additional transformation in my original code, but in this form you are right is just Partial. I'll correct that.

src/jsx.ts
WorldMaker: Good points. Because it is StackOverflow copypasta intended for more usages I left it as-is, but it might be good to simplify it.

paldepind:
Thanks for addressing the feedback.

Do you want the PR to be merged as is or do you want to perform changes to `h` as well? :)

WorldMaker:
Welcome!

I mentioned `h` mostly because I saw it as out of scope for this PR. I might follow up a later PR with it if I get the itch, but I probably won't as I personally tend to use JSX far more than `h` and it might be a fun task on the backlog for someone else to explore who is interested in TS types mechanics.

paldepind:
That makes sense and is completely understandable. Thanks a lot for the PR and its improvements! 👍

Issue #303, opened 2017-06-04
dataset module removes blank string data attributes


johanalkstal:
It's a usual case to want to have an empty data attribute as a JS hook, like `data-hook`.
There are two ways to add such an attribute, with the `snabbdom/modules/attributes` or `snabbdom/modules/dataset`
```
var snabbdom = require('snabbdom')
var patch = snabbdom.init([
  require('snabbdom/modules/attributes').default,
  require('snabbdom/modules/dataset').default
]);
var h = require('snabbdom/h').default

var container = document.getElementById('container');

var vnode = h('div', [
  h('p', { attrs: { 'data-hook': ''} }, 'Has data-hook by attribute'),
  h('p', { dataset: { hook: '' } }, 'Has data-hook by dataset')
])

patch(container, vnode)
```
```
<div>
  <p data-hook="">Has data-hook by attribute</p>
  <p data-hook="">Has data-hook by dataset</p>
</div>
```

Now let's update the vdom elements with new content.

```
var updatedVnode = h('div', [
  h('p', { attrs: { 'data-hook': ''} }, 'Still has data-hook by attribute'),
  h('p', { dataset: { hook: '' } }, 'Has lost its data-hook by dataset')
])

patch(vnode, updatedVnode)
```
```
<div>
  <p data-hook="">Still has data-hook by attribute</p>
  <p>Has lost its data-hook by dataset</p>
</div>
```

The element updated with `snabbdom/modules/dataset` has lost its data attribute.

This only happens with an empty value.

This forces you to use `snabbdom/modules/attributes` for these cases.

caridy:
we should probably update https://github.com/snabbdom/snabbdom/blob/master/src/modules/dataset.ts#L21, using `in` might be better here... but then we have the issue of `null` and `undefined` as values that will be stringify, which is the behavior of the browser anyways. e.g.: `elm.setAttribute("hook", undefined)` it produces `data-hook="undefined"`.

kuraga:
Still buggy?

wanghaida:
It also has a problem, when it's `data-hook: 0`,  you should use `if (oldDataset === undefined && dataset === undefined) return` in https://github.com/snabbdom/snabbdom/blob/master/src/modules/dataset.ts#L14

Issue #1093, opened 2024-01-15
the styleModule has problems handling style updates with attribute values of 0


chuifengji:
**the styleModule has problems handling style updates with attribute values of 0**

``` typescript
for (name in oldStyle) {
    if (!style[name]) { // here's the mistake
      if (name[0] === "-" && name[1] === "-") {
        (elm as any).style.removeProperty(name);
      } else {
        (elm as any).style[name] = "";
      }
    }
  }
```
The above code should be changed to：
``` typescript
for (name in oldStyle) {
    if (!(name in style)) { // how to fix it
      if (name[0] === "-" && name[1] === "-") {
        (elm as any).style.removeProperty(name);
      } else {
        (elm as any).style[name] = "";
      }
    }
  }
```

paldepind:
Thank you for reporting this @chuifengji. I'll take a look at this.

In the meantime, I think you should be able to work around the problem by using `"0"` instead of just `0` as your flex-shrink value.

paldepind:
Can you describe exactly what problem you are experiencing? I can see that the line you highlight is not ideal, but it's not clear how this results in a bug (except for the property being set to `""` and then re-set correctly when the new properties are handled).

paldepind:
@chuifengji Please see #1094 which should fix your problem.

chuifengji:
thanks!

PR #1094, opened 2024-01-16
Handle falsy values correctly in the dataset and style modules

paldepind:
This PR fixes an issue where falsy values such as `0` or `""` where accidentally being removed by the dataset and styles module during diffing.

ISSUES CLOSED: #303, #1093

### DIFF
--- a/src/modules/dataset.ts
+++ b/src/modules/dataset.ts
@@ -18,7 +18,7 @@ function updateDataset(oldVnode: VNode, vnode: VNode): void {
   const d = elm.dataset;
 
   for (key in oldDataset) {
-    if (!dataset[key]) {
+    if (!(key in dataset)) {
       if (d) {
         if (key in d) {
           delete d[key];
--- a/src/modules/style.ts
+++ b/src/modules/style.ts
@@ -1,12 +1,13 @@
 import { VNode, VNodeData } from "../vnode";
 import { Module } from "./module";
 
-export type ElementStyle = Partial<CSSStyleDeclaration>
+export type ElementStyle = Partial<CSSStyleDeclaration>;
 
-export type VNodeStyle = ElementStyle & Record<string, string> & {
-  delayed?: ElementStyle & Record<string, string>;
-  remove?: ElementStyle & Record<string, string>;
-};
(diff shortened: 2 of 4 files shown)


mreinstein:
Nice.  I think this would be considered an API breaking change, right?  We probably need to bump the npm major version when this lands?

kuraga:
It was a bug, no? 😁

mreinstein:
It was a bug, but it could be considered depended-upon behavior in the API. after this change more attributes will be present that weren't before.

iambumblehead:
like this change :+1:

paldepind:
Thanks for the feedback.

Regarding breakage, that is a good point. We need to be careful that this is not a breaking change in any way.

The original bug reported in #303 is the following:

```js
const vnode1 = h("i", { dataset: { foo: "" } });
const vnode2 = h("i", { dataset: { foo: "" } });
patch(elm, vnode1);
elm = patch(vnode1, vnode2).elm;
console.log(elm.dataset.foo); // is undefined but should be ""
```

The fact that `foo` disappears is a bug and not a behavior that anyone can rely on. So fixing this is not a breaking change.

The current check `!dataset[name]` is somewhat different from the new check `!(name in dataset)`. While different I don't see how it will result in a breaking change. Consider for instance, a `dataset` with the value `{ foo: "bar" }` getting updated to `{ foo: "" }`. Then, with the old check, the code will actually think that `foo` has been removed and delete the property. Then, further on when updating properties `foo` will be set to `""`. So this change is not observable but avoids the temporary accidental removal.

mreinstein:
> The fact that foo disappears is a bug and not a behavior that anyone can rely on. So fixing this is not a breaking change.

Maybe it would be more accurate to say that "it seems unlikely this behavior is being relied upon".  This incurs a change to the html that is produced (`foo` would be a present attribute now) and technically that _could_ be relied upon in some other place.

paldepind:
Yes, in theory, it could be relied upon. In this case, though, I think the buggy behavior is sufficiently unexpected and sufficiently not-useful that fixing the bug does not need to be considered a breaking change.

PR #1098, opened 2024-01-21
Document the selector property and its use in identifying nodes

### LINKED ISSUE #333: Document role of the `sel` property in comparison
Hi everyone,

I was looking through the source code of snabbdom, and I think the sel properties of a vnode should be more detailed in the documentation.

Indeed, one would want to use this sel properties as a shorthand for adding a class or an id to a vnode. However in the current implementation, despite the fact it is indeed a shorthand for doing that, it is also  a shorthand for setting a key-like property on this vnode (according to sameVnode function).

the same Vnode function:

```javascript
function sameVnode(vnode1: VNode, vnode2: VNode): boolean {
}
```
Example :

```javascript
var a = h("div#myDiv", ["This is a (slightly) hidden keyed div"])
// The selector property of the vnode is "div#myDiv" 
// props object will not exist nor contain reference to the id property

var b = h("div", {props: {id: "myDiv"}}, ["This is a simple div with same id as a"])
// The selector property of the vnode is "div" 
// The props object does contain a reference to the id property

patch(document.body, a)
patch(a, b)  // Triggers remove/destroy hook on a, and insert hook on b
```

That was confusing for me, and probably for other people too (like those coming from Mithril).

It is probably a wanted behavior, but it should be made more explicit in the documentation.

paldepind:
Document the selector better and mention that it is used to determine vnode "sameness" as pointed out in #333.

Closes #333

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,14 +229,14 @@ Of course, then there is still a single comment node at the mount point.
 ### `h`
 
 It is recommended that you use `h` to create vnodes. It accepts a
-tag/selector as a string, an optional data object, and an optional string or
-an array of children.
+[tag/selector](#sel--string) as a string, an optional data object, and an
+optional string or an array of children.
 
 ```mjs
 import { h } from "snabbdom";
 
-const vnode = h("div", { style: { color: "#000" } }, [
-  h("h1", "Headline"),
+const vnode = h("div#container", { style: { color: "#000" } }, [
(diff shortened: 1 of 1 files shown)


iambumblehead:
The writing is clear and compact. Like this :+1:

paldepind:
Thanks for the feedback!

PR #1101, opened 2024-01-24
chore(tools): add build job to publish spec

iambumblehead:
this PR adds a build job to the publish spec, when the publish spec was used a few days ago it failed because the spec includes "needs: build" and because no "build" job was defined. https://github.com/snabbdom/snabbdom/actions/runs/7593114398

"build" would be used to transpile or generate any resources that should be included in the published package.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,12 +5,18 @@ on:
     types: [created]
 
 jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: npm ci
+      - run: npm run build
   publish-gpr:
     needs: build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v4
         with:
           node-version: 20
           registry-url: https://npm.pkg.github.com/


.github/workflows/publish.yml
```diff
@@ -5,6 +5,11 @@ on:
     types: [created]
 
 jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: npm ci
```
jvanbruegge, line 12: This currently only installs dependencies, doesn't it have to build the `.js` files too?

iambumblehead: `npm run build` is added to the publish spec build job and setup-node action version is incremented
```diff
diff --git a/.github/workflows/publish.yml b/.github/workflows/publish.yml
index dd768c4..84399c8 100644
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,12 +10,13 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - run: npm ci
+      - run: npm run build
   publish-gpr:
     needs: build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v4
         with:
           node-version: 20
           registry-url: https://npm.pkg.github.com/
```

iambumblehead: Not sure if pressing the `[Resolve conversation]` button here is expected or faux pas, so just leaving this open. Feel free to ignore or resolve whichever way.

paldepind:
My vote would be for just removing this job again. As far as I know, no user has ever requested Snabbdom to be available on GitHub's npm package repository so I don't think it adds much value.

Issue #1103, opened 2024-01-28
Why can't the file be found after updating 3.5.1 to 3.6.0?


royWoody:
![image](https://github.com/snabbdom/snabbdom/assets/8044025/b98e6eb2-bae6-412d-abc6-5de8495f39e9)

royWoody:
If the modification is not backward compatible, can the version be upgraded to 4.0.0?

jvanbruegge:
This should be fixed in `3.6.1`, please let me know if it is not

Issue #1102, opened 2024-01-26
3.6.0 './h' failed to resolve only because it was resolved as fully specified


ForsakenE:
build/jsx.js  use './h' to import js file ，bug in package.json，use type:'module'

paldepind:
Hello @ForsakenE. This bug report is a little bit hard to read. Is this the same issue as #1087? If not, please provide a bit more information 🙂

ForsakenE:
yes, i use webpack to build my project,but found this error&nbsp;&nbsp;'./h' failed to resolve only because it was resolved as fully specified ,the reason same as you provide number #1087




怪咖-&nbsp;
***@***.***



&nbsp;




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "snabbdom/snabbdom"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2024年1月27日(星期六) 凌晨0:20
***@***.***&gt;;
抄送:&nbsp;"怪咖- ***@***.******@***.***&gt;;
主题:&nbsp;Re: [snabbdom/snabbdom] 3.6.0 './h' failed to resolve only because it was resolved as fully specified (Issue #1102)





 
Hello @ForsakenE. This bug report is a little bit hard to read. Is this the same issue as #1087? If not, please provide a bit more information 🙂

ryota-ka:
I also ran into this issue after updating to v3.6.0.
I've submitted #1104 to fix this problem.

jvanbruegge:
This should be fixed in `3.6.1`

PR #1104, opened 2024-01-29
Add `.js` extensions to the import specifiers

### LINKED ISSUE #1087: "module" publishing; [BUG] runtime error importing build/src/jsx.js
importing master build/src/jsx.js is causing this runtime error
```bash
node:internal/modules/esm/resolve:264
    throw new ERR_MODULE_NOT_FOUND(
          ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/bumble/software/snabbdom/build/h' imported from /home/bumble/software/snabbdom/build/jsx.js
    at finalizeResolution (node:internal/modules/esm/resolve:264:11)
    at moduleResolve (node:internal/modules/esm/resolve:917:10)
    at defaultResolve (node:internal/modules/esm/resolve:1130:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:396:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:365:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:240:38)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:85:39)
    at link (node:internal/modules/esm/module_job:84:36) {
  code: 'ERR_MODULE_NOT_FOUND',
  url: 'file:///home/bumble/software/snabbdom/build/h'
}
```

the build process generates files that incorrectly import local files,
```diff
-import { h } from "./h"; // incorrect
+import { h } from "./h.js"; // correct
```

cc @mreinstein

ryota-ka:
Closes #1087, #1102 (perhaps also #1103?)

Added `.js` extensions to the import specifiers, which have been to done via `ts-add-js-extension`.
(Reminder: A valid ES module requires the file extension in the import specifier.)

This makes `build/jsx.js` a valid ES module file, as it now imports `./h.js` instead of `./h`.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -53,7 +53,6 @@
     {
       "files": ["examples/**/*.js"],
       "rules": {
-        "import/no-unresolved": "off",
         "@typescript-eslint/explicit-module-boundary-types": "off"
       }
     },
@@ -72,15 +71,13 @@
     {
       "files": ["README*md/*.tsx"],
       "rules": {
-        "@typescript-eslint/no-unused-vars": "off",
-        "import/no-unresolved": "off"
+        "@typescript-eslint/no-unused-vars": "off"
       }
     },
     {
       "files": ["CHANGELOG.md/*"],
       "rules": {
         "@typescript-eslint/no-unused-vars": "off",
-        "import/no-unresolved": "off",
         "node/no-missing-import": "off"
       }
     }
@@ -89,6 +86,7 @@
     "max-statements-per-line": "error",
     "no-var": "error",
     "import/newline-after-import": "error",
-    "import/no-default-export": "error"
+    "import/no-default-export": "error",
+    "import/no-unresolved": "off"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,7 +31,6 @@
         "prettier": "^3.0.3",
         "release-it": "^16.2.1",
         "serve": "^14.2.1",
-        "ts-add-js-extension": "^1.6.0",
         "typescript": "5.2.2"
       },
       "engines": {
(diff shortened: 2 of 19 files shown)


kuraga:
Has the user import instructions  changed?

ryota-ka:
I believe no.
Assuming that users are importing the library with `'snabbdom'` or `'snabbdom/build/foo.js'`, there should be no breaking changes.
This PR only affects the internal module resolution.

jvanbruegge:
I am still against adding `.js` extensions to `.ts` files. There is no need to go that route. This is a small bug in `ts-add-js-extension` that can easily worked around: https://github.com/snabbdom/snabbdom/pull/1105

kuraga:
> I am still against adding `.js` extensions to `.ts` files. There is no need to go that route. This is a small bug in `ts-add-js-extension` that can easily worked around: https://github.com/snabbdom/snabbdom/pull/1105

Explicit [would have been] better than implicit...

jvanbruegge:
Not if the explicit reference is to a file that does not exist.

jvanbruegge:
I've published `3.6.1` now please have a look if the problem persists

kuraga:
> Not if the explicit reference is to a file that does not exist.

Yes, you're right.
A reference to _`.js`_ file in a _TypeScript_. The which doesn't exist. And is a TS artifact instead.
(Just image we compile TS to AnotherScript.
Interesting, another side, why TypeScript compiler doesn't handle this itself.)

paldepind:
> Interesting, another side, why TypeScript compiler doesn't handle this itself.

There are a number of TypeScript issues related to this (https://github.com/microsoft/TypeScript/issues/42151, https://github.com/microsoft/TypeScript/issues/16577).

As far as I can see the official stance is that TypeScript will not change your code (changing the import from `foo` to `foo.js` would be changing code) and thus such a feature will not be implemented. The "official" solution is that if you want `.js` extensions in the output then your imports in TypeScript should have `.js` extensions.

ryota-ka:
Anyway, thank you for publishing the fix @jvanbruegge!
Everything is working as expected.

PR #1106, opened 2024-01-30
Detect opportunity to insert new vnode earlier for more efficient DOM manipulation

### LINKED ISSUE #1099: Virtualized list issue
I'm trying to build out a virtualize list within a table and am running into rendering issues.  Following a standard pattern of removing  and adding rows when scrolling through a list on a table. Scroll down, remove nodes from the top and add rows to the bottom while  keeping all indexes up to date with the local state of the entire list. 

issue I'm running into is when I scroll down I'm finding that the entire list re-rerenders after rows are removed from the top of the list.  When scrolling up there is no issue: rows are added to the top of the list and removes from the bottom with existing rows not re-rerendered.

I've narrowed down this issue to this code block: https://github.com/snabbdom/snabbdom/blob/710b319622557196db90422d67a3ed44d48908c6/src/init.ts#L374

deleted rows are pushed down in the dom until they reach the bottom and are removed. This causes all rows to re-render. I believe this is the case. Is it possible to verify and possibly propose a solution to avoid re-rendering all untouched rows when removed top rows/nodes?

paldepind:
This changes the child reordering patch process slightly. Before beginning to move elements around we add an extra case where we check if the end of the new remaining vnodes is brand new and could thus be handled by inserting a new DOM node.

This should address and close #1099.

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -1,5 +1,5 @@
 import { Module } from "./modules/module";
-import { vnode, VNode } from "./vnode";
+import { Key, vnode, VNode } from "./vnode";
 import * as is from "./is";
 import { htmlDomApi, DOMAPI } from "./htmldomapi";
 
@@ -49,7 +49,7 @@ function isDocumentFragment(
   return api.isDocumentFragment!(vnode as any);
 }
 
-type KeyToIndexMap = { [key: string]: number };
+type KeyToIndexMap = { [key: Key]: number };
 
 type ArraysOf<T> = {
   [K in keyof T]: Array<T[K]>;
@@ -66,7 +66,7 @@ function createKeyToOldIdx(
   for (let i = beginIdx; i <= endIdx; ++i) {
     const key = children[i]?.key;
     if (key !== undefined) {
-      map[key as string] = i;
+      map[key] = i;
     }
   }
   return map;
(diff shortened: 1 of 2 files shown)


test/unit/core.ts
```diff
@@ -837,6 +838,44 @@ describe("snabbdom", function () {
...
+        });
+        it("simultaneous removal in beginning and addition in end", function () {
+          let insertCount = 0;
+          const domApi: DOMAPI = {
+            ...htmlDomApi,
+            insertBefore: (a, b, c) => {
+              insertCount++;
+              htmlDomApi.insertBefore(a, b, c);
+            }
+          };
```
paldepind, line 867: This change has no effect on the resulting DOM, only on the process by which we get there. Hence the test wraps the `htmlDomApi` to be able to observe what goes on in there.

test/unit/core.ts
```diff
@@ -837,6 +838,44 @@ describe("snabbdom", function () {
           assert.equal(elm.children[1].innerHTML, "symbol");
           assert.equal(elm.children[2].innerHTML, "symbol");
         });
+        it("simultaneous addition in beginning and removal in end", function () {
```
paldepind, line 841: This test also passed before this PR. I'm adding it for symmetry with the test below and to make sure that the non-inefficent case reported in #1099 is now tested.

src/init.ts
```diff
@@ -49,7 +49,7 @@ function isDocumentFragment(
   return api.isDocumentFragment!(vnode as any);
 }
 
-type KeyToIndexMap = { [key: string]: number };
```
paldepind, line 52: This is unrelated to the PR but I just happened to improve this type as well.

PR #1107, opened 2024-02-01
Use modern JS features in additional places

paldepind:
This PR refactors Snabbdom to use modern JavaScript features in more places.

Note, that most of the changes are simply from using arrow functions, which causes a lot of line changes in the tests. The only "interesting" changes are in `src/init.ts`.

The PR also removes the `isUndef` and `isDef` functions. I think those functions where worth it back before the optional operators, where Snabbdom needed to dig into deep paths with optional values at every level, but today I don't think those functions are justified.

### DIFF
--- a/README-in_HN.md
+++ b/README-in_HN.md
@@ -390,10 +390,10 @@ patch(vnode1, vnode2);
 
 ```mjs
 const myModule = {
-  create: function (oldVnode, vnode) {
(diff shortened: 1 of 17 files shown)


kuraga:
:partying_face: 

Nevertheless `() => {}` isn't semantically equal to `function () {}`. Just for other's note.

PR #1109, opened 2024-02-25
fix: camelize dataset

### LINKED ISSUE #1013: vNode created via toVNode looses dataset since 3.4.0 release
Forgive me if im wrong, but it seems like since 3.4.0 release and #976 was merged. creating a vNode via `toVNode` and then patching with that node, dataset is lost.

https://stackblitz.com/edit/js-znkhkf?file=index.js

example:
```javascript
import { datasetModule, init, toVNode } from 'snabbdom';

const patch = init([datasetModule]);
const $container = document.getElementById('app');

const $tmp = document.createElement('div');
$tmp.innerHTML = `<h1 data-something="yes">hello</h1>`;
const newNode = toVNode($tmp);

patch($container, newNode);
```

If i roll back to 3.3.1 its works as expected. Perhaps `toVNode` just missed an update with the above mentioned change?

sabberworm:
Closes #1013

~~Also fix issue with class attributes where the tokens are separated by something other than a single U+0020 SPACE.~~

### DIFF
--- a/src/tovnode.ts
+++ b/src/tovnode.ts
@@ -2,6 +2,18 @@ import { addNS } from "./h";
 import { vnode, VNode } from "./vnode";
 import { htmlDomApi, DOMAPI } from "./htmldomapi";
 
+/**
+ * Transforms the given attribute name into a valid dataset property key
+ * according to https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion
+ *
+ * @param attributeName data- attribute name, must start with data-
+ */
+function datasetKey(attributeName: string) {
+  return attributeName
+    .slice(5)
+    .replace(/-([a-z])/g, (_, ch) => ch.toUpperCase());
+}
+
 export function toVNode(node: Node, domApi?: DOMAPI): VNode {
   const api: DOMAPI = domApi !== undefined ? domApi : htmlDomApi;
   let text: string;
@@ -22,7 +34,7 @@ export function toVNode(node: Node, domApi?: DOMAPI): VNode {
     for (i = 0, n = elmAttrs.length; i < n; i++) {
       name = elmAttrs[i].nodeName;
       if (name.startsWith("data-")) {
-        dataset[name.slice(5)] = elmAttrs[i].nodeValue || "";
+        dataset[datasetKey(name)] = elmAttrs[i].nodeValue || "";
       } else if (name !== "id" && name !== "class") {
         attrs[name] = elmAttrs[i].nodeValue;
       }
--- a/test/unit/core.ts
+++ b/test/unit/core.ts
@@ -598,8 +598,10 @@ describe("snabbdom", () => {
         assert.deepEqual(toVNode(onlyAttrs).data, { attrs: { foo: "bar" } });
         const onlyDataset = document.createElement("div");
         onlyDataset.setAttribute("data-foo", "bar");
+        onlyDataset.setAttribute("data-foo-bar", "baz");
+        onlyDataset.setAttribute("data-foo-:bar", "baz");
         assert.deepEqual(toVNode(onlyDataset).data, {
-          dataset: { foo: "bar" }
+          dataset: { foo: "bar", fooBar: "baz", ["foo-:bar"]: "baz" }
         });
         const onlyDatasets2 = document.createElement("div");
         onlyDatasets2.dataset.foo = "bar";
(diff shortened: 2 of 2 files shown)


byterider (approved):

byterider (approved):

paldepind:
Thanks a lot for the PR @sabberworm. The fix for data attributes with dashes bug looks great :+1: 

About the class attribute change, is this something that you need? To me, classes separated by anything other than single spaces seem very uncommon. Personally, I'd be ok with simply not supporting that.

sabberworm:
> About the class attribute change, is this something that you need? To me, classes separated by anything other than single spaces seem very uncommon. Personally, I'd be ok with simply not supporting that.

Yes, when HTML is rendered using templating, it’s common to end up with more than a single space between class names and we have this situation a lot. Besides, it just makes sense to be aligned with the [spec](https://infra.spec.whatwg.org/#ascii-whitespace) on that front

paldepind:
I see. I agree that the spec is a fine thing to keep in mind. In this case, I also think that the spec allows for some things that are rarely ever used in practice (such as separating classes with tabs and line breaks). I think it's a good idea to only introduce complexity if there is an actual need among users. If additional spaces arrises from the templating system you're using, then that is a good reason to add support for that.

Some thoughts on the implementation:

* The regex class `\s` includes some unicode whitespace characters and the spec only mentions ASCII whitespace characters.
* The use of `Boolean` is a bit too implicit (in general I find that function to be a bit of an anti-pattern). If I understand correctly its role here is to remove empty strings, so using `(s) => s != ""` would be more explicit.
* `filter` is usually not good for performance and here the use of `split` and `filter` causes an additional array to be allocated.

I think perhaps this implementation would be cleaner and more performant as it only creates one array:

```ts
const c = cn ? "." + cn.match(/[^\ ]+/g).join(".") : "";
```

This only handles repeated spaces. To support the other whitespace characters allowed by the spec we would need:

```ts
const c = cn ? "." + cn.match(/[^\t\r\n\f\ ]+/g).join(".") : "";
```

I don't have a strong preference but I guess I lean towards the first option as I've ever only seen spaces in the `class` attribute. But perhaps others have different experiences.

src/tovnode.ts
```diff
@@ -2,13 +2,25 @@ import { addNS } from "./h";
...
+/**
+ * Transforms the given attribute name into a valid dataset property key
+ * according to https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion
+ *
+ * @param attributeName data- attribute name, must start with data-
+ */
+function datasetKey(attributeName: string) {
+  return attributeName
+    .slice(5)
+    .replace(/-([a-z])/g, (_, ch) => ch.toUpperCase());
```
paldepind, line 14: I think this doesn't carry out the first step of the [linked conversion](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion). But I guess that's not very important.

src/tovnode.ts
```diff
@@ -2,6 +2,18 @@ import { addNS } from "./h";
...
 
+/**
+ * Transforms the given attribute name into a valid dataset property key
+ * according to https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion
+ *
+ * @param attributeName data- attribute name, must start with data-
+ */
+function datasetKey(attributeName: string) {
+  return attributeName
+    .slice(5)
```
HT2313, line 13:
```suggestion
    .toLowerCase()
    .slice(5)
```

Just to be inline with https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion

paldepind: We can not use `toLowerCase` as that does not match the MDN description. It states that only `A-Z` should be converted to lowercase, but `toLowerCase` also handles unicode. For instance `"Æ".toLowerCase()` gives `æ`. If we want to be in line with the spec I think using a regex would be easiest.

paldepind: It seems that lowercasing A-Z characters is also done at the [HTML level](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#in_html). So when reading elements through the DOM we can assume that there are no upper case A-Z characters. The current implementation should therefore be fine :+1:

HT2313 (approved):

sabberworm:
@paldepind I have removed to part with the class splitting and will open a separate PR for that.

paldepind:
@sabberworm Makes sense. Thanks a lot for the PR 👍. It's great to get this issue resolved.

byterider:
After upgrading to 3.6.3 we have missing datasets, not sure yet why.

PR #1110, opened 2024-03-01
Restore Typescript intellisense

bgradin:
Typescript intellisense appears to have been broken by #1052. For example:

```tsx
<div>Hello, world!</div>
```

The above code generates `ts(7026)`: `JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.` This is because as of #1052, `src/jsx.ts` no longer exports the `JSX` namespace within the parent `jsx` namespace, since the tsconfig option `isolatedModules` doesn't support the `import export` syntax that was used for this previously.

This pull request contains a workaround that avoids `import export`, but still allows the `JSX` namespace to be exported, so that `isolatedModules` can be set, but Typescript intellisense also works.

### DIFF
--- a/src/jsx.ts
+++ b/src/jsx.ts
@@ -100,7 +100,7 @@ export function jsx(
 }
 
 // See https://www.typescriptlang.org/docs/handbook/jsx.html#type-checking
-export namespace jsx {
+namespace JSXTypes {
   export type Element = VNode;
 
   /*
@@ -140,5 +140,11 @@ export namespace jsx {
 
   export interface IntrinsicElements extends HtmlElements {
     [elemName: string]: VNodeData;
+  }
+}
+
+export namespace jsx {
+  export namespace JSX {
+    export interface IntrinsicElements extends JSXTypes.IntrinsicElements {}
   }
 }


paldepind:
Hi @bgradin. Thanks for reporting the issue, finding the cause, and for the proposed solution.

The diff of the PR is a bit noisy. Do you know what's up with that? Our Prettier config can be applied with `npm run format`.

Tbh, I don't fully understand why we have this `jsx` namespace when the TypeScript docs talks about a `JSX` (upper case) namespace. But I haven't looked into it myself.

bgradin:
@paldepind Yeah, diff is noisy due to indentation, which is necessary because we're wrapping things in a "new" (more like re-introduced) namespace. I'm not sure there's a good way around that, unless we keep everything inside `export namespace JSX {}` at the same indentation level as before, which would be a bit sloppy to read.

bgradin:
@paldepind yeah, by default intellisense will look for `JSX.IntrinsicElements` within a namespace named the same as the JSX factory function. However, I realized there was a way to rewrite things to minimize the diff noise by simply renaming the `jsx` namespace and adding a new one that refers to the original.

paldepind:
> yeah, by default intellisense will look for JSX.IntrinsicElements within a namespace named the same as the JSX factory function.

Ah, I see. This is not made very clear in the TypeScript documentation, but I found it mentioned in [this section](https://www.typescriptlang.org/tsconfig#jsxFactory).

Thanks a lot for the PR! 👍 I'm planning to do a release pretty soon.

PR #1111, opened 2024-03-01
Improve whitespace handling of class attribute in `toVNode`

sabberworm:
Improved from #1109, based on @paldepind’s input.

The multi-space-separated case is definitely common when HTML is generated by templating languages.

Prettier likes to format class names across multiple lines, so `\n`, `\r` and `\t` (for indentation) has to be supported anyway. This meant `\f` is the only ASCII whitespace character that’s uncommon in class attributes but I didn’t really see a reason to exclude it either.


src/tovnode.ts
```diff
@@ -19,8 +19,8 @@ export function toVNode(node: Node, domApi?: DOMAPI): VNode {
   let text: string;
   if (api.isElement(node)) {
     const id = node.id ? "#" + node.id : "";
-    const cn = node.getAttribute("class");
-    const c = cn ? "." + cn.split(" ").join(".") : "";
+    const cn = node.getAttribute("class")?.trim();
+    const c = cn ? "." + cn.match(/[^\t\r\n\f ]+/g)!.join(".") : "";
```
paldepind, line 23: If I understand correctly the `trim` is to handle the only-whitespace case.

Would this work as well?

```diff
-    const cn = node.getAttribute("class")?.trim();
-    const c = cn ? "." + cn.match(/[^\t\r\n\f ]+/g)!.join(".") : "";
+    const classNames = node.getAttribute("class")?.match(/[^\t\r\n\f ]+/g);
+    const c = classNames ? "." + classNames.join(".") : "";
```

This would avoid the `trim` and the use of `!`.

test/unit/core.ts
```diff
@@ -591,7 +591,17 @@ describe("snabbdom", () => {
         assert.deepEqual(children[0].data, { dataset: { env: "xyz" } });
         assert.strictEqual(children[1].text, "Foobar");
       });
-
+      it("handles class names correctly", () => {
+        for (const [cl, sel] of [
+          ["  one\ttwo  three\nfour  \t", ".one.two.three.four"],
+          ["  \t \n ", ""]
```
paldepind, line 597: Really great that you thought of this case 👍

paldepind:
Thanks a lot for the PR.

> Prettier likes to format class names across multiple lines, so `\n`, `\r` and `\t` (for indentation) has to be supported anyway. This meant `\f` is the only ASCII whitespace character that’s uncommon in class attributes but I didn’t really see a reason to exclude it either.

I think that's solid reasoning. Definitely agree that there is no need to leave `\f` out as the only character omitted from the spec.

src/tovnode.ts
sabberworm: Good idea, will do that

paldepind:
Looks great 💪. Thanks again.

Issue #1112, opened 2024-03-02
Attribute behavior should be improved


bgradin:
TL;DR: The module system in Snabbdom provides a powerful way to extend the core functionality, but it leads to conventions for vnode attributes that, as a newcomer to this project, I found counterintuitive and would like to improve.

## Problem statement

For example, whereas with `React.createElement`, you would write something like `createElement("a", { href: "/" }, "Home")`, Snabbdom requires you to wrap attributes in an object called `attrs` so they can be handled by the attributes module: `h("a", { attrs: { href: "/" } }, "Home")`. I found this particularly unexpected when using JSX - a JSX snippet of `<a href="/">Home</a>` will create an HTML element without an href: `<a>Home</a>`!

## Proposal

I think the crux of the issue is that Snabbdom does not provide a mechanism for modules to specify a limited set of keys they will use to access data on a vnode. If a comprehensive list of these keys was maintained internally, any additional keys in the vnode data could be treated as attributes, rather than being disregarded. I think it would make the most sense to make this update in the `h` function, so both JS and JSX syntax would reap the benefits.

My preference would be for this to be the default functionality, but I suppose this might require a new major or at least minor release. If the maintenance team prefers, it should be possible to gate this functionality behind a configuration option in the `init` method to preserve backward compatibility.

It's worth pointing out that other projects like [herp-inc/snabbdom-jsx](https://github.com/herp-inc/snabbdom-jsx) provide functionality similar to what I'm describing. However, my point is the behavior of attributes is worth improving within this project, either by default or as a configuration option.

Barring this update, the readme should at least be updated to explicitly call out that JSX syntax still requires usage of the attributes module to assign attributes on elements.

## Conclusion

I would appreciate feedback from a maintainer on these ideas. If an agreement is reached, I would be happy to create a pull request. Thanks for your time!

kuraga:
Besides the `attrs`, there is also the `dataset`:

https://github.com/snabbdom/snabbdom/blob/2e6ef1e3e8637c845dca3bed8e8b661b52b56cd4/test/unit/core.ts#L616-L619

paldepind:
Hi @bgradin. Thanks for sharing your thoughts. I definitely see where you're coming from. I like the separation between modules and think that one can get used to writing `attrs`. But what you're proposing definitely has advantages in terms of ergonomics.

> I think the crux of the issue is that Snabbdom does not provide a mechanism for modules to specify a limited set of keys they will use to access data on a vnode.

[...]

PR #1117, opened 2024-05-19
[POC] feat: adds jsx props module

geotrev:
_NOTE: This PR is a POC to illustrate how Snabbdom could enable cleaner JSX prop handling per #1112, but as a non-breaking change that retains the package's existing architecture. It's also based on my [own plugin](https://github.com/geotrev/snabbdom-transform-jsx-props), but pared down as a better starting point._

### Todo

- [ ] Evaluate automatic DOM property vs attribute handling (similar to React and others)
- [ ] Update READMEs
- [ ] Type interfaces

## Descriptions

Adds `jsxPropsModule` to enable cleaner Snabbdom JSX prop signatures.

In short, the module forwards JSX props to their appropriate module (`props`, `dataset`, `eventlisteners`, etc). As a fallback, any unhandled props are treated as attributes.

This isn't behind an `options.experimental` flag since using the module feels like enough of an opt-in already, but open to suggestions if certain functionality should be instated to prevent foot-🔫s.

## Details

Use the module with `init` to create virtual nodes.

```
const patch = init(
  [
    jsxPropsModule,
    classModule,
    eventListenersModule
  ],
  htmlDomApi
)
```

Example of JSX using this module:

```
<div className="my-component" hook-insert={() => {}}>
  <h1 data-foo-heading={true}>Hello world</h1>
  <svg aria-hidden="true" />
  <a href="#" attrs-style="color: blue" tabIndex="0" on-click={() => {}}>Link</a>
</div>
```

View the unit test file for more samples. :) Any additional concerns, such as other modules/features to handle but aren't included in this POC, can be noted for a full and proper update should this module make sense for the package.

Requirements:

- `@babel/plugin-transform-react-jsx`, per README
- Must be added as the first module so the next modules (`eventlisteners`, `class`, `style`, etc) can detect these props.

The module handles these specific cases / features (open to suggestions/changes if appropriate):

- **Module instruction:** Providing a prefix instructs the module to add the suffix of the prop to the appropriate module. Some modules (like `class`) are omitted as they serve no functional purpose given the existence of a pre-existing, ergonomic option (in this case, `className`).

  | prefix    | module/feature  | prefix example       | result                                   |
  | --------- | --------------- | -------------------- | ---------------------------------------- |
  | `hook-*`  | hooks           | `hook-insert={fn}`   | `{ data: { hooks: { insert: fn }}}`      |
  | `on-*`    | event listeners | `on-mouseDown={fn}`  | `{ data: { on: { mouseDown: fn }}}`      |
  | `attrs-*` | attributes      | `attrs-value="foo"`  | `{ data: { attrs: { value: "foo" }}}`    |
  | `props-*` | props           | `props-value="foo"`  | `{ data: { props: { value: "foo" }}}`    |
  | `data-*`  | dataset         | `data-foo-bar="baz"` | `{ data: { dataset: { fooBar: "baz" }}}` |

  Hyphens are used for simplicity of the prop check. If a hyphen doesn't exist on the prop, then no module forwarding needs to happen. Open to ideas if React-style camelCase is better here.

- **Smarter prop handling:** Always moves a predefined set of JSX props into the `props` module (e.g., `className`). See `PROPERTY_PROPS` in `modules/jsxprops.ts` for the initial list (many others could be added).

  Examples:

  ```
  <div tabIndex={0} /> // { data: { props: { tabIndex: 0 }}}
  ```

- **Defaults to attributes:** Forwards all other props to the attributes module.
- **Non-breaking:** Does not conflict with the existing form of Snabbdom JSX props (e.g., `attrs={{...}}`), and shouldn't cause regressions for consumers mixing prop signatures, while preferring the flattened style introduced by this module.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -32,6 +32,7 @@ export { attributesModule } from "./modules/attributes";
 export { classModule } from "./modules/class";
 export { datasetModule } from "./modules/dataset";
 export { eventListenersModule } from "./modules/eventlisteners";
+export { jsxPropsModule } from "./modules/jsxprops";
 export { propsModule } from "./modules/props";
 export { styleModule } from "./modules/style";
 export type { Attrs } from "./modules/attributes";
--- a/src/modules/jsxprops.ts
+++ b/src/modules/jsxprops.ts
@@ -15,6 +15,7 @@ const DATASET = "dataset";
 const ATTRS = "attrs";
 const PROPS = "props";
 const DATA = "data";
+const KEY = "key";
 const HYPHEN_CHAR = "-";
 const MODULE_PROPS: string[] = ["hook", "on", ATTRS, PROPS, DATASET];
 const DOM_PROPS: string[] = ["className", "tabIndex", "id"];
@@ -38,7 +39,7 @@ const forwardJsxProps = (oldVNode: VNode, _vnode: VNode): void => {
 
   const data: Record<string, any> = {},
     propKeys = [];
-  // preserve existing module values to be updated with
+  // preserve module values
   for (const key in vnode.data) {
     if (MODULE_PROPS.indexOf(key) > -1) {
       data[key] = vnode.data[key];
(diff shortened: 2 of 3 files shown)


geotrev:
Side note: I couldn't find any documentation about getting browserstack tests to run, and it doesn't look like `.tsx` test files are running. Not sure what's up there.

Issue #1121, opened 2024-09-12
VNodeStyle type checking error for "delayed" and "remove" styles


zaccrites:
Hello. I am getting a typing error on VNodeStyles which use `delayed` and/or `remove`, despite following the [examples](https://github.com/snabbdom/snabbdom?tab=readme-ov-file#delayed-properties) from the README almost exactly.

```ts
const x: VNodeStyle = {
    opacity: "0",
    transition: "opacity 1s",
    delayed: {
        opacity: "1",
    },
    remove: {
        opacity: "0",
    }
};
```

```
Type '{ opacity: string; delayed: { opacity: string; }; remove: { opacity: string; }; }' is not assignable to type 'VNodeStyle'.
  Type '{ opacity: string; delayed: { opacity: string; }; remove: { opacity: string; }; }' is not assignable to type 'Record<string, string>'.
    Property 'delayed' is incompatible with index signature.
      Type '{ opacity: string; }' is not assignable to type 'string'.ts(2322)
```

This works fine at runtime, and I was able to fix the typing error by making the following change in `style.d.ts`:

```diff
import { Module } from "./module.js";
export type ElementStyle = Partial<CSSStyleDeclaration>;
- export type VNodeStyle = ElementStyle & Record<string, string> & {
+ export type VNodeStyle = ElementStyle & {
    delayed?: ElementStyle & Record<string, string>;
    remove?: ElementStyle & Record<string, string>;
};
export declare const styleModule: Module;
```

It is possible that the problem was introduced in [this refactor](https://github.com/snabbdom/snabbdom/commit/357c5464562730d01a8ef4497c3fec9de847335a#diff-6a39d21853236a9f42d95bf2b4d43a503cddcb00703694975146eb41733673b6).

Thanks!

zaccrites:
Assuming I am not misunderstanding something, I have created a PR for the suggested change: https://github.com/snabbdom/snabbdom/pull/1122

paldepind:
Hello @zaccrites. Thanks for reporting this. You are absolutely right that this ought to work. It seems that this regression was introduced in https://github.com/snabbdom/snabbdom/pull/1092.

PR #1122, opened 2024-09-12
Fix VNodeStyle type

zaccrites:
See https://github.com/snabbdom/snabbdom/issues/1121


paldepind:
Thanks a lot for the PR. I think one problems is that we also want to allow unknown style properties to be used. Something like this should also type-check:
```ts
const x: VNodeStyle = {
    myOwnWeirdStyle: "foo"
};
```

I think https://github.com/snabbdom/snabbdom/pull/1127 should fix the problem you reported while still allowing for the above to work (accidentally actually, I was looking at a performance issue).

Issue #1114, opened 2024-03-16
Slow type checking with snabbdom 3.6+


schlawg:
Recent releases after 3.5.1 seem to really trouble tsc.

We import and use snabbdom from ~ 31 modules of a monorepo.

Our type checking takes 3x as long when those modules import snabbdom 3.6 vs 3.5.1.

48s vs 14s

paldepind:
Hi @schlawg. Thanks for reporting this. That is quite a noticeable slowdown. Would it be possible for you to git bisect this to figure out what caused the issue?

One thing I can think of regarding typing is that the types for JSX was improved in https://github.com/snabbdom/snabbdom/pull/1092.

schlawg:
Hello there. Sorry for the delay in responding.

I can't spare any cycles to make an MRE at the moment, but I do have steps.

```
git clone https://github.com/lichess-org/lila
cd lila && ui/build -c --tsc
```
Note the time. Now do a global search/replace in ui folder, changing every `"snabbdom": "3.5.1"` to `"snabbdom": "3.6.2"` in package.json files.

`ui/build -c --tsc` again and see it takes roughly 3x as long with 3.6.x. It's probably some bizarre interaction with our codebase (and we're fine on 3.5.1 fyi)

paldepind:
Thanks a lot @schlawg. That seems like it should be enough to bisect the issue. I'll take a look when I have the time.

schlawg:
Hello @paldepind 

Your instincts were correct and it's to do with the VNodeData for jsx. I made an MRE for you here https://github.com/schlawg/snabbdom-tsc-mre

I'm sorry it took me so long. I will try not to take this excellent library for granted in the future!

paldepind:
Thanks for the reproduction @schlawg. I can see the performance difference and will look into this.

paldepind:
Please check out this PR: https://github.com/snabbdom/snabbdom/pull/1127.

I'd be very interested in whether this performance improvement is good enough for you (I think the easiest way to test it is to checkout the branch and locally install it with `npm i ../path/to/snabbdom/`).

If this is still too slow I think we should revert the style changes done in https://github.com/snabbdom/snabbdom/pull/1092.

schlawg:
[...]

PR #1127, opened 2024-11-24
fix: tweak `VNodeStyle` to improve performance of type checking

paldepind:
This should _somewhat_ improve the regression described in #1114.

Timing the build of [this example](https://github.com/schlawg/snabbdom-tsc-mre) with `tsc --extendedDiagnostics --noEmit` changes the build time from currently at around 5.4 seconds to 3.02 seconds (on my machine). The 3.5.1 release takes around 1.30 seconds. So build times are still slower, but I don't think we can make it any better if we want the precise type for `style` that reflects the actual CSS properties.

If this is fast enough we should merge this PR. If it is still not acceptable then we should revert the change to the `style` property to the less precise type.

### DIFF
--- a/src/modules/style.ts
+++ b/src/modules/style.ts
@@ -1,13 +1,17 @@
 import { VNode, VNodeData } from "../vnode";
 import { Module } from "./module";
 
-export type ElementStyle = Partial<CSSStyleDeclaration>;
+type ElementStyle = { [K in keyof CSSStyleDeclaration]?: string };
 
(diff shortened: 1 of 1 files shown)


schlawg:
For some reason (probably my own fault), I had to cast `style: { remove: { opacity: '0', transform: 'translate(...)' } } as any` in order to pass type checking against `VNodeStyle` before.

But this PR fixes that, along with the performance improvement!

PR #1064, opened 2023-10-26
chore(docs): reference readme-title image from imgur

iambumblehead:
```diff
diff --git a/README.md b/README.md
index 1d72a15..92038f9 100644
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<img alt="Snabbdom" src="readme-title.svg" width="356px">
+<img alt="Snabbdom" src="https://i.imgur.com/4Bq0tyi.png" width="356px">
 
 A virtual DOM library with a focus on simplicity, modularity, powerful features
 and performance.
```

This PR updates the image reference in the README from a relative image path to an imgur image url, and the reason for the change is discussed at the [bottom of this PR](https://github.com/snabbdom/snabbdom/pull/1063#issuecomment-1781595164)

imgur does not support svg. To preserve as much fidelity as possible, the image was imported to inkscape, scaled down to the README's 356px and then exported to PNG

PR #1136, opened 2026-02-16
chore(readme): add dark mode logo support for readability

MoustaphaCamara:
## Description

Improves README presentation by adding support for dark mode-aware logos. (minor)
This brought me to make this PR:
Capture d’écran du 2026-02-16 22-56-59

## Changes made

- replaced static  tag with a <picture> element
- added `prefers-color-scheme` sources for light and dark themes
- fallbacks to the default SVG logo for compatibility

This ensures the Snabbdom logo remains clearly visible in both light and dark modes on platforms like GitHub.

> Note: The image source points to the GitHub `master` branch, so the dark mode change may not be visible until merged.
You can replace it temporarily with a local `./readme-title-light.svg` to preview the effect locally.

Also, I’m wondering whether there’s a specific workflow or automated translation process in place for `README-in_HN.md` / `README-zh_CN.md`


README.md
```diff
@@ -1,4 +1,8 @@
-<img alt="Snabbdom" src="https://raw.githubusercontent.com/snabbdom/snabbdom/master/readme-title.svg" width="356px">
+<picture>
+  <source media="(prefers-color-scheme: light)" srcset="https://raw.githubusercontent.com/snabbdom/snabbdom/master/readme-title.svg">
```
paldepind: I might be wrong, but I actually don't think we need the full url here
```suggestion
  <source media="(prefers-color-scheme: light)" srcset="readme-title.svg">
```

paldepind (commented):
Thanks for the PR. This is a great idea!

What about creating a new directory for the logos? Maybe call it `resources` and put `graphics.svg` in there as well. Just to avoid too many files in the root.

> Also, I’m wondering whether there’s a specific workflow or automated translation process in place for `README-in_HN.md` / `README-zh_CN.md`

It's a manual process unfortunately.

README.md
MoustaphaCamara: Hi,
Thanks for your comments, I'm sorry I had not seen them earlier. I'll be addressing them later today! :+1:

MoustaphaCamara: I remember this is what I originally intended to do, but I noticed it had previously been changed the other way around. See commit: https://github.com/snabbdom/snabbdom/pull/1064/changes

That said, I agree that it’s better not to use an absolute link.

MoustaphaCamara: Even better, now we can see the changes applied here:
https://github.com/MoustaphaCamara/snabbdom/tree/readme-display-theme-readable-logo

